PR #15, opened 2021-07-19
WIP: calc lead time test func

kevin-kline:
Need to know:

- format we want to store this data
- format we get from JIRA

### DIFF
--- a/src/util/jira.js
+++ b/src/util/jira.js
@@ -1,0 +1,12 @@
+const jiraUtil = {
+  calculateLeadTime: (startDate, endDate) => {
+		try {
+			let seconds = endDate.getTime() - startDate.getTime() / 1000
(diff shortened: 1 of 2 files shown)


src/util/jira.js
```diff
@@ -0,0 +1,12 @@
+const jiraUtil = {
+  calculateLeadTime: (startDate, endDate) => {
+		try {
+			let seconds = endDate.getTime() - startDate.getTime() / 1000
+			return new Date(seconds * 1000).toISOString().substr(11, 8)
```
paulgoertzen-merico, line 5: Can you just return the seconds here instead of a date? I think we just want to store the interval seconds

paulgoertzen-merico:
I closed this because the calculation of lead time is very different from this version. See https://github.com/merico-dev/lake/pull/18 for an example of how we'll be calculating it. Feel free to build off that.

PR #17, opened 2021-07-19
Job queue setup

paulgoertzen-merico:
This PR sets up the usage of the job queues. How it works:

- Run a collection REST api 
- Run a collection worker that listens for new collection jobs on a rabbit mq queue
- Run an enrichment REST api that gets called by the collection worker when it's done
- Run an enrichment worker that listens for new enrichment jobs on a rabbit mq queue
- You can run them all by using `npm run all`

It's about half-way done. The collection part is working (it fetches data from jira and stores it in mongo). The enrichment part isn't done.

I'm not sure how we want to handle the db connections. I'm concerned with the enrichment service using the mongoDB connection, since that really couples it with the collection service. Instead, the enrichment service should probably call the collection service's REST API and fetch all the jira issues.

It's a bit confusing right now. The folder structure could use a bit of work. And how/where we instantiate the mongodb and pass it around is worth thinking about.

Edit:

I've updated this PR. I'm much happier with how the mongo db is used:

- Use the mongo/connection file to create connections and collections
- Only the collection service interacts with the mongo db
- The enrichment service calls the collection service (direct import for now, could be api request later), and the collection service finds issues for the enrichment service.

I've tested out the entire flow, and it's all working right now.


src/collectors/jira/issues.js
```diff
@@ -1,9 +1,19 @@
...
 const config = require('@config/resolveConfig').jira
 
 module.exports = {
+  async collectIssues (db, projectId) {
+    const issues = await module.exports.fetchIssues(projectId)
+
+    const issueCollection = await findOrCreateCollection(db, 'jira_issues')
+
+    // Insert issues into mongodb
+    await issueCollection.insertMany(issues)
```
jaxzhou, line 13: insert would make data duplicate.we should generate '_id' by custom. like a hash from issue_key and jira_host

paulgoertzen-merico: Thanks Jax, you're right. I copied this code from `app.js`. Problems like this are more Justin's concern (which he will deal with in another PR). I'm working on creating the interconnections between the services.

paulgoertzen-merico:
I've updated this PR. I'm much happier with how the mongo db is used:

- Use the `mongo/connection` file to create connections and collections
- Only the collection service interacts with the mongo db
- The enrichment service calls the collection service (direct import for now, could be api request later), and the collection service finds issues for the enrichment service.

I've tested out the entire flow, and it's all working right now.

PR #19, opened 2021-07-21
Use project id instead of project name to fetch issues from mongo

UltimateBeaver:
@paulgoertzen-merico I followed the workflow for computing lead time in the README but found pg didn't have any issues. With some `console.log`, I found the return value of `issueCollector.findIssues` is empty. I was able to fix the issue by using project id as the query parameter instead of the project name.

After this change, both changelog collector started to fetch changelogs and JIRA enricher started writing to `jira_issues` table.


paulgoertzen-merico:
Nice, thanks for this change. Can you update README.md to match?

PR #21, opened 2021-07-21
WIP: gitlab collector project example

kevin-kline:


test/collectors/gitlab/projects.test.js
```diff
@@ -0,0 +1,21 @@
...
+			})
+		})
+		describe('fetchAllProjects', () => {
+			it('Gets more than 0 projects from gitlab', async () => {
+				let projects = await gitlabProjectCollector.fetchAllProjects()
+				assert.equal(projects.length > 0, true)
+			})
+		})
+	})
+})
```
paulgoertzen-merico, line 21: It's usually not great practice to actually call 3rd party APIs in your test suite. Eventually your test suite will fail because the api is down, and it'll cause a whole thing. Normally we just use a mock axios request/response.

src/collection/collectors/util/fetcher.js
```diff
@@ -0,0 +1,20 @@
...
+  },
+  async fetchAll (host, path, modelName){
+    try {
+      const res = await axios.get(`${host}/${path}/${modelName}`)
+      return res.data
+    } catch (error) {
+      console.error(error)
+    }
+  } 
+}
```
paulgoertzen-merico, line 20: I created a file very similar to this called `fetcher` .... didn't I? Can you check for overlap?

kevin-kline: It's in JIRA

src/collection/collectors/gitlab/index.js
```diff
@@ -0,0 +1,15 @@
+const mongo = require('../mongo')
+const projects = require('./projects')
+const projectId = '/28270340'
+const collectionName = 'gitlab_projects'
+
+module.exports = {
+  async collect () {
```
paulgoertzen-merico, line 7: `collect()` might be better if it took `projectId` as an argument

src/collection/collectors/util/mongo.js
```diff
@@ -0,0 +1,17 @@
...
+      await collection.insert(data)
+    } catch (error) {
+      console.error(error)
+    } finally {
+      dbConnector.disconnect(client)
+    }
+  }
+}
+
+module.exports = mongo
```
paulgoertzen-merico, line 17: I think this file is ok, but I'm trying to reconcile it with the `dbConnector` file ... I put a `findOrCreateCollection` function in the `connection` file. Maybe it should go in this one?

PR #22, opened 2021-07-21
clean up mongodb before collection

oddscenes:
Previously we had no mechanism to clean out the mongo db before pulling in Jira issues from the API. This will allow us to work from a clean slate, before pulling the data

- clean the mongodb before writing/enhancing

### DIFF
--- a/src/collection/collectors/jira/issues.js
+++ b/src/collection/collectors/jira/issues.js
@@ -12,6 +12,8 @@ module.exports = {
     try {
       const { issues } = await module.exports.fetchIssues(projectId)
 
+      await db.collection(collectionName).remove()
+
       const issueCollection = await dbConnector.findOrCreateCollection(db, collectionName)
 
       // Insert issues into mongodb


src/collection/collectors/jira/issues.js
```diff
@@ -12,6 +12,8 @@ module.exports = {
     try {
       const { issues } = await module.exports.fetchIssues(projectId)
 
+      await db.collection(collectionName).remove()
```
paulgoertzen-merico, line 15: Do we really want to hard code this in? I feel like this is just a temporary measure. I had something like this as well. 

The more natural way to do this is to put it on line 18, and call `await issueCollection.deleteMany()`.

But I'm not sure we want this. Instead, we should have a mechanism that prevents us from adding records that are already in the db.

oddscenes: @paulgoertzen-merico Agree this is a temporary step to clear out old records but going to refactor into a caching style setup

PR #23, opened 2021-07-21
add enrichment specific details to readme

oddscenes:
With this change we now have both **Collection** and **Enrichment** (new addition) usage documented in readme

- Added enrichment job specific details to the readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,15 +14,15 @@
 
 ## Usage
 
-### Create a Collection or Enrichment Job
+### Create a Collection Job
 
 1. From the terminal, execute `npm run all`
 2. From Postman (or similar), send a request like...
 
-- Collection: `POST http://localhost:3001/`
-- Enrichment: `POST http://localhost:3000/`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,22 +14,24 @@
...
     "projectId": 555555,
     "jira": {
         "projectId": "test-api",
         "accountUri": "merico.atlassian.net"
     }
 }
-
 ```
+
 3. See that the collection job was published, jira collection ran, the enrichment job was published, and jira enrichment ran
```
paulgoertzen-merico, line 35: The README.md is no longer accurate with this change. If the user calls just enrichment, then there will be no collection run. Can you make it clearer? Or make another section for enrichment only?

oddscenes: @paulgoertzen-merico Mainly for testing purposes, to avoid the delay of collecting etc. Updated the placement - how about having it below the collection job details, just as a point?

PR #25, opened 2021-07-21
Integrates the new plugin registry pattern

paulgoertzen-merico:
This PR implements the plugin registry pattern as described by Hezheng in the Dev Lake MVP doc (https://docs.google.com/document/d/19JqqF_-qVBBzbgbWB5ncaNjtETDMfBrLL7zsaLTtfr0)

The plugin registry doesn't do anything yet, it's not being used. But you can see how we configure it, and how it loads the plugins.

I created https://github.com/merico-dev/jira-pond as the other half to this PR. It implements the plugin interface. Please review it. It also does nothing really, it's just the interface.

You may find the plugin config file weird. I took a few different approaches, but this is the one that seemed best to me. It allows for importing multiple collectors or enrichers from the same package (using the `name` property).

### DIFF
--- a/config/plugins.json
+++ b/config/plugins.json
@@ -1,0 +1,12 @@
+[
+  {
+    "package": "jira-pond",
+    "name": "collector",
+    "type": "collector"
+  },
+  {
+    "package": "jira-pond",
+    "name": "enricher",
+    "type": "enricher"
+  }
+]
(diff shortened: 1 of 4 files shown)


paulgoertzen-merico:
@joncodo the build is failing because github can't pull the private `jira-pond` repo when it runs `npm install`. I didn't want to put a secret in our codebase, and it worked for me locally. @joncodo, I don't think I can learn github workflow secret management quickly enough, any chance you could solve this?

Issue #24, opened 2021-07-21
First batch of metrics to support from Gitlab


UltimateBeaver:
@kevin-kline Please see below for a list of metrics we want to support for our MVP for Gitlab:

1. Added lines (during a time window)
2. Removed lines (during a time window)
3. Accumulated lines (the sum of added lines and removed lines during a time window)
4. Number of lines in the repo at a given time
5. Number of commits (during a time window)
6. Number of committers (during a time window)
7. Number of code reviews (during a time window)
8. Number of code review rounds (during a time window)
9. Median code review time (during a time window)
10. Average code review time (during a time window)

I'm not sure if 4 and 8 are easy to support. Please do some research on Gitlab APIs and we can discuss if some of them are hard to implement. Thanks!

anystyle:
Please Use this table。https://airtable.com/invite/l?inviteId=invnlKL0nPCiDE3bF&inviteToken=c578ba0043ee5da8ecfa888b7f14b82ded5d3905adc21bbb56e1841be77d55f7&utm_source=email

UltimateBeaver:
@kevin-kline The metric list for Gitlab has been updated, please see the first post in this thread.

PR #29, opened 2021-07-22
types: add type definition for git(-project)

narrowizard:
I've tried adding some schema definition for git, git-project.  
not sure for which layer(collection or enrichment).  
please tell me if you have any good idea, thanks.  
/cc @UltimateBeaver @kevin-kline 

related to #24


joncodo:
I appreciate the work here but this is not totally compatible with our current code? Can you provide some unit tests so we can see how you intend this to work?

narrowizard:
it is just type definition, without any logic code(we cannot run this). In other words, It's document.  
if accepted(I mean the schema, not the MR), we will create the gitlab collection based on this.

joncodo:
@kevin-kline is working on this part right now for gitlab models for commits and contributors. Kevin, can you use this schema as your base schema for psql records?

joncodo:
This is now out of date

PR #30, opened 2021-07-22
New mongo find update setup

joncodo:
- Have only one db connection per collection task
- Fix the jira issues collection method
- Refactor methods
- Async bulk with Promise.all


db/mongo/connection.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   },
   disconnect: (client) => {
     try {
+			console.log('INFO >>> closing mongo connection')
```
paulgoertzen-merico, line 19: Weird tabbing bro

src/collection/collectors/jira/changelogs.js
```diff
@@ -6,29 +6,25 @@ const fetcher = require('./fetcher')
...
+          issueId: issue.id,
+          ...change
+        }))
       }
-    } catch (error) {
-      console.error(error)
-    } finally {
-      dbConnector.disconnect(client)
     }
+    await Promise.all(promises)
```
paulgoertzen-merico, line 27: Now there's no try catch, and everything will stop if one changelog fails to insert. Is that what you want?

src/enrichment/enrichers/jira.js
```diff
@@ -53,8 +93,8 @@ module.exports = {
       }
     }
 
-    return isDone
-      ? Math.round(leadTime / 1000)
-      : 0
+    return isDone ?
+      Math.round(leadTime / 1000) :
+      0
```
paulgoertzen-merico, line 98: Why change this formatting? Pretty sure the way it was is the standard way

src/enrichment/enrichers/jira.js
```diff
@@ -3,34 +3,74 @@ require('module-alias/register')
...
+    issues.forEach(async issue => {
+      leadTimePromises.push(module.exports.calculateLeadTime(issue, options.db))
+      issuesToCreate.push({
         id: issue.id,
         url: issue.self,
         title: issue.fields.summary,
         projectId: issue.fields.project.id,
-        description: issue.fields.description,
-        leadTime
+        description: issue.fields.description
```
paulgoertzen-merico, line 50: Can you comment out description? Hezheng and I found out yesterday that description is just a crazy object with all kinds of nested stuff. It's gonna take a while to figure out how to work with it.

paulgoertzen-merico (commented):
Overall it's good. I like that the db is being passed as an argument (which we'll need to do to use the plugin pattern). I'm not totally sure the use of Promise.all() is the right choice though, as it will stop all insertion as soon as one fails. I feel like it'd be better to allow the occasional failure for now.

src/enrichment/enrichers/jira.js
joncodo: Must have been my prettier format. I did not intentionally do this

src/enrichment/enrichers/jira.js
joncodo: ok

src/collection/collectors/jira/changelogs.js
joncodo: that is true. I need to think about this one

PR #33, opened 2021-07-22
add a schema to write mongodb

jaxzhou:
NOT TESTED!!!

an abstract model to write data into mongo.

keypoints are:
- add validator while create collection
- declare primary fields. to generate _id. 
- save by _id to avoid duplicate records.


jaxzhou:
closed by conflict

PR #36, opened 2021-07-22
Jira fetch issues pagination combinator

oddscenes:
Currently we will stop out at 50 Jira issues on request, this change will allow us to pull in as many exist

- increase the maxResults to 100 (from 50), the limit Jira API will allow
- Add paginating/concat setup to gather Jira issues before saving all to mongo

### DIFF
--- a/src/plugins/jira-pond/src/collector/issues.js
+++ b/src/plugins/jira-pond/src/collector/issues.js
@@ -44,9 +44,27 @@ module.exports = {
   },
 
   async fetchIssues (project) {
-    const requestUri = `search?jql=project="${project}"`
+    let issues = []
+    let retry = 0
+    const startAt = issues.length > 0 ? issues.length : 0
+    const searchUri = `search?jql=project=${project}`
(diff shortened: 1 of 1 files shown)


joncodo:
This goes and gets all issues no matter what? It loops until there are no more to loop through? Can we unit test this?

oddscenes:
> This goes and gets all issues no matter what? It loops until there are no more to loop through? Can we unit test this?

@joncodo Yeah that is right, `fetchIssues()` calls the data immediately same as it did that before but has a small /basic retry handler (3 failures) if more than that we throw an error.. 

As far as loop logic we're looping through the pages until the issues collected is greater than the total amount, in which case its the last page (`startAt`) and we can stop the concat. Then we just return all of them. And definitely will be possible and good to have a unit test on this one.

oddscenes:
closing this because its integrated into [https://github.com/merico-dev/lake/pull/43](https://github.com/merico-dev/lake/pull/43)

PR #38, opened 2021-07-23
added private token header, added fetch merge requests, added get all…

kevin-kline:
… project repo commits

### DIFF
--- a/src/collection/collectors/gitlab/projects.js
+++ b/src/collection/collectors/gitlab/projects.js
@@ -2,7 +2,8 @@ require('module-alias/register')
 
 const fetcher = require('../util/fetcher')
 const modelName = 'projects'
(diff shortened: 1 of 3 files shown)


src/collection/collectors/gitlab/projects.js
```diff
@@ -2,7 +2,8 @@ require('module-alias/register')
 
 const fetcher = require('../util/fetcher')
 const modelName = 'projects'
-const { gitlab: { host, apiPath } } = require('../../../../config/resolveConfig')
+const { gitlab: { host, apiPath, token } } = require('../../../../config/resolveConfig')
```
joncodo, line 5: you can use @config/resolveConfig. you can set up new directory aliases in the package.json file

PR #39, opened 2021-07-23
no more duplicate rows

joncodo:
This makes a primary key before insert so that no duplicates are stored in mongodb

Screen Shot 2021-07-06 at 11 25 19 AM


joncodo:
Screen Shot 2021-07-23 at 11 12 21 AM

PR #43, opened 2021-07-23
Better changelog fetching

joncodo:
Need to write tests on change logs and figure out why we are inserting duplicate change logs??


UltimateBeaver:
No longer needed since all the work is incorporated in this PR: https://github.com/merico-dev/lake/pull/46

PR #52, opened 2021-07-28
fix total results to show number

oddscenes:
Total amount for `fetchIssues()` in jira collector is showing undefined. Reworked it to return actual number of issues


UltimateBeaver (commented):
What's the difference between these two versions? Aren't they equivalent?

UltimateBeaver:
LGTM

PR #53, opened 2021-07-28
Configuration mapping with new IssueType

joncodo:
## Why we are doing this:

- Users can have different names in Jira, or other systems, than we have referenced in our code. For instance, you may think that when an issue is "done", the issue status is in "QA Review". With this code, you can map your own labels so we can better fit your Jira set up.

- We also wanted issue types captured for Jira issues which is what caused me to work on this. 

## What this involves: 

- Issue types in the pqsl database
- Mapper util file to map a value to a key
- Unit tests

This is the data after you run the code now with issue types populated.

Screen Shot 2021-07-28 at 11 12 00 AM

So you can read about this in CONFIGURATION.md


COFIGURATION.md
```diff
@@ -0,0 +1,66 @@
+# Configuration
```
UltimateBeaver, line 1: The file name is mis-spelled (`COFIGURATION.md`). Please fix it.

README.md
```diff
@@ -29,33 +29,9 @@
 1. Make a copy of `config/local.sample.js` under the name of `config/local.js`
 2. We can use default values for most fields except the Jira section. For how to set up basic authorization with Jira, please see this [section](#jira) below
```
UltimateBeaver, line 30: Since you removed jira section in `README.md`, this is a broken reference now, please fix it.

COFIGURATION.md
```diff
@@ -0,0 +1,66 @@
...
+2. We can use default values for most fields except the Jira section. For how to set up basic authorization with Jira, please see this [section](#jira) below
+
+## Plugin Level Configuration
+
+### Jira Specific String Configuration
+
+This can be set up in `/config/constants.js`.
+
+```
+"jira": {
```
UltimateBeaver, line 15: It seems the following format is what you chose at the end, could you please update the markdown file accordingly?

```
{
  "jira": {
    "mappings": {
      "Closed": ["Done", "Closed", "已关闭"],
      "Bug": "Bug",
      "Incident": "Incident"
    } 
  }
}
```

src/plugins/jira-pond/src/enricher/index.js
```diff
@@ -41,20 +43,28 @@ module.exports = {
...
         ...issue
       }
+      // Create all new records
       creationPromises.push(JiraIssue.findOrCreate({
         where: {
           id: issue.id
         },
         defaults: issue
       }))
+      // Update all existing records
```
UltimateBeaver, line 58: Can we use `upsert` here instead? Also, why do we use different conditions for creation and update (`issue.id` for creation vs `issue.url` for update)?

UltimateBeaver (commented):
Thanks for the PR, Jon. Looking good in general.

Some high-level suggestions besides the comments on the code:

1. The PR description does not provide much context for the reviewer. Here I suggest that we adopt our "Why, What, How" framework. If "Why" is obvious, then at least talk about the "What" and "How" on a high level in the MR description.

2. We need a solution for the linting problem. The goal is to standardize how we do linting locally so we never see the linting changes in PR in the future. I'll create a separate issue for this task.

src/plugins/jira-pond/src/enricher/index.js
joncodo: upsert does not work with findOrCreate() only with find()

This is why I had to do both.

joncodo:
@UltimateBeaver this is ready for review again :)

src/plugins/jira-pond/src/enricher/index.js
UltimateBeaver: If we use `upsert`, then we no longer need `find` or `findOrCreaet`? I thought we can just do something like:

```
upsertPromises.push(JiraIssue.upsert(issue))
```

Please correct me if I'm wrong.

UltimateBeaver:
@joncodo thanks for addressing the feedback, there's only one comment left, it's about the `upsert`.

src/plugins/jira-pond/src/enricher/index.js
joncodo: This does not work with sequelize

PR #54, opened 2021-07-28
chore: change husky permissions to avoid warning

joncodo:
We need this to fix this issue: https://github.com/merico-dev/lake/issues/51#issuecomment-888424313


UltimateBeaver:
LGTM

PR #62, opened 2021-07-29
Collect and enrich commits and projects from gitlab

joncodo:
## Why: 

We need to collect all github issues and projects so we can create metrics in grafana for this data

## What:

Commits are stored in psql with stats of added lines etc.
Projects are stored based on project details API request from gitlab

## How:

1. Fetch from the gitlab API
2. Save in mongo
3. Enrich by fetching from mongo
4. format for PSQL
5. Save in PSQL

Screen Shot 2021-07-29 at 11 43 28 AM


joncodo:
@UltimateBeaver feel free to review this after the fact and I will make a new PR to fix the comments you make here

PR #67, opened 2021-07-29
Gitlab - adding in commit level metrics to grafana dashboard

joncodo:
## Why

We want to show metrics based on user commits to a repo that have been enhanced with data

## What

Here are the metrics this collects

1. Number of contributors
2. Number of commits
3. Removed lines of code
4. Added lines of code
5. Accumulated lines (defined as the sum of added lines of code and removed lines of code during a time window)

## How

- To change this, I had to load up grafana on port 3002 and then I had to make queries for the dashboard and run them.
- I also had to add gitlab_commits.project_id
- set page size of gitlab commits to 100 and not the default 20

Screen Shot 2021-07-29 at 2 48 10 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
 ```
 3. Run `docker-compose ps` to see containers runnning.
 4. Install dependencies with `npm i`
-5. Run migration with `npx sequelize-cli db:migrate`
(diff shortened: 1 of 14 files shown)


UltimateBeaver:
LGTM

PR #68, opened 2021-07-29
Run docker from npm

oddscenes:
use `npm run docker` to startup docker, and force rebuilding/no cache


joncodo:
You need to update the docs bud....

PR #77, opened 2021-07-30
feat: update LeadTime calculation according to standard

klesh:
1. Lead-time calculation
   According to @Yi.Zheng, Lead time is the days between creation and
   resolution, simple subtraction, no other factor. The resolution date
   can be obtained from jira_issue.fields.resolutiondate according to
   @Jax. so, we don't need changelog at this point, deleting it can
   simplify our codebase and lessen performance concern.
2. Collect JIRA data by `boardId`
   In reality, a JIRA Project can be a Product(EE) or Team(EE Backend),
   and a JIRA board can covers issues among single or multiple projects,
   collecting issues by boardId can help dealing with these situations.
3. Collect JIRA data page by page to lower down memory usage footprint
   User may run our product on some huge JIRA database, load all records
   in memory before flushing them to database could cause OOM
4. Collecting/Enriching changed data only
   By filtering out unneeded data we could reduce subsequent collection
   time drastically, by fetching the max updated time from local db and
   send it to JIRA through jql we can gain much faster collection time
   in future collection. Same technique applied to enricher by
   maintaining a `enriched` timestamp, whenever issue got
   enriched, the timestamp got synchronized to `fields.updated`, so we
   can filter out records that don't need to be re-enriched.
5. Refactor pagination fetching to fetcher as a common function
   `fetcher.fetch` should be responsible for retrying logic due to it
   was designed for a specific request. Pagination for JIRA APIs are
   uniform, a shared function for pages iteration could be used for
   different APIs.


db/migrations/20210729104517-issue-id-uniq.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    // This would fail if your table has duplicated id, which doesn't suppose to happen
+    await queryInterface.addConstraint('jira_issues', {
```
UltimateBeaver, line 6: Jira issue id's uniqueness is already guaranteed by the primary key constraint. We don't need another unique constraint.

src/plugins/jira-pond/src/collector/changelogs.js
```diff
@@ -1,71 +0,0 @@
-const {
```
UltimateBeaver, line 1: Could you please explain that we no longer need changelog data for now because we can get the `resolution_date` directly from issue data, and that we'll compute lead time as `resolution_date` - `creation_date`?

UltimateBeaver: And by removing changelog, we have one less performance concern for the MVP.

src/plugins/jira-pond/src/collector/issues.js
```diff
@@ -1,70 +1,34 @@
...
-          upsert: true
-        }))
-      })
-
-      await Promise.all(promises)
-    } catch (error) {
-      console.error(error)
+  async collectByBoardId(db, boardId, forceAll) {
+    const issuesCollection = await findOrCreateCollection(db, collectionName)
+    const latestUpdated = await issuesCollection.find().sort({ 'fields.updated': -1 }).limit(1).next()
```
UltimateBeaver, line 19: Could you please add to the PR description that it implements incremental update for Jira issues by leveraging `fields.updated`?

src/plugins/jira-pond/src/enricher/index.js
```diff
@@ -1,87 +1,61 @@
...
+          title: issue.fields.summary,
+          projectId: issue.fields.project.id,
+          leadTime: null
+        }
+        // by standard, leadtime = days of (resolutiondate - creationdate)
+        if (issue.fields.resolutiondate) {
+          enriched.leadTime = dayjs.duration(dayjs(issue.fields.resolutiondate) - dayjs(issue.fields.created)).days()
         }
+        try {
+          await JiraIssue.create(enriched)
```
UltimateBeaver, line 45: Can we replace this `create` and `update` with a call to `upsert`?

src/plugins/jira-pond/src/enricher/index.js
```diff
@@ -1,87 +1,61 @@
...
 
-  async enrichLeadTimeOnIssues (rawDb, enrichedDb, projectId) {
+  async enrichIssues (rawDb, enrichedDb, forceAll) {
+    const issueCollection = await findOrCreateCollection(rawDb, 'jira_issues')
     const { JiraIssue } = enrichedDb
+    // filtering out portion of records that need to be enriched
+    const curosr = (
+      forceAll
+        ? issueCollection.find()
+        : issueCollection.find({ $where: 'this.enriched < this.fields.updated || !this.enriched' })
```
UltimateBeaver, line 26: Please add to the PR description that issue enrichment also supports incremental update now.

UltimateBeaver (commented):
Looking good in general, please see the individual comments. Thanks!

README.md
```diff
@@ -59,14 +59,18 @@ POST http://localhost:3001/
 
 {
     "jira": {
-        "projectId": "10003",
-        "accountUri": "merico.atlassian.net"
```
joncodo, line 63: How do you know which gitlab account to call? Or which project Id? Board 8 can exist in many places

README.md
```diff
@@ -59,14 +59,18 @@ POST http://localhost:3001/
...
     "gitlab": {
         "projectId": 19688130
     }
 }
 
+```
+    Or, by using `curl`
+```sh
+curl -X POST "http://localhost:3001/" -H 'content-type: application/json' \
+    -d '{"jira":{"boardId": 8}}'
```
joncodo, line 73: Same

README.md
klesh: I saw nowhere in original code refering to this field.
And this is for jira data collection, I think you were wondering about **jira** account?
If that was the case, the sample value of `jira.host` from `local.sample.js` says `https://your-domain.atlassian.net` which is sufficient to indicate which account we are requesting for.
So, I think it was a outdated documentation.

README.md
klesh: Same too :)

PR #80, opened 2021-07-30
refactor: reorganize gitlab enricher to match its collector

klesh:
Other improvement:
1. Move functions to top level to support IDE function jumping/peeking
2. Enrich data with mongodb cursor for lower memory consumption
3. Fix `findOrCreateCollection` exception swallowing issue by removing
   `try/catch` statement. Error-printing on main function is sufficient
   `try/catch console.log` everywhere is tedious and error-prone


src/plugins/gitlab-pond/index.js
```diff
@@ -39,8 +34,8 @@ if (require.main === module) {
   (async function() {
     const { db, client } = await dbConnector.connect()
     try {
-      await module.exports.collector.exec(db, { projectId: process.argv[2] })
-      // await module.exports.enricher.exec(db, enrichedDb, { projectId: process.argv[2] })
+      await module.exports.collector.exec(db, { projectId: 24547305 })
```
joncodo, line 37: We can't hardcode this

klesh: This is for debugging purpose, notice the `require.main === module` condition, those code wouldn't be executed in production environment.

PR #82, opened 2021-07-30
fix: removed grafana internal db things from working db

oddscenes:
We actually don't need to specify the database + datasource for Grafana. This is what's causing all the internal Grafana DB stuff to populate the `lake` postgres db.

**Updated:**
- Removed Grafana database env variables

**Fixes:** 
- Stops cluttering the working db with Grafana internal db things that we don't really need

**Note**
- It's possible in the future if we want this Grafana internal db stuff back, we can just make another db in `docker-compose.yml`. Then populate it. As of now though, I don't see any use for it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,12 +48,5 @@ services:
       - postgresdb
     command: sh -c './scripts/wait-for postgresdb:5432 -- grafana'
     restart: always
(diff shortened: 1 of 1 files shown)


joncodo:
This is duplicated work. Klesh already did this

PR #93, opened 2021-07-30
chore(npm): we need a way to run without DEBUG when desired

joncodo:
Concurrently seems to not allow you to run something like `DEBUG=true npm run all` so we need `npm
run all` and `npm run all-debug`. The reason to have DEBUG is so we can comment out console logs
with `process.env.DEBUG && console.log()` so we don't remove console logs needed by others.


UltimateBeaver:
LGTM

Issue #95, opened 2021-07-30
Too many commits seem to be pulled in for gitlab pond collector


joncodo:
```
async collectByProjectId (db, projectId, forceAll) {
    const commitsCollection = await findOrCreateCollection(db, collectionName)
    for await (const commit of fetcher.fetchPaged(`projects/${projectId}/repository/commits?all=true&with_stats=true`)) {
      commit.projectId = projectId
      console.log('JON >>> saving commit to mongo')
      await commitsCollection.findOneAndUpdate(
        { id: commit.id },
        { $set: commit },
        { upsert: true }
      )
    }
  }
```

I sent a post with: 

```
{
    "jira": {
        "boardId": 8
    },
    "gitlab": {
        "projectId": 8967944
    }
}
```

This is for vdev.co

<img width="339" alt="Screen Shot 2021-07-30 at 2 37 29 PM" src="https://user-images.githubusercontent.com/3011407/127691074-d7905560-0c36-46f3-8662-216ffee9d904.png">

I expected to see only 15k commits in my mongo db but I saw many more. I stopped the process when it hit 350 pages long

<img width="1260" alt="Screen Shot 2021-07-30 at 2 38 13 PM" src="https://user-images.githubusercontent.com/3011407/127691156-64acab52-1d23-4861-837b-d6aa27caa28e.png">

## EXPECTED: 

- I should get the exact number of commits in the repo stored in mongo

## Questions

- Is this maybe all commits from all branch even if they are not merged?

joncodo:
So it seems that when you set `?all=true` in the request, it pulls down duplicate commits.

This is a duplicate commit for the repo ce-backend: 20103385

<img width="980" alt="Screen Shot 2021-07-30 at 3 13 43 PM" src="https://user-images.githubusercontent.com/3011407/127695018-491a3591-35c2-4a72-bd59-38aebfc71341.png">

I am removing the all=true

joncodo:
@klesh I am closing this but you should read it

klesh:
> ```
> async collectByProjectId (db, projectId, forceAll) {
>     const commitsCollection = await findOrCreateCollection(db, collectionName)
>     for await (const commit of fetcher.fetchPaged(`projects/${projectId}/repository/commits?all=true&with_stats=true`)) {
>       commit.projectId = projectId
>       console.log('JON >>> saving commit to mongo')
>       await commitsCollection.findOneAndUpdate(
>         { id: commit.id },
>         { $set: commit },
[...]

PR #96, opened 2021-07-30
fix(gitlab collection): fix pagination, table types, mongo names

joncodo:
1. There was an issue with the fetchPaginated where we were using `?all=true`. I removed that since
it was not giving the correct number of issues in the default branch in gitlab. If you use all=true,
it seems to get about double the commits. I have a theory that this is caused by teams that use
rebase in gitlab. Gitlab seems to save the commits that were rebased with a diffdrent parent so the
ids of the commmits are different. see #95 

2. I also changed the type of the gitlab postgres models to TEXT instead of STRING since some descriptions and title of our data were too long. There is a migration for this and you have to run `npx secquelize db:migrate`

3. I also changed the names of the mongo collections back to `gitlab_commits` and `gitlab_merge_requests` as approved by @UltimateBeaver . This was needed since the collection name in the collectors and enrichers were different. The new names match the Postgres table names defined in `/db`

4. I also changed all the gitlab code for project ids and converted them to projectId.

### DIFF
--- a/db/migrations/20210730182134-change-string-to-text-gitlab.js
+++ b/db/migrations/20210730182134-change-string-to-text-gitlab.js
@@ -1,0 +1,26 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.changeColumn('gitlab_commits', 'message', {
+      type: Sequelize.TEXT,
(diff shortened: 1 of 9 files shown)


UltimateBeaver:
LGTM

PR #99, opened 2021-07-30
feat(gitlab collection for notes): to collect and enrich notes from G…

kevin-kline:
…itLab API

Includes config flags to skip collectors, migration, model, collector, and enrichment function for
GitLab Notes on Merge Requests.

Fixes: https://github.com/merico-dev/lake/issues/86

Why: 
- We want to track comments on merge_requests

What: 
- skips collection and enrichment of certain things based on config (optional)
- add first_comment_time to merge_requests (This is based on notes. Find the earliest note and use that created_at )
- collects all notes
- enriches all notes that are not system notes
- notes model and migrations


UltimateBeaver:
Replaced by #125

PR #117, opened 2021-08-02
feat: plugins initialization before all workers

klesh:
1. Support plugin configuration injection
   Plugin can declare a default configuration on its main module, which could be overridden by 
   `config/plugins.js`. We should move their configurations from `local.js` to `plugins.js` file
2. Collectors now execute in parallel, so do enrichers
   Our collectors have no dependency to each other at this point, all plugins we specified in 
   request body will be executed in parallel
3. Ensure correct data types of parameters(boardId/projectId) on root functions `collect/enrich`
   Once and for all
4. `axios` timeout mitigation for `jira-pond`
5. Mitigation for scenario that jira data collecting procedure got interrupted(Ctrl-C) with sorting
   jira issues from API by `updated` field

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,15 +62,20 @@ POST http://localhost:3001/
         "boardId": 8
     },
     "gitlab": {
-        "projectId": 19688130
+        "projectId": 8967944
     }
 }
 
 ```
     Or, by using `curl`
 ```sh
+# ee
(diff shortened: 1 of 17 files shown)


UltimateBeaver:
LGTM

PR #121, opened 2021-08-03
Initial TypeScript MonoRepo Projects

jaxzhou:
Two Apps Initial

- Queue: the task execute queue service.
- Rest: the Restful API service

Plugins Initial

- The executable tasks would executed in queue service


UltimateBeaver:
LGTM

PR #124, opened 2021-08-03
test(tests): fetch paged is now tested

joncodo:
@oddscenes @kevin-kline @klesh @UltimateBeaver 

This is an example of sinon and testing a pretty complex thing. The first step is to decide which parts of your code you need to test and which parts you don't. 

In this case, I assume that axios.get works and I don't want my unit tests going to the web. However, I want to test the results if the web returns the next page response.


joncodo:
There is a unit test failing here since I am missing the notes.js in enricher for gitlab

```
Error: Cannot find module './notes'
Require stack:
- /Users/jonathanodonnell/Code/merico/lake/src/plugins/gitlab-pond/src/enricher/notes.test.js
```

I think once we merge some other PRs, this will resolve itself

joncodo:
Here are my local tests passing for this:

Screen Shot 2021-08-03 at 5 11 20 PM

klesh:
I suggest that we put all test files into `test` folder, it's more widely adopted, easier to exclude when packaging like building docker image

joncodo:
> I suggest that we put all test files into `test` folder, it's more widely adopted, easier to exclude when packaging like building docker image

The only problem I have with that is maintaining a full folder structure. Its easy to exclude *.test.js in builds and then we don't have to use require('../../../../../../src/file.js)

I really think its easier this way but if feel really strongly about a /test folder, we can do that. Let me know :)

klesh:
ok, Iet's go with your way.

kevin-kline:
This PR contains mostly tests

- mock sample data
- a string util function to convert camel case to snake case
- a test to make sure paging works
- a sample config file for testing
- an npm script for testing plugins

- tests for mapResponseToSchema:
  - commits
  - merge requests
  - notes

These three tests check that when we pass raw data object from our DB to the "mapResponseToSchema" function, it returns an object with the number of fields that match with our Postgres Model dynamically. This means if the model changes, the test will break and we will know we have to fix the mapResponseToSchema function.

joncodo:
@UltimateBeaver this is ready to merge I think :)

joncodo:
I think we should close this PR and not merge it in for now. We are moving to TS soon and this is too big. We can just use this as a reference for future testing as needed. Please re-open if you want @UltimateBeaver

PR #127, opened 2021-08-04
feat: deployment and security enhancement

klesh:
## Why

To increase the security of a hosted instance of dev lake (right now, all databases have default passwords and are publicly accessible).

## What

1. cleanup `docker-compose.yml`
   1. mongodb requires authentication now
   2. unify `environment` declaration format
   3. increase complexity of default passwords a little bit
   4. all services except Grafana bind to localhost by default for security reason
   5. all services now run within isolated network `lake`
   6. all services now `restart: always` so they can be fired up at system boot automatically
   6. remove redundant field such as `hostname` and `container_name`
2. set collection/enrichment server host binding by EnvVar

## [Important!] Steps to migrate

1. shutdown services: `docker-compose down`
2. remove old data for new passwords to kick in: `sudo rm -rf data postgres-data`
3. fire up services: `docker-compose up -d --build`
4. observe service log by its name, like `docker-compose logs -f mongodb`
5. run migration for pg: `npx sequelize-cli db:migrate`


config/database.json
```diff
@@ -1,7 +1,7 @@
 {
   "development": {
     "username": "postgres",
-    "password": "postgres",
+    "password": "postgresWhat",
```
joncodo, line 4: :D - Let's use a random string instead just in case someone tries to guess it

config/local.sample.js
```diff
@@ -3,15 +3,14 @@ module.exports = {
     token: 'mytoken'
   },
   mongo: {
-    connectionString: 'mongodb://localhost:27017/test?retryWrites=true&w=majority'
+    connectionString: 'mongodb://lake:lakeIScoming@localhost:27017/lake?authSource=admin'
```
joncodo, line 6: let's use a stronger password

config/local.sample.js
```diff
@@ -3,15 +3,14 @@ module.exports = {
...
   },
   rabbitMQ: {
-    connectionString: 'amqp://guest:guest@localhost:5672/rabbitmq'
+    connectionString: 'amqp://guest:guestWhat@localhost:5672/rabbitmq'
```
joncodo, line 9: same

config/local.sample.js
```diff
@@ -3,15 +3,14 @@ module.exports = {
...
   },
   postgres: {
-    connectionString: 'postgres://postgres:postgres@localhost:5432/lake',
     username: 'postgres',
-    password: 'postgres',
+    password: 'postgresWhat',
```
joncodo, line 13: same.

config/local.sample.js
```diff
@@ -40,8 +39,5 @@ module.exports = {
   },
   jiraBoardGitlabProject: {
     8: 8967944
-  },
-  enrichment: {
-    connectionString: 'http://localhost:3000/'
```
joncodo, line 45: Did you mean to remove this?

docker-compose.yml
```diff
@@ -1,46 +1,56 @@
...
-  postgresdb:
-    hostname: postgresdb
+  mongodb:
+    image: mongo:5.0.0
     environment:
-      POSTGRES_DB: lake
-      POSTGRES_PASSWORD: postgres
-      POSTGRES_USER: postgres
+      - MONGO_INITDB_ROOT_USERNAME=lake
+      - MONGO_INITDB_ROOT_PASSWORD=lakeIScoming
```
joncodo, line 7: same

docker-compose.yml
```diff
@@ -1,46 +1,56 @@
...
+    ports:
+      - 127.0.0.1:27017:27017
+    restart: always
+    networks:
+      - lake
+  postgresdb:
     image: postgres:13
+    environment:
+      - POSTGRES_DB=lake
+      - POSTGRES_PASSWORD=postgresWhat
```
joncodo, line 19: same

grafana/grafana.config
```diff
@@ -1,3 +0,0 @@
-# Config Options
-GF_USERS_ALLOW_SIGN_UP=false
```
joncodo, line 2: Why did we remove this? Does @oddscenes need this?

run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+docker-compose up -d
+
+npm i
+npx sequelize-cli db:migrate
+
+export ENRICHMENT_PORT=43000
```
joncodo, line 8: I think these belong in a config file. I also think this does all the steps but I don't want to always do all the steps. Sometimes, I don't want to migrate if I am working on a new migration. I think you can keep this locally but it may be confusing to new users? What do you think @UltimateBeaver ?

docker-compose.yml
joncodo: Also, since we are updating a bunch of passwords, we need a README entry for users to know how and where to set their own passwords for if they want to host and run this securely. Maybe a general doc on security is a good idea

joncodo:
@klesh overall looks great! These are just small things

PR #129, opened 2021-08-04
refactor(gitlab collection): removing the 'skip' flags from collection

kevin-kline:
Skipping collection causes more problems than it solves. Better to collect all at once, and maybe
later implement a configuration by API request.


UltimateBeaver:
@kevin-kline Thanks for the PR, I think it's fine to keep the code. We'll implement this functionality more formally later.

PR #130, opened 2021-08-04
feat: allow users to configure which branch to fetch for gitlab repos

joncodo:
## WHY

Users sometimes do not work off of the main branch on gitlab. The main branch could be 'main' but all the current work is on the 'development' branch. 

## WHAT

You can now send a post request to localhost:3001 with this body:

```
{
    "gitlab": {
        "projectId": 19688130,
        "branch": "jon-test"
    }
}
```

Where "branch" is optional. It defaults to the default branch of gitlab.

Here is the terminal output calling this api

Screen Shot 2021-08-04 at 9 23 19 AM

Here is the record in mongo locally

Screen Shot 2021-08-04 at 9 30 57 AM

Here is the branch I tested in gitlab

Screen Shot 2021-08-04 at 9 31 11 AM

Here is the record in psql

Screen Shot 2021-08-04 at 9 32 53 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,8 @@ POST http://localhost:3001/
         "boardId": 8
     },
     "gitlab": {
-        "projectId": 8967944
+        "projectId": 8967944,
+        "branch": "<your-branch-name>", // branch is optional, we fetch Gitlab default branch if this arg is absent
     }
 }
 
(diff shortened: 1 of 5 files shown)


UltimateBeaver:
LGTM

PR #131, opened 2021-08-04
Move grafana config readme

oddscenes:
- Grafana data source connection info should go in configuration readme with other config
- Duplicate info is also removed from main `readme.md` 

As I'm working on some other readme stuff, I'm considering Grafana and other collection/enrichment tasks as "core" functionality (hence the title). Jira/gitlab etc are "plugin" functionality. I think this split will be useful for scaling the docs


UltimateBeaver:
LGTM

PR #132, opened 2021-08-04
Docker compose all services

joncodo:
## WHY

We want to cater to two types of users: Developers and End Users. An End User is a user who only wants to run our application and see results. They are not interested in making changes to the code. 

## What

Since we run lake in docker compose, it has to wait for all other services to be ready in order to run. This is why lake has `restart: always` in the docker compose file. However, since we use the service called concurrently, process.exit() does not work by itself. I had to run concurrently in its own file so I could handle if a child process exits.

So now the code tries to connect to mongo and rabbit and psql but if any of the services times out, the application exits and docker-compose tries to restart it. This is similar to how grafana works.

I also had to not use nodemon since that will always try to restart things on its own. 

For the developer, there is another docker compose file. Since the developer does not want to run lake in docker.

## How

You can see the new Dockerfile for lake development

See Readme changes for how this can be run


joncodo:
missing some clean up that I had locally and need to still add

joncodo:
@UltimateBeaver this is now ready for merge

config/docker.js
```diff
@@ -0,0 +1,35 @@
+module.exports = {
```
UltimateBeaver, line 1: Do we use this file anywhere?

docker-compose.development.yml
```diff
@@ -0,0 +1,48 @@
+version: "3"
```
UltimateBeaver, line 1: I suggest we keep this file's name as `docker-compose.yml` and use `docker-compose.prod.yml` for the other one.

docker-compose.yml
```diff
@@ -1,5 +1,17 @@
 version: "3"
```
UltimateBeaver, line 1: Please rename this to `docker-compose.prod.yml`

docker-compose.development.yml
UltimateBeaver: Also, please use the `docker-compose.yml` on the latest main, it has some changes we want (via rebase).

docker-compose.yml
```diff
@@ -12,6 +24,7 @@ services:
     volumes:
       - ./postgres-data:/var/lib/postgresql/data
   rabbitmq:
+    hostname: rabbitmq
```
UltimateBeaver, line 27: hostnames are not needed.

package.json
```diff
@@ -4,19 +4,23 @@
   "description": "The data lake for dev.",
   "main": "test-docker-compose.js",
   "scripts": {
-    "start": "node app.js",
+    "start": "docker-compose up -d --build --force-recreate",
+    "logs": "docker-compose logs",
+    "stop": "docker-compose down",
+    "migrations": "npx sequelize-cli db:migrate",
+    "dev": "docker-compose -f docker-compose.development.yml up -d --build --force-recreate",
```
UltimateBeaver, line 11: `docker-compose.development.yml` => `docker-compose.yml`

src/plugins/index.js
```diff
@@ -63,17 +63,20 @@ registryConfig.forEach((item, index) => {
...
+
     // assuming no dependencies among plugins
     await Promise.all(
       Object.values(plugins)
         .filter(plugin => plugin.initialize)
-        .map(plugin => plugin.initialize(rawDb, enrichedDb, plugins))
+        .map(plugin => plugin.initialize(connection.rawDb, enrichedDb, plugins))
     )
+  } catch (error) {
+    console.log('ERROR: connecting to mongodb', error)
```
UltimateBeaver, line 78: Should we also `process.exit()` if we can't connect to pg?

package.json
```diff
@@ -4,19 +4,23 @@
   "description": "The data lake for dev.",
   "main": "test-docker-compose.js",
   "scripts": {
-    "start": "node app.js",
+    "start": "docker-compose up -d --build --force-recreate",
```
UltimateBeaver, line 7: Please add `-f docker-compose.prod.yml`

README.md
```diff
@@ -7,46 +7,27 @@
...
 1. Clone this repository
-1-1. Create the rabbitMQ log directory: `mkdir -p /tmp/rabbitmq/logs/`
-2. From the newly cloned repo directory, run `docker-compose up --build`
-
-    > NOTE: If you get an error like this:
-
-    >"Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /tmp/rabbitmq/etc/"
-
-    >You can fix it by adding the directory in the terminal:
+2. Make a copy of `config/local.sample.js` under the name of `config/local.js`
```
UltimateBeaver, line 13: Please cherry pick this section to Justin's latest readme on main

UltimateBeaver (commented):
Looking good! Please address the comments so we can merge this.

config/docker.js
joncodo: This is the config file used for the docker container within the docker compose file. Notice the host names are set to something other than 'localhost' since docker compose assigns IP addresses to all services within the docker compose network.

docker-compose.yml
joncodo: I actually do need them for the config/docker.js to pick up the host to point to for the lake services.

config/docker.js
joncodo: Screen Shot 2021-08-05 at 8 46 43 AM

This even happens when I set it on the same network

Screen Shot 2021-08-05 at 8 47 04 AM

README.md
joncodo: Done

joncodo:
So the final problem here is working with docker-compose networks. 

**Summary of problem**
So in my docker-compose.prod.yml file, I have added a new service called 'lake'. This is the service that runs four node processes with concurrently. You can test this with 'npm start'. However, the 'lake' service can't connect to the other docker-compose services such as mongo, psql and rabbitmq. 

**Assumptions**
- network: lake is needed for our app
- hostname should not be set
- network should just make everything communicate over localhost but it seems not to.

**Here is what I have tried:**
- Removing hostname from each service and setting network instead. Setting the connections to (localhost|127.0.0.1|<service_name>) in the config/docker.js file

**What I know works**
- If I set a hostname on each service, it works properly.

**Research links:**
- https://stackoverflow.com/questions/38088279/communication-between-multiple-docker-compose-projects
- https://docs.docker.com/engine/reference/commandline/network_connect/
- https://docs.docker.com/compose/networking/

joncodo:
I think this configuration is the closest I have come:

**docker.js**
```
mongo: {
    connectionString: 'mongodb://lake:lakeIScoming@mongodb:27017/lake?authSource=admin'
  },
  rabbitMQ: {
    connectionString: 'amqp://guest:guestWhat@rabbitmq:5672/rabbitmq'
  },
  postgres: {
    username: 'postgres',
    password: 'postgresWhat',
    host: 'postgresdb',
    database: 'lake',
    port: 5432,
    dialect: 'postgres'
  },
```

**docker-compose.prod.yml**
```
services:
  lake:
    container_name: lake
    restart: always
    build: .
    ports:
      - 3000:3000
      - 3001:3001
    networks:
      - lake-network
```

**Logs:**
Screen Shot 2021-08-05 at 1 14 14 PM

Which shows that the container is trying to access the internal port of: `172.21.0.4:5672`

klesh:
Hi, @joncodo 
1. network is for isolation, so services inside our docker-compose can only be accessed by services within same network (on docker virtual network context), so you can create multiple networks on the same docker-compose and they are isolated to each other. By creating a network, we isolated our services from other docker containers on machine level.
2. hostname need not be set unless it's different to service name which is not recommended
3. network doesn't do that, need to access service by its name inside a container, each service is a host to other containers

Say we have a docker-compose like this:
```yml
version: "3"
services:
  postgresdb:
    ports:
      - 127.0.0.1:1111:5432
    networks:
      - lake
  myapp:
    ports:
      - 3000:5432
    networks:
      - lake
```
now, 
1. `myapp` can connect to `postgres` with connection string `postgres://postgresdb:5432/dbname` via network `lake`
2. any one on the LAN can access `myapp` because it's binded to `0.0.0.0` (omitted)
3. any one on the LAN can NOT access `postgres` but you (on your machine), because it's binded to `127.0.0.1`

joncodo:
The resolution was that I needed to set these in the Dockerfile

```
ENV NODE_ENV=docker
ENV ENRICHMENT_HOST=0.0.0.0
ENV COLLECTION_HOST=0.0.0.0
ENV ENRICHMENT_PORT=3000
ENV COLLECTION_PORT=3001
```

PR #133, opened 2021-08-04
docs(how to add metrics): just a short readme to add metrics

kevin-kline:
WIP, first draft. Looking for feedback, please! 

(I know it is not fully complete, but I'd like to get some direction about general feel, formatting, and what level of detail to pursue.

My instinct is not to get too detailed with the technicalities because:
a) They may change
b) Contributors should feel free to do it their own way


UltimateBeaver:
@kevin-kline I'll merge this one now and we'll revise it later (since we're gonna switch to the new architecture soon).

PR #135, opened 2021-08-05
fix: `authored_date` and `committed_date` of `gitlab_commits` show be…

### LINKED ISSUE #113: The type of `authoredDate` and `committedDate` should be `DATE` instead of `VARCHAR`
These two field need to be aggregated in many scenarios, It's hard and inefficient to aggregated on `VARCHAR` when they should be `DATE`

klesh:
… DATE instead of VARCHAR

Closes #113


UltimateBeaver:
LGTM

PR #146, opened 2021-08-06
feat: allow the migrations to be run while running the app in prod mode

joncodo:
## WHY

We are trying to reduce the number of steps that a new user needs to do when they attempt to run the application for the first time.  Fixes: https://github.com/merico-dev/lake/issues/145

## WHAT

All we need to do here is to run the migrations when the `npm run compose-prod` command is run.

## HOW

In order to do this, the Dockerfile has `ENV NODE_ENV=docker`. This means that sequelize will look for a file called database.json and then look for a key called `docker`. This is why I had to add it with the `"host": "postgresdb",` so sequelize could connect to the correct port within docker compose.

You can see the docker logs that show this working here:

Screen Shot 2021-08-06 at 9 40 12 AM

And this is a result of modifying the `npm run lake-prod` command.

Screen Shot 2021-08-06 at 9 28 50 AM

Which is run from the Dockerfile as `CMD ["npm", "run", "lake-prod"]`

### DIFF
--- a/config/database.json
+++ b/config/database.json
@@ -6,5 +6,13 @@
     "database": "lake",
     "port": 5432,
     "dialect": "postgres"
+  },
+  "docker": {
+    "username": "postgres",
+    "password": "postgresWhat",
+    "host": "postgresdb",
+    "database": "lake",
+    "port": 5432,
+    "dialect": "postgres"
   }
 }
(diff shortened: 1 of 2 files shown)


UltimateBeaver:
LGTM

PR #155, opened 2021-08-09
fix: incorrect data-source of new dashboards

klesh:
The `datasource` needs to be `Postgres` instead of `PostgreSQL2`.


UltimateBeaver:
LGTM

PR #156, opened 2021-08-09
fix: make provision datasource editable

### LINKED ISSUE #138: Default datasource is uneditable
Should we consider the fact that user might change database password afterward, if so, this is problematic.
image

klesh:
Closes #138


UltimateBeaver:
LGTM

PR #164, opened 2021-08-09
fix(gitlab collection for commits): fixed query parameter

kevin-kline:
Change from withStats to with_stats is necessary to get additions / deletions / total from gitlab commits API.


UltimateBeaver:
@kevin-kline So right now we don't get any additions / deletions data because the query parameter is incorrect?

UltimateBeaver:
Screen Shot 2021-08-09 at 3 26 36 PM
Screen Shot 2021-08-09 at 3 26 22 PM


Verified with postman, going to merge this PR.

PR #165, opened 2021-08-09
chore: removed unused wait-for script

oddscenes:
the `wait-for` script was added as a workaround for Grafana startup issues within docker-compose. But this issue is resolved via `restart: always` config setting in docker-compose settings.

- Removed the `/scripts/wait-for` file


UltimateBeaver:
LGTM

PR #167, opened 2021-08-09
Auto run collection loop

joncodo:
## WHY

We want new users to only have to run a very few set of commands in order to see results.

## WHAT

Before this PR, users will have to send post requests manually to fetch the data they need. Once this is merged, Users only have to set a few values in their config file and then run application. 

## HOW

There is a new file called concurrently.js. This file runs all our processes. A new process was added called `src/dataCollectionLoop.js`. This file runs when the application is started and every 60 minutes after that. (Can be set in config as well.) The request is sent as a post request to the enrichment api service.

## Further improvements that I did not see as necessary at this stage

- Right now, I only allow you to set one jira board and one gitlab project id in the config. An improvement would be to take in an array in case users want to automatically run several projects at once.

## How to test this

1. Get a clean env. (clone fresh)
2. run `npm start`
3. Check the docker logs to collection and enrichment to happen
4. Look in mongo to see records added automatically
5. Look in PG for enriched records to be added automatically
6. You can then set your config.lake.loopIntervalInMinutes to something smaller than 60 so see the loop happen again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,13 +55,8 @@ Contributing | How to contribute to this repo | [Link](CONTRIBUTING.md)
 3. Configure settings for services & plugins by editing the newly created `config/docker.js`. For how to configure plugins, please refer to the [plugins](#plugins) section.
 4. Start the service with `npm start`
     > you can stop all docker containers with `npm run stop`
-5. Send a post request to the service
-```
-curl -X POST "http://localhost:3001/" -H 'content-type: application/json' \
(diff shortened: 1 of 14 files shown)


joncodo:
@Jonathan ODonnell In your loop PR, could you please also test the scenario where user starts the loop without configuring the docker.js first? And in that case, some guidance or warning that explains to the users what needs to be done would be very helpful

UltimateBeaver:
@joncodo I made some minor adjustments. Now it's ready to merge.

PR #169, opened 2021-08-09
fix: fixed type in migration file for gitlab commits

oddscenes:
### WHY
A few charts in Grafana were not showing up due to `authored_date` column being a string in `gitlab_commits` table. We needed this to be a date type.

### WHAT
- Updated a migration file so now when **running a migration** the postgres db will update the correct type for the column. On initial setup, it should be the correct (date) type

**Delivery Quality (dashboard) - panels 6 and 7 now working:**
Screen Shot 2021-08-09 at 5 22 19 PM

### DIFF
--- a/db/migrations/20210729134703-add-gitlab-commits.js
+++ b/db/migrations/20210729134703-add-gitlab-commits.js
@@ -25,7 +25,7 @@ module.exports = {
         field: 'author_email'
       },
       authoredDate: {
-        type: Sequelize.STRING,
+        type: Sequelize.DATE,
         field: 'authored_date'
       },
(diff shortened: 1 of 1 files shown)


UltimateBeaver:
@oddscenes This is actually not the cause of the issue. We've already changed that column to date type in a later migration file, see here: https://github.com/merico-dev/lake/blob/main/db/migrations/20210804073648-commit-dates.js

The cause is that we didn't pass the right query parameter to Gitlab so we're not getting additions and deletions data back. It's fixed in https://github.com/merico-dev/lake/pull/164.

Issue #159, opened 2021-08-09
[Feat][ts] use redis pub sub to emit service event


jaxzhou:
a message broker would be helpful, and we have the Redis for task queue
instead of import some message queue service Redis is easier and cheeper 

## HOW

```
// app.module.ts
@Module({
  imports: [
    EventsModule,
    SampleService,
   ObserverService
  ],
  providers: [],
})
export class AppModule {}

//sample.service.ts
@Injectable()
export class SampleService {
  constructor(private eventsService: EventsService) {}

  async startTask(): Promise<void> {
        const info = this.doSomeTask();
        this.eventsService.emit('Sampe:TaskStart', info);
  }
}
//observer.service.ts
@Injectable()
export class ObserverService {
  constructor(private eventService: EventsService) {
     this.eventsService.on('Sampe:TaskStart', this.handleTaskStart)
  }

  async handleTaskStart(info): Promise<void> {
     console.info('the task is start', info);
  }
}

```

## Tests

```
\\e2e tests
\\ add redis mock
jest.mock('ioredis', () => {
  return {
    default: require('ioredis-mock/jest'),
  };
});

  beforeEach(async () => {
    const moduleFixture: TestingModule = await Test.createTestingModule({
      imports: [EventsModule],
    }).compile();

    app = moduleFixture.createNestMicroservice(moduleFixture);
    await app.init();
    const sub = app.get('REDIS_SUB_CLIENT');
    const subSyncClient = sub.createConnectedClient();
    const pub = app.get('REDIS_PUB_CLIENT');
   //change the pub publish method with mocked client publish
    jest.spyOn(pub, 'publish').mockImplementation((env, message) => {
      return subSyncClient.publish(env, message);
    });
  });

 it("the event e2e tests", (done) => {
   // mock the callback func to  validate the event data
    const mockedFuc = jest.fn().mockImplementation((value) => {
      expect(value).toEqual({ ut: 'test' });
      //use done to make sure function be called
      done();
      return;
    });
})

```

joncodo:
@jaxzhou I am closing this as a duplicate of #154 . Feel free to re-open and tell me how they are different.

PR #172, opened 2021-08-10
feat: add events service to make event broadcast between app instances

jaxzhou:
#159 

## WHY

the services should be scalable and multi services instance would started. Like queue service.
the task running in the queue service, and we need to send event task events in the cluster, so one of the instance would process the event.
a message broker could help us to broadcast event message in cluster.

## WHAT

Redis has the Pub/Sub used as a message broker. so we may not need any other message broker service. that would reduce the service running cost. because we already has redis for task queue.

## HOW

```
// app.module.ts
@Module({
  imports: [
    EventsModule,
    SampleService,
   ObserverService
  ],
  providers: [],
})
export class AppModule {}

//sample.service.ts
@Injectable()
export class SampleService {
  constructor(private eventsService: EventsService) {}

  async startTask(): Promise<void> {
        const info = this.doSomeTask();
        this.eventsService.emit('Sampe:TaskStart', info);
  }
}
//observer.service.ts
@Injectable()
export class ObserverService {
  constructor(private eventService: EventsService) {
     this.eventsService.on('Sampe:TaskStart', this.handleTaskStart)
  }

  async handleTaskStart(info): Promise<void> {
     console.info('the task is start', info);
  }
}

```

## Tests

```
\\e2e tests
\\ add redis mock
jest.mock('ioredis', () => {
  return {
    default: require('ioredis-mock/jest'),
  };
});

  beforeEach(async () => {
    const moduleFixture: TestingModule = await Test.createTestingModule({
      imports: [EventsModule],
    }).compile();

    app = moduleFixture.createNestMicroservice(moduleFixture);
    await app.init();
    const sub = app.get('REDIS_SUB_CLIENT');
    const subSyncClient = sub.createConnectedClient();
    const pub = app.get('REDIS_PUB_CLIENT');
   //change the pub publish method with mocked client publish
    jest.spyOn(pub, 'publish').mockImplementation((env, message) => {
      return subSyncClient.publish(env, message);
    });
  });

 it("the event e2e tests", (done) => {
   // mock the callback func to  validate the event data
    const mockedFuc = jest.fn().mockImplementation((value) => {
      expect(value).toEqual({ ut: 'test' });
      //use done to make sure function be called
      done();
      return;
    });
})

```


joncodo:
## Questions:

1. Is this used for scheduling collection tasks?
2. Does this mean we no longer need to use RabbitMQ or Bull?
3. Does each api call to external services add a new Task to redis?
4. How do we handle 10000x tasks at once with Redis?
5. How do you know which tasks get processed first?

joncodo:
> scalable and with load balance

Can you please elaborate on how you plan to make this part work?

apps/core/test/events.e2e-spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  it('Event', (done) => {
+    const service = app.get(EventsService);
+    const mockedFuc = jest.fn().mockImplementation((value) => {
+      expect(value).toEqual({ ut: 'test' });
+      //use done to make sure function be called
+      done();
+      return;
+    });
+    service.on('Custom', mockedFuc);
+    service.emit('Custom', { ut: 'test' });
```
joncodo, line 48: Thank you for adding these tests. This makes the code a lot more clear to me. Great job 👍

joncodo:
Is this PR still good?

jaxzhou:
Questions:
Is this used for scheduling collection tasks?  
YES, But not only for scheudling collection task.

Does this mean we no longer need to use RabbitMQ or Bull?
no need to RabbitMQ, bull is for task queue still required.

Does each api call to external services add a new Task to redis?
Yes, add new task to bull, send event message use eventsService. similar but not the same

How do we handle 10000x tasks at once with Redis?
Push 10000x task with name and datas into bull queue. the bull queue service would run task with concurrency. (Not run at once but push at once)

How do you know which tasks get processed first?
the bull queue service consumer the task from queue FIFO.

klesh:
LGTM, I'm merging this PR, since it's needded for  @jaxzhou  to continue his work.

PR #173, opened 2021-08-10
chore: change database connection log message to include any db

oddscenes:
We have a log that was outputting `ERROR: connecting to mongodb` even when there was postgres connection issues

- Change log message to include any database connection issues

---

**Previous log example**
```sh
ERROR: connecting to mongodb DatabaseError [SequelizeDatabaseError]: relation "..." does not exist
```

**New log example**
```sh
ERROR: connecting to database DatabaseError [SequelizeDatabaseError]: relation "..." does not exist
```

### DIFF
--- a/src/plugins/index.js
+++ b/src/plugins/index.js
@@ -75,7 +75,7 @@ async function initialize () {
         .map(plugin => plugin.initialize(connection.rawDb, enrichedDb, plugins))
     )
   } catch (error) {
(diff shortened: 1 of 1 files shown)


UltimateBeaver:
@oddscenes It's not necessarily a db error, so it's better to just say `ERROR: Failed to initialize plugins`. I'm gonna make the change in a separate PR, thanks for the PR though.

PR #181, opened 2021-08-11
docs: adding issue templates and pull request templates

joncodo:
## Why

We need a way to organize our issues with enough context for us to communicate effectively. This is just a sample but we can update these files to anything we want. Looking for feedback on this.

These are similar to the ones provided by github by default.

The motivation is to remind developers to provide many details so others can help them develop the code :)


oddscenes:
@joncodo @UltimateBeaver this one is updated now - PR template and 4 issue templates

joncodo:
Since we have no issue templates now and these changes are only in the .github directory, I am going to merge this.

PR #185, opened 2021-08-11
docs: minor fixes to readme to help developers set up

kevin-kline:
This fixes README commands to start project. Also added grafana_icon.svg to correct directory.

package.lock.json is also committed as a first time commit - hence the large number of lines in this PR.


klesh:
Hi, @kevin-kline , I saw some changes from `ts-main-doc` which had been merged, could you rebase `ts-main` ?

joncodo:
@kevin-kline any idea why git always commits a new package.json for you?

plugins/jira/src/index.ts
```diff
@@ -37,6 +38,9 @@ class Jira implements Plugin {
     console.info('Excute Jira', options);
     return;
   }
+  async collectData(primaryKeys: PrimaryValues, consumerModule) {
```
klesh, line 41: I think this method and PrimaryValues type should not be included in this PR.

klesh (commented):
Hi, @kevin-kline , looking good overall, except some minor issue, I think we better keep this PR in `doc` level. 😊

kevin-kline:
> @kevin-kline any idea why git always commits a new package.json for you?

if you mean package.lock.json I think the reason this time was that none existed at the time of committing, which I mentioned in the description. If you mean package.json, I'm not sure I understand.

plugins/jira/src/index.ts
kevin-kline: I reverted the commit as per your feedback.

klesh:
LGTM

.env.sample
```diff
@@ -1,4 +1,4 @@
 # Main Database Connetion String
-DB_URL=mysql://root:root@localhost:3306/lake_db
+DB_URL=postgresql://postgres:postgres@localhost:5432/lake
```
narrowizard, line 2: I think we will use mysql connection string in `.env.sample` instead of postgres. But the PR is merged. I'll fix it in another PR.

narrowizard: BTW, services declared in `devops` and documents should be updated as well. #237

klesh: I checked the `devops/docker-compose.yml` and it was still using postgresdb. There were some inconsistency, pls update docker-compose.yml as well, make sure whole process go through when you do that.👍

PR #187, opened 2021-08-11
feat(typescript): new plugins template

jaxzhou:
https://github.com/merico-dev/lake/issues/184

## Why
Add Plugin should be easier. and we should take the advantage  of nestjs/injectable
Plugin would focus on the task it should completed. the migrates steps should move to some foundation components.
Plugin entry , collector and enricher should be injectable. so the queue service or the plugin entry could get a instance of collector/enricher to execute

## What
move plugin register into plugin/core. The Plugin module loaded by the consumer module 
Add decorator of collector and enricher. so the pluginmodule register collectors and enrichers 
the plugin entry load collector and enricher by inject ref. 
Add entities sample of jira issue collector. entity would be loaded/migrated by typeorm module supported

### How Plugin Start

The plugins is a sequence of tasks. then entry of a plugin would organization the tasks in sequence. 
All tasks should run in queue service. So we need to register plugin entry with a name in queue service

```
// <root>/apps/queue/src/consumer/index.ts
export class ConsumerModule {
  static forRoot(queue = 'default'): DynamicModule {
    return {
      module: ConsumerModule,
      imports: [
        BullQueueModule.forRoot(queue),
        PluginModule.Register([{ name: 'Jira', schedule: Jira }]),
        ConsumerService,
      ],
      providers: [],
      exports: [ConsumerService],
    };
  }
}
```

```
// Call Producer Service to add job
import ProducerService from 'apps/queue/src/producer/service';

@Injectable()
export class SomeService {
    constructor(private produer: ProducerService) {}

    methodToAddJob() {
        this.producer.addJob('Jira', {projects: ['MyJiraProject']})
    }
}
```

### How To Add a Plugin

Add the plugin entry extends the Scheduler and implement the execute method.

- create your collector an enricher
```
\\ plugins/myplugin/src/collector/sample.ts

@Injectable({
  scope: Scope.TRANSIENT, //set to TRANSIENT, so collector would not be single instance
})
class SampleCollector implements IExecutable<void> {
  constructor() {}
  async execute(): Promise<void> {
    return;
  }
}

export default SampleCollector;

```

declare the collectors in plugin entry

```
\\ plugins/myplugin/src/index.ts

@Injectable({
  scope: Scope.TRANSIENT,
})
@Collector({
  'Sample': SampleCollector,
})
class MyPlugin extends Scheduler<void> {
   async execute(options: any, contextId: ContextId): Promise<void> {
    const collector = await this.collectorRef.resolve(
      'Sample',
      'Jira',//name Registed in PluginModule
      contextId,
    );
    await collector.execute(options);
    return;
  }
}
```

## Not Include

Load the entities and migrates in plugins. that's another part about typeorm models.


likyh:
Schedule should not be a foundation feature.  
Schedule should decide by plugin self.
I support PluginModule should save in plugins/core But not support interface plugin replace to PluginModule. It should be independent for other plugin extends

jaxzhou:
Plugin should organize the sequence of tasks but not to manage the tasks. A scheduler is an abstract queue task manager to receive the organized sequence. 
Plugin would have two part. 1.the plugin entry would receive the options and organized the tasks. 2.the plugin tasks process data.(the task would have two type: collector, enricher)

jaxzhou:
new pull request at #244

PR #190, opened 2021-08-11
docs: add data source definition details to plugin references

oddscenes:
## WHY
The plugins section was far down in the main readme and hard to navigate to. as well the plugins section itself needs better definition, and a more descriptive name.

## WHAT 
- Change references to "Plugins" to "Data Source Plugins"
  - When we add more plugins that do other things (not data source plugins), we will need to use "Plugins" as a main category, and `data source`/`enhancement`/`utility` etc as sub-categories for plugins. But for now, just saying Data Source is easier.
- Moved "Data Source Plugins" under "Requirements" in main readme. Up above user setup near the top.
- Added a short paragraph explanation under **Data Source Plugins** title

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Section | Description | Link
 :------------ | :------------- | :-------------
 Requirements | Underlying software used | [View Section](#requirements)
 User Setup | Quick and easy setup | [View Section](#user-setup)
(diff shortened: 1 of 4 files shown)


UltimateBeaver:
LGTM

PR #194, opened 2021-08-12
Feat: add request param validator

narrowizard:
## why
We should validate incomming requests before handling it. The validator will reject bad requests and protect our service.

## what
- This PR added request parameter validator for restful api.  
- `SourceController` was added as an example for request validator.
- Also, some end-to-end test suit is added.

/review @likyh @jaxzhou
/cc @UltimateBeaver


narrowizard:
This PR is out of date, I'll commit another soon.

PR #195, opened 2021-08-12
docs: collaboration notes

joncodo:
# Why

We need to document these things in the repo and not only in internal communication. This improves our docs substantially :)


UltimateBeaver:
LGTM

PR #203, opened 2021-08-12
docs: small formatting updates to plugin doc

oddscenes:
### WHY
There was very small formatting issues with plugin docs. All the details are looking good. This PR will just fix a few of those minor things

## WHAT
- bullet points instead of inline a) b) etc (more traditional markdown style)
- formatted numbering flow


UltimateBeaver:
LGTM

PR #204, opened 2021-08-12
Sample config file doc updates

oddscenes:
## WHY
Some minor spelling/grammar things in example config files. As well I think it's nice to highlight what parts of these example config files actually **NEED** replacing (vs config settings that can be left alone).

## WHAT
- Added unicode symbol `➤➤➤` to highlight "Replace" sections
- Fixed some spelling things


UltimateBeaver:
LGTM

PR #211, opened 2021-08-13
docs: gitlab docs for config and summary

joncodo:
## WHY

We need more information at the gitlab level. This was too confusing for Maxim as it was. 

## WHAT

- New section for summary with images
- Configuration
- Gathering Data with Github

### DIFF
--- a/src/plugins/gitlab-pond/README.md
+++ b/src/plugins/gitlab-pond/README.md
@@ -13,7 +13,45 @@ Commit Count | Number of Commits
 Added Lines | Accumulated Number of New Lines
 Deleted Lines | Accumulated Number of Removed Lines
 
+## Configuration
+
+First, you have to configure `lake/config/plugins.js`.
+
+1. `cp config/plugins.sample.js config/plugins.js`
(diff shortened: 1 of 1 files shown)


src/plugins/gitlab-pond/README.md
```diff
@@ -13,6 +25,42 @@ Commit Count | Number of Commits
...
+ {
+     "gitlab": {
+         "projectId": 8967944,
+         "branch": "<your-branch-name>", (Optional, default branch is used)
+     }
+ }
+```
+2. You can configure lake to get all of your data automatically.
+
+- Make sure you have a file called `config/local.js`
```
UltimateBeaver, line 50: If we're guiding users through the "User Setup" flow, then we should use `docker.sample.js` here instead of `local.sample.js`?

UltimateBeaver (commented):
Looking good in general, please see the only comment.

src/plugins/gitlab-pond/README.md
joncodo: yes

hezyin:
LGTM

PR #213, opened 2021-08-13
docs: general doc feedback from maxim v1

joncodo:
## WHY

These are feedback items from @MaximDub that we would like to add to our readme. The hope is that this makes it easier for new users to run the application.

## WHAT

The largest change is to move the developer setup to another file to not overwhelm new users.

You can see this ticket for more details: https://github.com/merico-dev/lake/issues/200

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,17 +8,17 @@
 
 ### What is Dev Lake?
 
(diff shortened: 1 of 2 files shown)


UltimateBeaver:
LGTM

PR #218, opened 2021-08-13
Leadtime

UltimateBeaver:
1. New configuration schema to describe type and status conversion between raw and standard.
    Each type of issue may have different set of statuses representing `Resolved` or `Rejected`, so
    in new schema, status-mapping are nested within type-mapping. User can choose to map one or
    more JIRA types to a single standard type, same logic for status conversion.
 2. Calculates Lead-Time for those `Resolved` issues only.

Some notes for future work:

1. We need more documentation for configuration as some mappings are not self-explanatory (jira-pond/README.md needs to be updated)
2. We need to verify that grafana dashboards are still working.

### DIFF
--- a/config/docker.sample.js
+++ b/config/docker.sample.js
@@ -2,10 +2,7 @@ module.exports = {
   // Configuration of lake's own services
   lake: {
     // Enable basic authentication to the lake API
-    // token: 'mytoken',
-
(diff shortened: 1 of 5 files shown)


UltimateBeaver:
LGTM

PR #222, opened 2021-08-16
chore(ci): update github actions

narrowizard:
## why
Previous CI configuration combined unit test and e2e test.
- it's not a great practice since these two kind tests are not related. 
- Another reason is that we need setup third-party services for e2e test but not for unit test.

## what
- separate unit test and e2e test into different github actions
- setup mysql for e2e test
- update github actions scope/name

### DIFF
--- a/.github/workflows/commit-msg.yml
+++ b/.github/workflows/commit-msg.yml
@@ -1,8 +1,8 @@
-name: Lint Commit Messages
+name: lint
 on: [pull_request]
 
 jobs:
-  commitlint:
+  commit-msg:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.env.sample
```diff
@@ -1,4 +1,6 @@
 # Main Database Connetion String
 DB_URL=postgresql://postgres:postgres@localhost:5432/lake_db
+# Test Database Connection String
+DB_TEST_URL=postgresql://postgres:postgres@localhost:5432/lake_test_db
```
jaxzhou, line 4: do not use xx_TEST in config file. 
if we want to support different env config file. do in ConfigModule. load .env.${NODE_ENV} while apps bootstrap.
don't forget to add .env.${NODE_ENV} file to .gitignore

narrowizard: fixed

jaxzhou:
LGTM

PR #223, opened 2021-08-16
feat: add typeorm module for rest api

narrowizard:
- [x] Please merge this PR after #222 , as database setup is required after typeorm module including.  

## Why
We need to persist source info and task info into database.

## What
This PR added typeorm support for ts-main branch, with synchronizing models feature.

- [x] init  typeorm module
- [x] add some models
- [x] auto sync models when start service

## Notice
This PR has not handled database migration, we will add db migration in another PR.

> using synchronize feature is not recommended in production environment.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,4 +1,4 @@
 # Main Database Connetion String
-DB_URL=postgresql://postgres:postgres@localhost:5432/lake_db
+DB_URL=mysql://root:root@localhost:3306/lake_db
 # Redis Connection String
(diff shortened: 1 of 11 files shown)


likyh (commented):
LGTM

narrowizard:
The CI failed because coverage down. But I think it will be a too large PR if add more logic and test files. What's your suggestion? @jaxzhou

jaxzhou:
LGTM

PR #225, opened 2021-08-16
fix: remove mapping validation

klesh:
mapping was no longer needed, validation should be removed, or error would be throwed if setting missed in configuration file.


hezyin:
LGTM

PR #229, opened 2021-08-16
Cli docker

joncodo:
## WHY

When we use the CLI, it is usually for new users of the app that are following the readme instructions. Mainly `npm run configure` and `npm start`. However, the CLI did not write to the config/docker.js file. This PR fixes that.

## WHAT / HOW

It uses the same things that the CLI was using to do this except it now has methods to ask for data and separate methods to write to the file

## Testing

- I ran `npm run configure` and saw all prompts. 
- Verified all of config/plugins,local and docker were set properly
- Ran `npm start` and saw it pull down valid data

### DIFF
--- a/config/docker.sample.js
+++ b/config/docker.sample.js
@@ -7,31 +7,31 @@ module.exports = {
   // Configuration of MongoDB
   mongo: {
     connectionString:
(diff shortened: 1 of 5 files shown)


hezyin:
LGTM

PR #234, opened 2021-08-16
fix: update local sample dev setup and adjust dev setup docs

oddscenes:
## WHY
`local.sample.js` needed to be updated to work with the dev setup. As well the dev setup docs needed an update to use the new command line config setup.

## WHAT
- Uncomment `cron` section in local sample file (just like docker sample)
- Update `DEVELOPER_SETUP.md` to have proper flow for command line tool 
- Update `DEVELOPER_SETUP.md` to have better info about how we immediately start the data collection

### DIFF
--- a/config/local.sample.js
+++ b/config/local.sample.js
@@ -23,17 +23,14 @@ module.exports = {
     dialect: 'postgres'
   },
   cron: {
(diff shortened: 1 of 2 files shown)


hezyin:
LGTM

PR #235, opened 2021-08-17
feat: add query parameter validation

narrowizard:
## why
We should validate incomming requests before handling it. The validator will reject bad requests and protect our service.

## what
- This PR added request parameter validator for restful api.  
- `SourceController` was added as an example for request validator.
- Also, some end-to-end test suit is added.


likyh (commented):
reviewed, but I don't know if it good 😢

jaxzhou:
LGTM

PR #238, opened 2021-08-18
fix: remove some gitub config on the main branch

joncodo:
# Summary

I pushed up a .github copy of the ts-main branch to main. This was a mistake and some files will not work on the JS main branch that are on the ts-main branch. However, I do want the issue and PR templates. 

### Key Points

- [x] Removes broken CI steps

### Current Behavior
Currently, the main build seems to pass but this is better


hezyin:
LGTM

Issue #151, opened 2021-08-09
[ts] Plugin configuration management


jaxzhou:
## Description
With our plugin system, there are configurations to each plugin, we should provide a easy-use way to set configuration for plugins. The configuration system should meet following requirements:  
- **Dynamically** update the configuration.
- Support multiple sets of configurations.

## Describe the solution you'd like
We will create the configurations through restful api. for example
```
POST  /sources
{
    type: 'Jira', // plugin name
    options: {
        host: '',
        username: '',
        password: 'password or api token'
    } // plugin configurations
}

response:
{ source: 'source id' }
```

## Context
The configuration system is based on our plugin system design, related to #244

narrowizard:
We have some more work to do before adding the source api. 
## My plans
- [ ] adding class validator for request validation. #194 
- [ ] adding database layer and schema definition. 
- [ ] add source api (CRUD)
- [ ] add unit test and e2e test for source api
- [ ] add error handler for rest api

joncodo:
So for this one, I have a few questions:

1. Does the data get fetched automatically on a loop for all sources in the DB?
2. Does this support many gitlab projects and many jira projects at once?
3. When the job to fetch data is created, do we create a parent job and many child jobs to be processed by bull? How does that part work?

Thanks in advance for the info 👍

narrowizard:
> Does the data get fetched automatically on a loop for all sources in the DB?

There's no plan to implement the synchronize logic in lake, data fetch task for each source will be added by task api(#152 ). Maybe we can create a cron job to trigger the task api schedually.
> Does this support many gitlab projects and many jira projects at once?

Yes

joncodo:
@hezyin would like us to change the name 'source' to 'plugin'

narrowizard:
Close this because #242 is merged.  
> would like us to change the name 'source' to 'plugin'

Open another issue as a proposal for this, thank u : )

PR #242, opened 2021-08-19
feat: source api implement

narrowizard:
# Summary

### Key Points

### Description
#151 

#151 

### Other Information
Now we can add plugin configuration via source api.


klesh (commented):
LGTM

klesh (approved):

apps/rest/src/controllers/source.spec.ts
```diff
@@ -34,34 +63,63 @@ describe('SourceController', () => {
       const reqListSource = {
         page: 1,
         pagesize: 10,
-        type: 'jira' as const,
+        type: 'jira',
       };
+
+      const fn = jest
+        .spyOn(sourceService, 'list')
+        .mockImplementation(async () => {
```
joncodo, line 71: I think this test needs to call 'expect' in order to be a real test

apps/rest/src/controllers/source.spec.ts
```diff
@@ -34,34 +63,63 @@ describe('SourceController', () => {
...
 
   describe('get', () => {
     it('should return target source', async () => {
+      const fn = jest
+        .spyOn(sourceService, 'get')
+        .mockImplementation(async () => {
+          return new Source();
+        });
+
       await sourceController.get('id');
```
joncodo, line 91: I think this test needs to call 'expect' in order to be a real test

apps/rest/src/controllers/source.spec.ts
```diff
@@ -34,34 +63,63 @@ describe('SourceController', () => {
...
           auth: 'base64EncodedAuthToken',
         },
       };
+      const fn = jest
+        .spyOn(sourceService, 'update')
+        .mockImplementation(async () => {
+          return new Source();
+        });
+
       await sourceController.update('id', reqUpdateSource);
```
joncodo, line 111: I think this test needs to call 'expect' in order to be a real test

apps/rest/src/controllers/source.spec.ts
```diff
@@ -34,34 +63,63 @@ describe('SourceController', () => {
...
 
   describe('delete', () => {
     it('should delete source', async () => {
+      const fn = jest
+        .spyOn(sourceService, 'delete')
+        .mockImplementation(async () => {
+          return new Source();
+        });
+
       await sourceController.delete('id');
```
joncodo, line 123: I think this test needs to call 'expect' in order to be a real test

apps/rest/src/services/source.ts
```diff
@@ -0,0 +1,61 @@
...
+    const sources = await this.em.getRepository(Source).find(options);
+    return {
+      offset,
+      total,
+      data: sources,
+    };
+  }
+
+  async create(data: CreateSource): Promise<Source> {
+    const source = new Source();
```
joncodo, line 35: Can we just do `const source = new Source(data);`? That way, we don't have to manage the individual attributes here?

apps/rest/src/types/pagination.ts
```diff
@@ -12,7 +12,7 @@ export class PaginationRequest {
 }
 
 export class PaginationResponse<T> {
-  page: number;
```
joncodo, line 15: Don't we need page-size, offset and total to do proper paging?

apps/rest/test/source.e2e-spec.ts
```diff
@@ -19,23 +24,33 @@ describe('SourceController (e2e)', () => {
...
       };
       return request(app.getHttpServer())
         .post('/source')
-        .send(sourceWithoutType)
-        .expect(400)
+        .send(newSource)
+        .expect(201)
         .expect((res) => {
-          expect(res.body).toMatchSnapshot();
+          expect(ignoreEntityProps(res.body)).toMatchSnapshot();
```
joncodo, line 47: How does this method know which snapshot you are referencing?

joncodo:
It seems like sourceTask routes are not being loaded by Nest

Screen Shot 2021-08-20 at 9 48 04 AM

I created a source fine:

Screen Shot 2021-08-20 at 9 48 26 AM

But source task failed:

Screen Shot 2021-08-20 at 9 48 42 AM

I also see no tests for the sourceTask file

apps/rest/src/controllers/source.spec.ts
narrowizard: You are right. In my previous plan, I will add some real test suit when `SourceController.list` contains its own logic.  
> `SourceController.list` only call `SourceService.list` method without any other codes now, maybe it make little sense to `expect` it.

apps/rest/test/source.e2e-spec.ts
narrowizard: I think jest handled it well.
> jest combined `describe` content, `it` content and call order of `toMatchSnapShot` as a primary key, to match specified snapshot. (I guess)

apps/rest/src/services/source.ts
narrowizard: I think both solution is acceptable. No much ideas.

PR #253, opened 2021-08-19
Ts migrations

likyh:
# Summary

### Key Points

- [*] New or existing documentation is updated

### Description
We need migrations and entity load for plugin to finish db init.

https://github.com/merico-dev/lake/issues/247

### New Behavior
Now Core will read `plugins/pluginName/src/migrations/*.ts` or `plugins/pluginName/src/**/*.migration.ts` ro run migration when queue start.
Rest will read `rest/migrations/*.ts` ro run migration when rest start.

### DIFF
--- a/apps/rest/src/app.module.ts
+++ b/apps/rest/src/app.module.ts
@@ -10,16 +10,18 @@ import { SourceController } from './controllers/source';
 import { SourceTaskController } from './controllers/sourceTask';
 import { SourceService } from './services/source';
 import { SourceTaskService } from './services/sourceTask';
+import { migrations } from './migrations';
+
 @Module({
   imports: [
     ConfigModule.forRoot({
       isGlobal: true,
       envFilePath: process.env.NODE_ENV === 'test' ? '.env.test' : '.env',
     }),
-    CustomTypeOrmModule.forRootAsync(null, {
+    CustomTypeOrmModule.forRootAsync({
       entities,
-      // FIXME: using db migration instead of synchronize
-      synchronize: true,
+      migrations,
+      migrationsRun: true,
     }),
   ],
   controllers: [AppController, SourceController, SourceTaskController],
--- a/apps/rest/src/migrations/index.ts
+++ b/apps/rest/src/migrations/index.ts
@@ -1,0 +1,12 @@
+export const migrations = [];
+
+if (process.env.NODE_ENV !== 'test') {
+  {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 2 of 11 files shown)


likyh:
It's work base on https://github.com/merico-dev/lake/pull/243

apps/rest/src/migrations/index.ts
```diff
@@ -0,0 +1,12 @@
+export const migrations = [];
+
+if (process.env.NODE_ENV !== 'test') {
```
narrowizard, line 3: If we don't run migration on test env, can we run e2e test properly?  
Just a question, keep free. @likyh

likyh: Because we won't compile in the test, we can use migrations to add directories directly in the test.

narrowizard: we should run migration before e2e test, is that right?  
If so, we may need extra job on github actions `e2e-test`.  
anyway, it doesn't matter now. we can add it when we need.

likyh: I don't know if we need run migrations 😹

narrowizard (approved):
LGTM

narrowizard:
Is this PR still good? Seems it need a reabase locally.

jaxzhou:
i think we should only add migration files in the plugin.  plugins/migrations/*.ts should not be supported. [proposal]

likyh:
> i think we should only add migration files in the plugin. plugins/migrations/*.ts should not be supported. [proposal]

off course.   `plugins/pluginName/migrations/*.ts` or `plugins/pluginName/**/*.migration.ts`

apps/rest/src/app.module.ts
```diff
@@ -10,16 +10,18 @@ import { SourceController } from './controllers/source';
...
     ConfigModule.forRoot({
       isGlobal: true,
       envFilePath: process.env.NODE_ENV === 'test' ? '.env.test' : '.env',
     }),
-    CustomTypeOrmModule.forRootAsync(null, {
+    CustomTypeOrmModule.forRootAsync({
       entities,
-      // FIXME: using db migration instead of synchronize
-      synchronize: true,
+      migrations,
```
narrowizard, line 23: According to the latest pipeline (failed because table is not existed), I think we should 
- add migrations for rest api models in this PR, and also 
- add migration cli at e2e test pipeline. 
- maybe do these jobs for plugin as well.

narrowizard (changes requested):

Issue #192, opened 2021-08-12
[feat][ts] we need a logger in ts-main


likyh:
https://docs.nestjs.com/techniques/logger

joncodo:
Feel free to re-open with more details on why we need this and what you are proposing.

narrowizard:
nestjs providers a built-in log module, I've added some example at #265(for controllers, services and providers). Seems the buit-in log is enough for us now, do you have any other ideas? @likyh

likyh:
> nestjs providers a built-in log module, I've added some example at #265(for controllers, services and providers). Seems the buit-in log is enough for us now, do you have any other ideas? @likyh

No more idea about this. Just want a example for how to log some message

narrowizard:
close due to language switch

PR #265, opened 2021-08-23
feat: enhence rest api tools

narrowizard:
# Summary
This PR did some enhencement for app module(rest api).

### Key Points

### Description
- [x] add global excepiton handler
- [x] add logger example #192 
- [x] log exceptions
- [x] add default value for request pagination parameter.

### DIFF
--- a/apps/rest/src/types/pagination.ts
+++ b/apps/rest/src/types/pagination.ts
@@ -4,11 +4,11 @@ import { Max, Min } from 'class-validator';
 export class PaginationRequest {
   @Min(1)
   @Type(() => Number)
-  page: number;
+  page = 1;
 
   @Max(100)
   @Type(() => Number)
-  pagesize: number;
+  pagesize = 20;
 }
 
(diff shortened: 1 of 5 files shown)


apps/rest/src/types/pagination.ts
```diff
@@ -4,11 +4,11 @@ import { Max, Min } from 'class-validator';
 export class PaginationRequest {
   @Min(1)
   @Type(() => Number)
-  page: number;
+  page = 1;
 
   @Max(100)
   @Type(() => Number)
-  pagesize: number;
+  pagesize = 20;
```
joncodo, line 11: We should set a default page size but allow it to be overwritten I think. This way users get 20 by default but can pass in a value no greater than 100.

narrowizard: Yeah, as pagesize is a property of class `pagination`, I think it meets your requirement.

narrowizard:
close due to language switch

PR #281, opened 2021-08-25
feat: log filename and line

narrowizard:
# Summary
This PR optimized logger utils.

### Key Points

### DIFF
--- a/api/controllers/source/source.go
+++ b/api/controllers/source/source.go
@@ -17,7 +17,7 @@ func Post(ctx *gin.Context) {
 		ctx.AbortWithError(http.StatusBadRequest, err)
 		return
 	}
-	logger.Debug("CreateSource", data)
+	logger.Debug(data)
 	err = services.NewSource(data)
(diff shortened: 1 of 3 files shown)


api/controllers/source/source.go
```diff
@@ -17,7 +17,7 @@ func Post(ctx *gin.Context) {
 		ctx.AbortWithError(http.StatusBadRequest, err)
 		return
 	}
-	logger.Debug("CreateSource", data)
+	logger.Debug(data)
```
joncodo, line 20: With lines like this all over the application, it can be hard to read. I think we need something like this:

`logger.Debug("My context string", data)`

I will add this

joncodo:
Hey @narrowizard ! I have added some improvements to this here: https://github.com/merico-dev/lake/pull/284

joncodo:
LGTM. Please merge both mine and this one @hezyin

klesh (approved):
LGTM

Issue #277, opened 2021-08-24
Add integration tests for api routes and run them in the pipeline


joncodo:
## Description
We need E2E tests so we can test all layers.

We will need:

- [ ] a new .github/workflow like: https://github.com/merico-dev/lake/blob/ts-main/.github/workflows/test-e2e.yml
- [ ] Some E2E tests like: https://github.com/merico-dev/lake/blob/ts-main/apps/rest/test/source.e2e-spec.ts

## Describe the solution you'd like
- [ ] Make https call `POST /source`
- [ ] Connect to DB
- [ ] Save record
- [ ] Verify record exists

klesh:
@narrowizard Pls take a look at this

narrowizard:
We should divide there two type of test.
## unit test
test files under `test` folder.

## e2e test
test file end with `_test.go` postfix and outside `test` folder.

PR #282, opened 2021-08-25
Feat/#277

narrowizard:
# Summary
This PR do following things, base on #281 
- [x] add e2e test pipeline
- [x] divide e2e test and unit test (according to https://github.com/merico-dev/lake/issues/277#issuecomment-905362158)
- [x] add e2e test for `post source api`

### Key Points

### Description

#277

### Current Behavior

### New Behavior

### Screenshots

### Other Information


.github/workflows/test.yml
```diff
@@ -15,4 +15,4 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
     - name: Test
-      run: go test -v ./...
\ No newline at end of file
+      run: go test -v $(go list ./... | grep -v /test/)
```
joncodo, line 18: Maybe we should name our tests differently.

E2E - my_file_e2e.go
Unit - my_file_test.go

This is since Go advises that you write all your tests in place with the main files:

"Implement test functions in the same package as the code you're testing." from : https://golang.org/doc/tutorial/add-a-test

api/controllers/source/source.go
```diff
@@ -14,13 +14,14 @@ func Post(ctx *gin.Context) {
...
-	logger.Debug("CreateSource", data)
-	err = services.NewSource(data)
+	logger.Debug(data)
+	source, err := services.NewSource(data)
 	if err != nil {
-		ctx.AbortWithError(http.StatusInternalServerError, err)
+		_ = ctx.AbortWithError(http.StatusInternalServerError, err)
 		return
 	}
+	ctx.JSON(http.StatusCreated, source)
```
joncodo, line 26: Nice. This returns the JSON or the error back to the HTTP caller? IE: Postman or cUrl?

config/config.go
```diff
@@ -1,15 +1,19 @@
...
 func init() {
 	V = viper.New()
 	V.SetConfigFile(".env")
+	V.AutomaticEnv()
 	err := V.ReadInConfig()
 	if err != nil {
-		panic(err)
+		logger.Error(err)
 	}
 	V.SetDefault("PORT", ":8080")
```
joncodo, line 18: Don't we want to read the port from the config OR use the default?

.github/workflows/test.yml
narrowizard: > Implement test functions in the same package as the code you're testing.
I totally agree with above point when writing unit test, as unit test is target to test a function.  
But an e2e test it target to test the whole feature (including many layer).  
Eg: building the rest api service and try to do some requests, it may not properly to put the e2e tests in controllers package?

api/controllers/source/source.go
narrowizard: `ctx.JSON` write a json to http response. (with content-type: application/json header).

config/config.go
narrowizard: We will use the default port in most case, but providers the ablity to custom port for users who want.

.github/workflows/test.yml
joncodo: "But an e2e test it target to test the whole feature (including many layer)."

Yes. E2E can be in another folder. Like in ts-main

narrowizard: Oh, I got it, here `go test -v $(go list ./... | grep -v /test/)` means to test files exclude `test` folder. We put e2e test into the test folder and here is unit test configuration.

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: setup mysql
+      run: |
+        sudo /etc/init.d/mysql start
+        mysql -e 'CREATE DATABASE ${{ env.DB_DATABASE }};' -u${{ env.DB_USER }} -p${{ env.DB_PASSWORD }}
+    - name: setup redis
+      uses: shogo82148/actions-setup-redis@v1
+      with:
+        redis-version: '6.x'
+    - name: Test
+      run: go test -v ./test/...
```
narrowizard, line 33: And here is e2e test configuration, only test files in `test` folder @joncodo

PR #306, opened 2021-08-27
Jira plugin v2

joncodo:
@klesh I got stuck on this and could not make it work. Maybe you can take a look at this PR and see if you can get it working.

The main problem I have is to map the response from the api to many JiraIssues

Take a look and tell me what you think


klesh (approved):
LGTM

PR #310, opened 2021-08-30
Collect gitlab commits

joncodo:
NOTE: This is the part for "Commit" Collection from this ticket: https://github.com/merico-dev/lake/issues/309

TODO: 
- [x] Gather all fields we used to gather in the DB from the API response
- [x] Upsert records instead of insert
- [x] Add docs and comment code?

### Key Points

### Description

This code gathers the first page of commits from gitlab. (Paging handled here: https://github.com/merico-dev/lake/issues/316)

## Configuration

In your .env file, you will need to set up

    ```
    # Gitlab
    GITLAB_ENDPOINT=https://gitlab.com/api/v4/
    GITLAB_AUTH=<your access token>
    ```

You can get your access token from `https://gitlab.com/-/profile/personal_access_tokens`

## Gathering Data with Gitlab

To collect data, you can make a POST request to `/task`

    ```
    curl --location --request POST 'localhost:8080/task' \
    --header 'Content-Type: application/json' \
    --data-raw '{
        "Plugin": "gitlab",
        "Options": {
            "projectId": <Your gitlab project id>
        }
    }'
    ```

### Screenshots
Post request to trigger collection:

Screen Shot 2021-08-31 at 3 45 24 PM

Logs output:

Screen Shot 2021-08-31 at 3 40 25 PM

DB Records:

Screen Shot 2021-08-31 at 3 45 56 PM

## Important notes

We had to use `Upsert` here so we did not set duplicate records. This involved setting the primary key to:

`GitlabId       string `gorm:"primary_key"``

And doing this: 

```
                 err = lakeModels.Db.Clauses(clause.OnConflict{
			UpdateAll: true,
		}).Create(&gitlabCommit).Error

		if err != nil {
			logger.Error("Could not upsert: ", err)
		}
```


plugins/gitlab/init.go
```diff
@@ -1,4 +1,15 @@
 package main
 
+import (
+	"github.com/merico-dev/lake/logger"
+	lakeModels "github.com/merico-dev/lake/models"
+	"github.com/merico-dev/lake/plugins/gitlab/models"
+)
+
 func (plugin Gitlab) Init() {
```
kevin-kline, line 9: Is this meant to be lower case init() ? Lower case fires automatically to the best of my knowledge.

joncodo:
Also, we modified the logger:

```
newLogger := logger.New(
		log.New(os.Stdout, "\r\n", log.LstdFlags), // io writer
		logger.Config{
			SlowThreshold:             time.Second, // Slow SQL threshold
			LogLevel:                  logger.Info, // Log level
			IgnoreRecordNotFoundError: true,        // Ignore ErrRecordNotFound error for logger
			Colorful:                  true,       // Disable color
		},
	)
```

you can set, 

`LogLevel:                  logger.Silent`

If you want to suppress the DB statements

plugins/gitlab/init.go
hezyin: @kevin-kline This `Init` is required by ourselves:

```
type Plugin interface {
	Description() string
	Init()
	Execute(options map[string]interface{}, progress chan<- float32)
}
```

hezyin (commented):
LGTM

hezyin (approved):

PR #311, opened 2021-08-31
Feat/add pokemon plugin

ptzolov:
# Summary

### Description
Adding the pokemon pond plugin, as well as a small refactor to UnmarshalResponse. 
UnmarshalResponse was using `io.ReadAll()` and `json.Unmarshall()`. json has a way of unmarshalling data without converting to a `[]byte`. Updated `UnmarshalResponse` to use the built in java.Decoder.

### New Behavior
New pokemon pond plugin is available for ingesting pokemon data.

### Other Information
I was thinking it might be nice to introduce some interfaces around the database and http client objects. It will make it easier to change implementations in the future, as well as give us an easy way to stub them out in the unit tests.

### DIFF
--- a/plugins/core/apiclient.go
+++ b/plugins/core/apiclient.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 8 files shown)


joncodo:
Closing for now. Feel free to re-open

PR #325, opened 2021-09-02
Feat/jenkins

narrowizard:
# Summary
Update jenkins plugins.
- read config from env file
- add start_time for jenkins builds
- remove time filter when save jenkins builds

### Key Points


joncodo:
LGTM

PR #330, opened 2021-09-02
feat: allow post to task with many tasks

joncodo:
# Summary

Users need to be able to query many tasks at once with the api

IE:

```
curl --location --request POST 'localhost:8080/task' \
--header 'Content-Type: application/json' \
--data-raw '[
    {
        "Plugin": "gitlab",
        "Options": {
            "projectId": 20103385
        }
    },
    {
        "Plugin": "jira",
        "Options": {
            "boardId": 8
        }
    }
]'
```

NOTE: This means you will have to send down an array of json objects. Otherwise, you get this error:

Screen Shot 2021-09-02 at 10 51 07 AM

### Key Points

Screen Shot 2021-09-02 at 10 55 11 AM

### DIFF
--- a/api/task/task.go
+++ b/api/task/task.go
@@ -5,20 +5,31 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/binding"
+	"github.com/merico-dev/lake/logger"
+	"github.com/merico-dev/lake/models"
 	"github.com/merico-dev/lake/services"
 )
 
 func Post(ctx *gin.Context) {
-	var data services.NewTask
+	var data []services.NewTask
+
 	err := ctx.MustBindWith(&data, binding.JSON)
 	if err != nil {
-		_ = ctx.AbortWithError(http.StatusBadRequest, err)
+		logger.Error("", err)
+		ctx.JSON(http.StatusBadRequest, "You must send down an array of objects")
 		return
 	}
-	task, err := services.CreateTask(data)
-	if err != nil {
-		_ = ctx.AbortWithError(http.StatusInternalServerError, err)
-		return
+
+	tasks := make([]models.Task, len(data))
+
+	for _, value := range data {
+		task, err := services.CreateTask(value)
+		if err != nil {
+			_ = ctx.AbortWithError(http.StatusInternalServerError, err)
+			return
+		}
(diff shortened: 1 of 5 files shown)


api/task/task.go
```diff
@@ -5,20 +5,31 @@ import (
...
+
+	tasks := make([]models.Task, len(data))
+
+	for _, value := range data {
+		task, err := services.CreateTask(value)
+		if err != nil {
+			_ = ctx.AbortWithError(http.StatusInternalServerError, err)
+			return
+		}
+		tasks = append(tasks, *task)
```
narrowizard, line 31: You should not use `append` with a certain length array, here the length of tasks will be double length as expacted.  

> https://play.golang.org/p/Aqk4KbigBZE

api/task/task.go
```diff
@@ -5,20 +5,31 @@ import (
...
 
 func Post(ctx *gin.Context) {
-	var data services.NewTask
+	var data []services.NewTask
+
 	err := ctx.MustBindWith(&data, binding.JSON)
 	if err != nil {
-		_ = ctx.AbortWithError(http.StatusBadRequest, err)
+		logger.Error("", err)
+		ctx.JSON(http.StatusBadRequest, "You must send down an array of objects")
```
narrowizard, line 19: Why use `ctx.JSON` instead of `ctx.AbortWithError` ?

narrowizard (changes requested):

api/task/task.go
likyh: Unavailable For Legal Reasons
Viewing and/or sharing code snippets is not available in your country for legal reasons. This message might also appear if your country is misdetected. If you believe this is an error, please file an issue.

narrowizard: What about this? https://goplay.space/#Aqk4KbigBZE

api/task/task.go
joncodo: I wanted to send back a message to the user

api/task/task.go
joncodo: Seems like I need this: https://play.golang.org/p/4JjXhCgMvJL

joncodo: Screen Shot 2021-09-03 at 9 00 05 AM

PR #333, opened 2021-09-02
Gitlab pagination continued

kevin-kline:
### Key Points

### Description

This is a WORK IN PROGRESS.

A necessary change to GitLab pagination on collection.
The previous method extracted from the Jira Plugin does not work because GitLab has a different implementation using headers (x-next-page), etc.

This new way is much more similar to the way this was done in the JavaScript version of Lake for reference. 

This change does NOT yet make use of the worker_scheduler that leverages the "ants" package for Go.


plugins/gitlab/tasks/common.go
```diff
@@ -2,19 +2,54 @@ package tasks
...
+			10*time.Second,
+			3,
+		)
+	}
+	return gitlabApiClient
+}
+
+type GitlabPaginationHandler func(res *http.Response) error
+
+func (gitlabApiClient *GitlabApiClient) FetchWithPagination(resourceUri string, page string, pageSize string, handler GitlabPaginationHandler) error {
```
kevin-kline, line 36: This is the important part of the PR to be continued.

joncodo:
Going to close this for now to reduce confusion. I will pick this up tomorrow

PR #343, opened 2021-09-07
feat: add lake cli

narrowizard:
# Summary
This PR aimed to implement auto sync data from data source feature, periodically.

### TODOs
- [x] add cli tool chain
- [x] add api request
- [x] add cron job
- [x] add cron job in docker compose 

### Key Points

### Design
- This PR added a command toolchain called `lake-cli`, which is only for making a request to lake server.
- We aslo added cron job to `lake-cli`, it can request api every some interval.
- We may add more feature to `lake-cli` in the future, eg: create a plugin template. 
- We added a docker-compose.yml in `devops/sync` with a service that run `lake-cli` to trigger cron job.

#### why using single `.env.sync` and `docker-compose.yml`
This feature is implemented based on lake api, but not belongs to lake itself. Aka, it's an extra and optional feature. 
- If we put it in the same `docker-compose.yml` file with lake, it's confused to people who don't want it.
- if we combine `.env.sync` and `.env`, the `lake-cli` application can read content in `.env` file, it's not safe if an app can read information which is no use for itself.


cmd/lake-cli/README.md
```diff
@@ -0,0 +1,29 @@
...
+```shell
+# install from source
+git clone https://github.com/merico-dev/lake.git
+cd lake
+go install ./cmd/lake-cli/
+
+# go version lower than 1.17
+go get -u github.com/merico-dev/lake/cmd/lake-cli
+
+# go 1.17 or higner version
```
klesh, line 16: typo: higher

cmd/lake-cli/README.md
```diff
@@ -0,0 +1,29 @@
...
+go install github.com/merico-dev/lake/cmd/lake-cli
+```
+
+### Usage
+
+```shell
+# trigger lake api
+# cron schedule defined at https://pkg.go.dev/github.com/robfig/cron#hdr-Predefined_schedules
+$ lake-cli api task -m POST --body "[{\"plugin\":\"jira\", \"options\": {\"boardId\": 8}}]" --cron "@hourly"
```
klesh, line 25: maybe we should use config.json of some sort? it's tedious to write json on commandline, especially when it gets longer and longer

narrowizard: What about adding an option to read request body from file to `lake-cli api` command?

klesh (approved):
LGTM

PR #348, opened 2021-09-07
feat: scaffold for config ui with next

oddscenes:
# Summary

### Key Points

### Description
Added the foundation for a config UI, in a `/frontend` directory. Uses nextjs (express/react setup). Integrated with docker-compose. Will need some additional env/connections going forward, but want to get it into the `go-main` branch before it becomes too stale.

>  ⚠️ Updated: Now also has read/write setup for `.env`

Logic is:
- It will read existing key/value pairs and overwrite if exist
- It will create a key/value pair if it doesn’t exist in env file
- Missing: create a new env file if one doesn't exist (will do another PR)

No validation/styling/additional logic or anything. I think the PR is already starting to get too big. But as of now you can go to browser, see env config, and overwrite existing values. 

Screen Shot 2021-09-08 at 1 59 53 PM

no, early stages of [https://github.com/merico-dev/lake/issues/337](https://github.com/merico-dev/lake/issues/337)

### Current Behavior
No frontend config ui to edit any settings/files

### New Behavior
New ui setup that we will use to edit config. For now it can edit the main env file, in a basic way from the browser

### Other Information
After running `make compose` the frontend UI will be available at `localhost:4000/frontend`


joncodo:
@oddscenes I am going to challenge you to make the backend with golang and not node.js. I think this makes a deviation from a decision we made to have all of our backend in golang. 

@hezyin what do you think about this?

oddscenes:
@joncodo I don't think thats a good idea. Reason most frontend scaffolds start with next/create react app etc is because theres so much stuff out of the box to build on. We would need to set up most of this stuff ourselves from scratch...

- Server side vs client side rendering
- Hot reloading
- Babel/webpack transpiring
- Dev/prod setup
- Compiling styles
- etc

I guess it could be a basic react app with minimal stuff, missing most of this. In which case we just compile it manually or something after saving. But regardless, doesn't make sense to me to avoid all this just to adhere to same language. Unless there is a strong case for server performance or something that im missing here...

oddscenes:
After discussion we're also going to add the capability to read and write to ENV files in this PR. So I'm moving it to draft status. As of now it's able to read env (just updated). Still needs overwrite capability.

joncodo:
Sounds good to me

docker-compose.yml
```diff
@@ -29,23 +27,28 @@ services:
     env_file:
       - ./grafana/grafana.config
     restart: always
-    networks:
-      - lake
     depends_on:
       - mysql
+  frontend:
```
klesh, line 32: this docker-compose file is user-oriented,  which means user can download our docker-compose.yml file to launch our app without cloning our whole repo (there are still some grafana files needed, but that is the goal). So we should not put building process here.

klesh: and maybe we should name it more specific, like Configurator or something, frontend are too generic and I don't know, it's not a cool name... just personal feeling

oddscenes:
@klesh updated `frontend` name to be `config-ui` - agree it can be more descriptive. 

Regarding the moving to docker hub, it will require quite a bit of changes/a few decisions i think. The cross communication between containers, read/write of env files logic will be different. I agree it will be a better setup, to just have docker-compose setup containers, but all the setup for env file read/write needs to be figured out again.

joncodo (approved):

PR #350, opened 2021-09-07
feat: remove fk from gitlab plugin

joncodo:
# Summary
 
Remove FK since we are only collecting data and we do not want to store based on data integrity.

From @hezyin 👍 

```
I think there's a misconception, foreign key is not a performance tool. It's meant to ensure the data integrity. It actually degrades performance slightly during insert / update / delete due to the time spent on checking the foreign key constraint. What helps with performance is index. They're two different things. 

I guess what confuses you is that people often create indexes on those foreign key columns, which mixes these two things up a little bit.
```

Removed all the gorm foreign keys but left the primary keys.

### How I tested this

- dropped all DB tables
- removed fk on models
- restarted with `make dev`
- Sent post request to gitlab plugin
- All data was still stored and no FK existed
- Checked for records in all tables

These tables were affected:

Screen Shot 2021-09-07 at 1 12 17 PM

This is the log output showing it working

Screen Shot 2021-09-07 at 1 12 09 PM

### DIFF
--- a/plugins/gitlab/models/gitlab_commit.go
+++ b/plugins/gitlab/models/gitlab_commit.go
@@ -7,7 +7,6 @@ import (
 type GitlabCommit struct {
 	GitlabId       string `gorm:"primary_key"`
 	ProjectId      int
-	Project        GitlabProject `gorm:"foreignKey:ProjectId"`
 	Title          string
 	Message        string
 	ShortId        string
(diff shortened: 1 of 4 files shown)


plugins/gitlab/models/gitlab_commit.go
```diff
@@ -7,7 +7,7 @@ import (
 type GitlabCommit struct {
 	GitlabId       string `gorm:"primary_key"`
 	ProjectId      int
-	Project        GitlabProject `gorm:"foreignKey:ProjectId"`
+	Project        int
```
klesh, line 10: I think it should be removed instead of changing type, Since there were ProjectId for that already.

klesh (changes requested):

plugins/gitlab/models/gitlab_commit.go
joncodo: Fixed

PR #355, opened 2021-09-08
fix: jira board complains `created_at` being 0

klesh:
# Summary

jira board updation would fail due to `created_at` being `0000-00`

### Key Points

### Description

This fix makes jira board information update works


narrowizard (approved):
LGTM

PR #357, opened 2021-09-08
feat: add step concurrency for gitlab apiclient

klesh:
# Summary

To handle some pagination api doesn't return `x-Token` header

### Key Points

### Description
Some gitlab api like commits doesn't provide `x-total` header, causing plugin collect only the first page for gitlab_commits .
This PR handles this situation in step concurrency manner.
1. For each STEP, send out 1...10 request for each page
2. If page 10 has 'x-next-page' header, go for next STEP
3. If page 1..10 has no 'x-next-page' header, break the loop

### DIFF
--- a/plugins/gitlab/tasks/gitlab_api_client.go
+++ b/plugins/gitlab/tasks/gitlab_api_client.go
@@ -2,6 +2,7 @@ package tasks
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


plugins/gitlab/tasks/gitlab_api_client.go
```diff
@@ -80,31 +111,67 @@ func (gitlabApiClient *GitlabApiClient) FetchWithPaginationAnts(resourceUri stri
...
-		err1 := scheduler.Submit(func() error {
-			url := fmt.Sprintf(resourceUriFormat, pageSizeInt, currentPage)
-
-			res, err := gitlabApiClient.Get(url, nil, nil)
-
-			if err != nil {
-				return err
+	// not all api return x-total header, use step concurrency
+	if total == -1 {
+		conc := 10
```
likyh, line 116: conc maybe should equal to workerNum?

plugins/gitlab/tasks/gitlab_api_client.go
```diff
@@ -80,31 +111,67 @@ func (gitlabApiClient *GitlabApiClient) FetchWithPaginationAnts(resourceUri stri
...
+				page := step*conc + i
+				err := scheduler.Submit(func() error {
+					url := fmt.Sprintf(resourceUriFormat, pageSizeInt, page)
+					res, err := gitlabApiClient.Get(url, nil, nil)
+					if err != nil {
+						return err
+					}
+					handlerErr := handler(res)
+					if handlerErr != nil {
+						return handlerErr
```
likyh, line 130: when got a error, maybe we need a `c <- false`?

plugins/gitlab/tasks/gitlab_api_client.go
```diff
@@ -36,23 +37,47 @@ func CreateApiClient() *GitlabApiClient {
...
+	}
+	rateLimitResetTime, err := http.ParseTime(res.Header.Get("ratelimit-resettime"))
+	if err != nil {
+		return 0, 0, err
+	}
+	rateLimitInt, err := strconv.Atoi(rateLimit)
+	if err != nil {
+		return 0, 0, err
+	}
+	rateLimitPerSecond := rateLimitInt / int(rateLimitResetTime.Unix()-date.Unix()) * 9 / 10
```
likyh, line 79: why  `* 9 / 10`?

plugins/gitlab/tasks/gitlab_api_client.go
```diff
@@ -36,23 +37,47 @@ func CreateApiClient() *GitlabApiClient {
...
+	if total != "" {
+		totalInt, err = convertStringToInt(total)
+		if err != nil {
+			return 0, 0, err
+		}
 	}
 	logger.Info("JON >>> totalInt", totalInt)
-	return totalInt, nil
+
+	rateLimit := res.Header.Get("ratelimit-limit")
```
likyh: ratelimit-remaining

plugins/gitlab/tasks/gitlab_api_client.go
klesh: Leave some space to avoid hitting the limit

plugins/gitlab/tasks/gitlab_api_client.go
klesh: no, that would hit the limit. rateLimit is a better choice. but it is good enough for now, we can do better after Tomorrow

plugins/gitlab/tasks/gitlab_api_client.go
klesh: no, not really, one page failed should be break the whole process.

likyh (approved):
LGTM

PR #361, opened 2021-09-08
Fix deadlock by only allowing the last page in each batch to send a single message

hezyin:
# Summary

There's a deadlock in collecting Gitlab commits. This is due to that multiple goroutines in `FetchWithPaginationAnts` can send `false` to the channel while the main for loop breaks immediately after receiving the first `false`. The fix is to only allow the last page in each batch to send a single message (whether `true` or `false`) to the main for loop.

### Key Points


kevin-kline (approved):
Looks great, good job @hezyin :D

joncodo:
LGTM

PR #362, opened 2021-09-08
fix: added index to model fields, rm extra fields

kevin-kline:
### Description
We have removed foreign keys, but we need to add an index on those fields.

We also can clean them up by removing the additional GORM syntax (ie. MergeRequest int) for MergeRequestId foreign key.


klesh (approved):
LGTM

PR #370, opened 2021-09-10
docs: add lake-cli usage to user setup

narrowizard:
# Summary
Add lake-cli usage (cron sync job) to user setup

### Key Points


hezyin (approved):
LGTM

PR #374, opened 2021-09-10
fix: export env remove all comments

joncodo:
This failed before in .env

```
DB_URL=foobar # nnnnnn
```

Since the script only removes lines that start with '#'. I changed it to remove end of line comments too.

`export $(grep -v '^#' .env | sed 's/#.*$//g' | xargs)`

Screen Shot 2021-09-10 at 2 48 27 PM


hezyin (approved):
LGTM

PR #377, opened 2021-09-12
refactor: optimize building and configuration

liangjingyang:
# Summary
Optimize building and configuration.

### Key Points

- [ ] Need to rebuild docker image, mericodev/lake-builder and mericodev/lake

### Description
1. Remove zoneinfo.zip and add timezone info file from alpine package.
2. Change default mysql hostname in .env.example from `localhost` to `mysql` for compatibility with docker-compose.
3. Refactor config package to load .env file first, then override with environments.
4. Optimize building output path of core and plugins, and refactor the related configurations and scripts.


.env.example
```diff
@@ -1,13 +1,14 @@
-###################################
-# Lake Database Connection String #
-###################################
+#############
+# Lake core #
+#############
 
-DB_URL=merico:merico@tcp(localhost:3306)/lake?charset=utf8mb4&loc=Asia%2fShanghai&parseTime=True
+# Lake plugin dir, absolute path or relative path
+PLUGIN_DIR=bin/plugins
```
klesh, line 6: plugin may have grafana dashboard or other static resource, so, one `.so` one plugin might not work, maybe we should keep the one folder for one plugin logic

liangjingyang: Changed to compile each plugin into its own output dir, and the plugin loader will walk through `bin/plugins` to find all `.so` file to load.

.github/workflows/test-e2e.yml
```diff
@@ -5,7 +5,7 @@ jobs:
   e2e-mysql:
     strategy:
       matrix:
-        go-version: [1.15.x]
+        go-version: [1.16.x]
```
liangjingyang, line 8: Hi @joncodo can we upgrade go-version to 1.16.x? I want to use [WalkDir](https://pkg.go.dev/path/filepath#WalkDir) function which added in go 1.16. If there is some reason that we must use 1.15.x, I can fall back to use [Walk](https://pkg.go.dev/path/filepath#Walk) instead.

klesh (approved):

.github/workflows/test-e2e.yml
hezyin: Jon's not working today. I think 1.16.x is fine

PR #384, opened 2021-09-13
chore: build config-ui

narrowizard:
# Summary
This PR added Dockerfile for config-ui

### Key Points

### Why using `npm run dev`?
Because nextjs will prerender pages when building, and we don't have .env file when building the nextjs app, it causes an error on `npm run build`. So we simply use `npm run dev` to disable the prerender feature. Please update this PR if you have any better ideas.

### DIFF
--- a/config-ui/Dockerfile
+++ b/config-ui/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:14-alpine
+
+COPY package.json /src/package.json
+WORKDIR /src
+RUN npm i
+
+COPY . /src
+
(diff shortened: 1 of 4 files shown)


oddscenes (approved):
Looks good to me. 

Finding a solution to `npm run build` and `npm start` will be nice for future versions I think (for the user). Would be cool in the future to have one command run `npm run dev` and one setup for build/npm start. I think this setup is fine as is for now though

hezyin (approved):
LGTM

PR #385, opened 2021-09-13
fix: remove timezone from db connection string to avoid connection failure

hezyin:
## Description

To fix the `MySQL connection failed: unknown timezone Asia/Shanghai` error, this PR temporarily removes the timezone info from the connection string.


oddscenes (approved):
Looks good

PR #386, opened 2021-09-13
feat: trigger ui and doc adjustments

oddscenes:
# Summary

### Key Points

### Description
Main update to this PR is a new `/trigger` page to the config ui. This allows us to trigger data collection.

Other updates:
- Updated `README.md` for user focused setup (now that we have containerized config ui)
- `Makefile` now has `make configure` and `make configure-dev`. dev will just run yarn/node. `make configure` will run through docker compose (and most recent docker hub image)

### Current Behavior
Only way to trigger data collection is through the CLI. We want to offer users (especially non-devs) a way to do it easily through the browser)

### New Behavior
Offer an option to trigger data collection through the configuration UI. Also change the docs for better user onboarding. 

### Screenshots
Screen Shot 2021-09-13 at 4 49 04 PM

### Other Information
- We will need to update docker hub image (config-ui) to see the latest changes for user setup (running `make configure`)
- Testing of this PR would involve mainly changing config info in the `localhost:4000/triggers` page. and monitoring what data is collected in the `devlake` logs
- Future will be good to have `/triggers` page using form/input elements and not just raw json in a textarea. If we extracted the data to inputs, it would be a lot easier on the user. We could also offer both, input/forms and raw json editing
- Currently no visual indicator for data being collected on button press (specifically in browser, on `/trigger` page). We should add in some indication somehow in the UI. But I think that can be another PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ configure-dev:
 	cd config-ui; yarn; npm run dev;
 
 compose:
-	docker-compose up grafana devlake
+	docker-compose up grafana
 
 compose-down:
 	docker-compose down
--- a/Makefile
+++ b/Makefile
@@ -18,10 +18,13 @@ run:
 	go run main.go
 
 configure:
+	docker-compose up config-ui
+
+configure-dev:
 	cd config-ui; yarn; npm run dev;
 
 compose:
-	docker-compose up grafana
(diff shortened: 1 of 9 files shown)


hezyin (approved):
LGTM

PR #390, opened 2021-09-14
fix: envvar typo

klesh:
# Summary

config-ui progress not working when launched by `docker-compose up`

### Key Points


hezyin (approved):
LGTM

PR #406, opened 2021-09-16
feat: gitlab pipelines

joncodo:
Fixes: https://github.com/merico-dev/lake/issues/405

(Please see the ticket for more details)

@Startrekzky After this is merged, you can make us some graphs for gitlab ci?

Here is the data we collect:

Screen Shot 2021-09-16 at 9 46 40 AM

Note: we added this to the current gitlab plugin since it is pulled from the same api.

### DIFF
--- a/plugins/gitlab/gitlab.go
+++ b/plugins/gitlab/gitlab.go
@@ -31,50 +31,46 @@ func (plugin Gitlab) Execute(options map[string]interface{}, progress chan<- flo
 
 	progress <- 0.1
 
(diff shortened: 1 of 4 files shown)


joncodo:
I have now updated this with new data:
Screen Shot 2021-09-17 at 3 13 43 PM

It has to fetch the builds individually and can fetch 1800/min so if you have a lot, it will add time to running gitlab plugin.

@klesh Ready for review

kevin-kline (approved):
LGTM

PR #422, opened 2021-09-20
feat: domain specific dashboards but not loaded properly

joncodo:
# Summary

We need a way for users to experience Grafana if they do not analyze all data sources. For instance, if they just analyze Jira, we can only show a few working charts since a lot of charts rely on multiple data sources to be completed.

I have added three new dashboards:

- Jira
- Gitlab
- Jenkins

Each one is a collection of working charts if you only enable one data source.

To do this, I converted all existing charts to library components so if the original gets updated, they all get updated. You can also now add charts to a dashboard from a library of charts which made this task faster.

## Questions

1. It seems like I can't upload my Library Panels so others can use them? When @kevin-kline pulled this down, he could not see any panels

# Related Issue

https://github.com/merico-dev/lake/issues/408

### DIFF
--- a/grafana/dashboards/Gitlab.json
+++ b/grafana/dashboards/Gitlab.json
@@ -1,0 +1,1696 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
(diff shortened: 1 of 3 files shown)


kevin-kline:
ERROR:

image

joncodo:
This is the request that gets them from somewhere...
Screen Shot 2021-09-20 at 11 22 21 AM
Screen Shot 2021-09-20 at 11 22 15 AM

joncodo:
This feature of Grafana is not ready yet I think: https://community.grafana.com/t/provisioning-for-panel-library/47831

I will make the charts manually I guess :)

kevin-kline (approved):
LGTM

PR #427, opened 2021-09-22
test: review rounds

joncodo:
# Summary

When we wrote the review rounds code for gitlab, we did not get a chance to add tests to it. I have added these tests here.

I also added a case to handle when there are no notes or comments in the case where a pull request is just merged. This counts as a single review. We assume that the person who merged the code reviewed it.


kevin-kline (approved):
LGTM

PR #439, opened 2021-09-24
feat: set OriginKey as primary to support indate

klesh:
# Summary

@joncodo @kevin-kline 
I'was working on `jira-domain` plugin, had this `indate` operation problem,  maybe we could set `OriginKey` as PrimaryKey so we could do sth like:

```golang
err = lakeModels.Db.Clauses(clause.OnConflict{
	UpdateAll: true,
}).Create(jiraIssue).Error
```

any thought?

### Key Points


hezyin (approved):
LGTM

PR #440, opened 2021-09-26
Jira Domain Plugin

klesh:
# Summary

Convert Jira Entities to Domain Entites:
- jira_board -> board
- jira_issue -> issue
- jira_changelog -> changelog

### Key Points

### DIFF
--- a/plugins/domainlayer/okgen/originkey_generator.go
+++ b/plugins/domainlayer/okgen/originkey_generator.go
@@ -1,0 +1,80 @@
+package okgen
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/merico-dev/lake/plugins/core"
+)
+
+type OriginKeyGenerator struct {
+	prefix  string
+	pkNames []string
+	pkTypes []reflect.Type
+}
+
+func walkFields(t reflect.Type, pkNames *[]string, pkTypes *[]reflect.Type) {
+	for i := 0; i < t.NumField(); i++ {
+		field := t.Field(i)
+		if field.Type.Kind() == reflect.Struct {
+			walkFields(field.Type, pkNames, pkTypes)
+		} else {
+			gormTag := field.Tag.Get("gorm")
+
+			// TODO: regex?
+			if gormTag != "" && strings.Contains(strings.ToLower(gormTag), "primarykey") {
+				*pkNames = append(*pkNames, field.Name)
+				*pkTypes = append(*pkTypes, field.Type)
+			}
+		}
+	}
+}
+
+func NewOriginKeyGenerator(entityPtr interface{}) *OriginKeyGenerator {
+	v := reflect.ValueOf(entityPtr)
+	if v.Kind() != reflect.Ptr {
+		panic("entityPtr is not a pointer")
+	}
+	t := reflect.Indirect(v).Type()
+
+	// find out which plugin holds the entity
+	pluginName, err := core.FindPluginNameBySubPkgPath(t.PkgPath())
+	if err != nil {
+		panic(err)
+	}
+	// find out entity type name
+	structName := t.Name()
+
+	// find out all primkary keys and their types
+	pkNames := make([]string, 0, 1)
+	pkTypes := make([]reflect.Type, 0, 1)
+
+	walkFields(t, &pkNames, &pkTypes)
+
+	if len(pkNames) == 0 {
+		panic(fmt.Errorf("no primary key found for %s:%s", pluginName, structName))
+	}
+
+	return &OriginKeyGenerator{
+		prefix:  fmt.Sprintf("%s:%s", pluginName, structName),
+		pkNames: pkNames,
+		pkTypes: pkTypes,
+	}
+}
+
+func (g *OriginKeyGenerator) Generate(pkValues ...interface{}) string {
+	originKey := g.prefix
+	for i, pkValue := range pkValues {
+		// type checking
+		if reflect.ValueOf(pkValue).Type() != g.pkTypes[i] {
+			panic(fmt.Errorf("primary key type does not match: %s should be %s",
+				g.pkNames[i],
+				g.pkTypes[i].Name(),
+			))
+		}
+		// append pk
+		originKey += ":" + fmt.Sprintf("%v", pkValue)
+	}
+	return originKey
+}
--- a/go.mod
+++ b/go.mod
@@ -50,6 +50,11 @@ require (
 )
 
 require (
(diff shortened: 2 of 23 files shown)


plugins/domainlayer/okgen/originkey_generator.go
```diff
@@ -0,0 +1,80 @@
...
+	pkTypes []reflect.Type
+}
+
+func walkFields(t reflect.Type, pkNames *[]string, pkTypes *[]reflect.Type) {
+	for i := 0; i < t.NumField(); i++ {
+		field := t.Field(i)
+		if field.Type.Kind() == reflect.Struct {
+			walkFields(field.Type, pkNames, pkTypes)
+		} else {
+			gormTag := field.Tag.Get("gorm")
```
narrowizard, line 23: Do we strongly depend on gorm?

plugins/domainlayer/okgen/originkey_generator.go
```diff
@@ -0,0 +1,80 @@
...
+
+func NewOriginKeyGenerator(entityPtr interface{}) *OriginKeyGenerator {
+	v := reflect.ValueOf(entityPtr)
+	if v.Kind() != reflect.Ptr {
+		panic("entityPtr is not a pointer")
+	}
+	t := reflect.Indirect(v).Type()
+
+	// find out which plugin holds the entity
+	pluginName, err := core.FindPluginNameBySubPkgPath(t.PkgPath())
```
narrowizard, line 42: Can we use PkgPath instead of pluginName? Consider that we may register plugins distributly. A plugin may not always exist in a `plugin/core` instance.

narrowizard:
Does domainlayer a plugin or a feature of devlake? If it is a plugin  
- jiradomain called `NewOriginKeyGenerator` method from domainlayer plugin. Is it ok to called plugin's methods from each other?

plugins/domainlayer/okgen/originkey_generator.go
klesh: What do u mean by distributly? 
The `PkgPath` is a good approach, but it's kinda too long, and the `OriginKey` must be unique indexed, I'm kinda worry about the  column length and the performance.
To clarify, do u mean `OriginKey` to be like `github.com/merico-dev/lake/plugins/jira/models.JiraIssue:4234` ?


@hezyin @joncodo @kevin-kline  which way do u like?

klesh:
> Does domainlayer a plugin or a feature of devlake? If it is a plugin
> 
> * jiradomain called `NewOriginKeyGenerator` method from domainlayer plugin. Is it ok to called plugin's methods from each other?

🤔
Why not? What is your concern?
How to refer the origin data source record should be defined and applied universally, it makes sense to define it in `domainlayer` plugin.
Why we can use `models` from another plugin but not other? What is the alternative you suggest?

plugins/domainlayer/okgen/originkey_generator.go
klesh: I think we can depend on gorm for now, to simply things a bit, unless we see a reason not to.

narrowizard:
> > Does domainlayer a plugin or a feature of devlake? If it is a plugin
> > 
> > * jiradomain called `NewOriginKeyGenerator` method from domainlayer plugin. Is it ok to called plugin's methods from each other?
> 
> 🤔
> Why not? What is your concern?
> How to refer the origin data source record should be defined and applied universally, it makes sense to define it in `domainlayer` plugin.
> Why we can use `models` from another plugin but not other? What is the alternative you suggest?

Even models is not suggest to use from each other.  
- When we build a plugin in `go plugin` mode, it will compile everything imported in the plugin. It means `B.so` will include A plugin's code if B plugin reference something in A plugin. What if `A` plugin is updated? Should we rebuild every plugin that reference A plugin? How can we find these plugins as our plugin number grows up.

plugins/domainlayer/okgen/originkey_generator.go
narrowizard: > What do u mean by distributly?

Task runner in devlake should be deployed in multi nodes. We may not register all plugins in every node. If so, we may not find plugin name with `core.FindPluginNameBySubPkgPath`. 

Anyway, I just think get plugin name in this way is unnecessary(it make complex to call plugin/core from plugin). Even if add a parameter `pluginName` when create generator is better?

klesh:
> > > Does domainlayer a plugin or a feature of devlake? If it is a plugin
> > > 
> > > * jiradomain called `NewOriginKeyGenerator` method from domainlayer plugin. Is it ok to called plugin's methods from each other?
> > 
> > 
> > thinking
> > Why not? What is your concern?
> > How to refer the origin data source record should be defined and applied universally, it makes sense to define it in `domainlayer` plugin.
> > Why we can use `models` from another plugin but not other? What is the alternative you suggest?
> 
> Even models is not suggest to use from each other.
> 
> * When we build a plugin in `go plugin` mode, it will compile everything imported in the plugin. It means `B.so` will include A plugin's code if B plugin reference something in A plugin. What if `A` plugin is updated? Should we rebuild every plugin that reference A plugin? How can we find these plugins as our plugin number grows up.

OK, How do we solve the problem then?

plugins/domainlayer/okgen/originkey_generator.go
klesh: I see your point. But what is the necessity of devlake being a Distributed System? Up to now, I see no proof of that.
Anyway I like your idea for the `OriginKey`, it's simpler, maybe we could do it in next iteration.
As for now, today is the deadline for feature devlopment, so, we'll keep it the way it is for now, is it ok for you?

klesh:
I like the questions you raised, but we also need some solutions for those, would you mind post a Proposal in more depth with both problems and solutions? and we'll address those in future iterations.
@narrowizard

narrowizard:
> I like the questions you raised, but we also need some solutions for those, would you mind post a Proposal in more depth with both problems and solutions? and we'll address those in future iterations.
> @narrowizard

In general, we prefer every plugin to be independent. As for detail implement, I have no very good idea now. I'll create an issue to clarify the problem first.

plugins/domainlayer/okgen/originkey_generator.go
narrowizard: > is it ok for you?

It's ok.  

Maybe the word `distributly` is not so exact, we hope to deploy multi `lake worker` to speed up task running.

joncodo (approved):
I would like to see more comments and more details in the PR description but I will merge it.

PR #447, opened 2021-09-27
feat: github performance

joncodo:
# Summary

This work allows us to do a few things differently when it comes to github. The main problem is that their api only allows us 5000 requests/minute.

Because of this, we need to be really careful how many requests we make and how quickly we make them.

## Tasks covered

- DONE - @Jon Performance - Remove the get total requests and use sequential for next page header
- DONE - @Jon Performance - **Refactor the one network scheduler
- DONE - @Jon Performance set the rate ants rate to (number of tokens*1 = 3600 REQUESTS PER HOUR 
- Remove the non ants requests


kevin-kline (approved):
Changes are made in future PR

PR #475, opened 2021-09-30
Fix config UI again

klesh:
# Summary

Fixes issues from https://github.com/merico-dev/lake/issues/474

### DIFF
--- a/config-ui/pages/plugins/jira/index.page.jsx
+++ b/config-ui/pages/plugins/jira/index.page.jsx
@@ -2,14 +2,12 @@ import Head from 'next/head'
 import { useState, useEffect } from 'react'
 import styles from '../../../styles/Home.module.css'
 import {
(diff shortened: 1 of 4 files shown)


config-ui/pages/plugins/jira/index.page.jsx
```diff
@@ -30,22 +47,56 @@ export default function Home(props) {
...
+          Rejected: [],
+        }
+      }
+    ])
+    setCustomStatusOverlay(false)
+    e.preventDefault()
+  }
 
 
   function updateEnv(key, value) {
```
narrowizard, line 102: If value is undefined, it'll be saved to .env file like 'JIRA_ISSUE_STORYPOINT_COEFFICIENT=undefined', Pls check the key and value before request. @oddscenes   

## Reproduce
If there is not a key called 'JIRA_ISSUE_STORYPOINT_COEFFICIENT' in the original `.env` file, the value will be undefined.

narrowizard (approved):

PR #478, opened 2021-10-04
feat: cancel running tasks

joncodo:
# Summary

We want to cancel tasks while they are running on the server. This is so we can stop long running processes with ants. 

## New technology we use

Simple golang event bus - https://nats.io/

## How the code works:
0. Run docker compose with the new nats server.
1. Run a task for collection with Github
2. Cancel the task by POST to `localhost:8080/task/cancel?taskName=github`
3. The api call hits a route that sends a NATS message to a subscriber in the github plugin
4. The plugin subscriber calls `scheduler.Release()` (An ants function to stop the ants worker pool)
5. See the logs stop collecting data

TODO:
- [ ] All plugins need to implement this in their execute function

@klesh @hezyin looking for feedback


hezyin:
@joncodo @kevin-kline I haven't dug too deep into this, but it seems like Golang's built-in `Context` is designed to solve this problem: https://stackoverflow.com/a/59608070

klesh:
@joncodo shall we use `Context` instead of `nats`, it seems too bloated...

joncodo:
> @joncodo shall we use `Context` instead of `nats`, it seems too bloated...

I agree for sure. I tried to use it but could not get it at first. I will try to get it again on Tuesday.

PR #585, opened 2021-11-01
Multi token client

joncodo:
## Description

Ticket: https://github.com/merico-dev/lake/issues/567

You can also see the README in this PR changes

### DIFF
--- a/plugins/github/README.md
+++ b/plugins/github/README.md
@@ -1,0 +1,25 @@
+# Github Pond
+
+## Metrics
+
+Currently the data is only fetched and stored in the DB. Soon we will have charts in Grafana to support this data.
+
+## Github rate limits
+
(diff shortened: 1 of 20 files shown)


joncodo:
@klesh updated and working.

Tested locally with one token and three tokens. Logged out the auth header to make sure it was using all three tokens

plugins/github/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+In your .env file, you will need to set up
+
+```
+
+GITHUB_AUTH=XXX
+
+or...
+
+GITHUB_AUTH=XXX,YYY,ZZZ // where each token is a different user's token (optional)
```
joncodo, line 27: @klesh this is how you set tokens so you can use multiples

klesh (approved):
LGTM

PR #598, opened 2021-11-02
Config UI Refactor - Connection Limiting & Integration Bugfixes

e2corporation:
# Connection Limiting and use $ID Primary Key
> **WIP** DO NOT MERGE

### Key Points

- [x] Implement **Connection Limit** Feature so that providers w/o Multi-data support can be restricted until backend is able to offer these features.
    - **JIRA** Supports Multiple Connections
    - Jenkins and Gitlab Limited to 1 Connection Only
    - Add Connection Functionality LOCKED when limit reached
- [x] Use **$ID** as primary key parameter for connection sources
- [x] Fix Loading of JIRA Settings
- [x] Update Connection Manager Hook
  - Add routine for fetching All Connections
  - Add Logic for Connection Limiting

### Description
This PR is to continue the **Configuration UI Refactor** with regard to Multi-data connection sources UX. As JIRA is the only provider at this time that supports multi-connection sources, a connection limit feature has been added so that Gitlab and Jenkins will have restricted modes limiting them to only 1 instance.

### Screenshots
Screen Shot 2021-11-02 at 5 29 03 PM
Screen Shot 2021-11-02 at 4 44 28 PM

### DIFF
--- a/config-ui/src/data/NullConnection.js
+++ b/config-ui/src/data/NullConnection.js
@@ -1,0 +1,23 @@
+const NullConnection = {
+  // id: null,
+  ID: null,
+  name: null,
+  endpoint: null,
+  token: null,
+  username: null,
+  password: null,
+  basicAuthEncoded: null, // NOTE: we probably want to exclude/null this when exposing this object
+  JIRA_ISSUE_TYPE_MAPPING: null,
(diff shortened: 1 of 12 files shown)


joncodo:
LGTM

PR #664, opened 2021-11-05
Partial fix to GitLab Config-UI 

kevin-kline:
### Description
This PR contains a partial fix to this problem: https://github.com/merico-dev/lake/issues/644

This PR repairs the data flow between the frontend and backend. However, it reveals a new problem on the backend. There is an issue with Viper reading the current values in the .env file after it's been written to. 
The problem may exist with this function:

```V := config.LoadConfigFile()```

Not yet.

### DIFF
--- a/config-ui/src/hooks/useConnectionManager.jsx
+++ b/config-ui/src/hooks/useConnectionManager.jsx
@@ -239,6 +239,7 @@ function useConnectionManager ({
           setPassword(activeConnection.password)
           break
         case Providers.GITLAB:
+          setToken(activeConnection.basicAuthEncoded || activeConnection.Auth)
         case Providers.JIRA:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: config-ui/src/hooks/useConnectionManager.jsx [hunk 1]
```diff
@@ -239,6 +239,7 @@ function useConnectionManager ({
           setPassword(activeConnection.password)
           break
         case Providers.GITLAB:
+          setToken(activeConnection.basicAuthEncoded || activeConnection.Auth)
```
e2corporation, line 242: @kevin-kline This line should not be necessary, both Gitlab and JIRA share same case flow.

kevin-kline: Maybe it shouldn't, but it did fix the behaviour for me. Either way I don't think it can hurt.

PR #668, opened 2021-11-08
fix: jira sprint collection incompleteness

### LINKED ISSUE #665: Jira Sprint missing records
## Describe the bug
@runjinz  reported some latest sprints are not being collected.
After some investigate, I found out the `maxResults = 100` parameter wont work for this endpoint, seems like the maximum value for this parameter is 50:
image

## To Reproduce
1. Trigger jira sprint collection on board 8
2. Check database `jira_sprints` have no sprint 59 and 60

## Expected behavior
All sprint should be collected

klesh:
# Summary

Closes #665 

### Key Points

### Description
We assumed all APIs accept 100 as maximal value for `maxResults` parameter, which is not true.
API like `agile sprints` is configurable, and it was 50 on our instance.
So use a fixed value like 100 to increase our `startAt` parameter is unreliable, and causing some sprints were missed out during collection.
This PR fix this issue by updating the `maxResults` value dynamically base on the `length` we actually received, and use this value  to increase `startAt` parameter.

Closes #665

### DIFF
--- a/plugins/jira/tasks/jira_api_client.go
+++ b/plugins/jira/tasks/jira_api_client.go
@@ -48,7 +48,7 @@ type JiraPagination struct {
 }
 
 type JiraPaginationHandler func(res *http.Response) error
-type JiraSearchPaginationHandler func(res *http.Response) (bool, error)
+type JiraSearchPaginationHandler func(res *http.Response) (int, error)
 
 func (jiraApiClient *JiraApiClient) FetchPages(scheduler *utils.WorkerScheduler, path string, query *url.Values, handler JiraPaginationHandler) error {
 	if query == nil {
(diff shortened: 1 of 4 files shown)


plugins/jira/tasks/jira_api_client.go
```diff
@@ -123,37 +123,39 @@ func (jiraApiClient *JiraApiClient) FetchWithoutPaginationHeaders(
...
+	// `maxResults` and `length` into one variable
+	for maxResults > 0 {
 		// get page
-		query.Set("startAt", strconv.Itoa(nextStartTmp))
+		query.Set("startAt", strconv.Itoa(startAt))
 		res, err := jiraApiClient.Get(path, query, nil)
 		if err != nil {
 			return err
 		}
 		if res.StatusCode == 401 {
```
mindlesscloud, line 148: Does `res.Body` will be closed?

mindlesscloud (approved):
👍

PR #674, opened 2021-11-08
Adds watching to read/write config with viper

kevin-kline:
### Description
We need to watch for changes to the config file using Viper in order to read/write values and serve them to the front end.

https://github.com/merico-dev/lake/issues/671

### Current Behavior
No read/write capability. Writes only.

### New Behavior
Enables reading and writing for config file.


joncodo (approved):
LGTM

Issue #680, opened 2021-11-08
Config-UI: Jira subtitle (connection name) doesn't refresh when saved.


kevin-kline:
## Describe the bug
If you save a new connection name, the connection name doesn't update in the subtitle (or breadcrumb) until you manually refresh.

## To Reproduce
1. Go to http://localhost:4000/connections/configure/jira/1
2. Edit "Connection Name" field
3. Click "Save Connection"
4. See error as described.

## Expected behavior
All fields are refreshed.

## Screenshots
![image](https://user-images.githubusercontent.com/27032263/140806761-34b55aff-7353-4ced-a84b-cd7868840fdf.png)

e2corporation:
Nice find, we need to retrigger a call to `fetchConnection` routine once the save successfully completes so that the `activeConnection` object is refreshed.

PR #683, opened 2021-11-08
#680 Config UI: Connections - Refresh $activeConnection Object on SAVE

e2corporation:
### Key Points

- [x] Update Connection Manager Hook
  - On `saveComplete` Success, refresh the `activeConnection` object
- [x] Relocate `Providers` constant to resolve React console warning(s)
- [x] JSX Formatting Cleanups

### Description
This PR refreshes **Active Connection Object** being modified so that updated parameters such as the source instance's Name are properly reflected in the title and breadcrumbs.

#680 

### Current Behavior
When a user updates a connection instance details, the new values aren't automatically refreshed which may cause confusion. If page is reloaded new properties are shown.

### New Behavior
When a user updates a connection instance, the details are seamlessly refreshed so the updated Instance Name (and other properties) are properly reflected.

### DIFF
--- a/config-ui/src/hooks/useConnectionManager.jsx
+++ b/config-ui/src/hooks/useConnectionManager.jsx
@@ -7,6 +7,12 @@ import { DEVLAKE_ENDPOINT } from '@/utils/config'
 import request from '@/utils/request'
 import { NullConnection } from '@/data/NullConnection'
 
+const Providers = {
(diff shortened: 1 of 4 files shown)


e2corporation:
@joncodo @kevin-kline This one is ready for `main`. Thanks!

PR #701, opened 2021-11-10
feat: secure api with token

mindlesscloud:
# Summary

Secure api with token

### Key Points

### Description
The table `auth_tokens` stores API tokens which would be checked against for every request

https://github.com/merico-dev/lake/issues/551

### Current Behavior
APIs could be accessed publicly

### New Behavior
The API caller should set his/her token to the header `X-Lake-Token`. If the authentication fails, HTTP status code `401` would be returned from the server.

### Screenshots

### Other Information


joncodo:
@mindlesscloud can this be closed?

PR #703, opened 2021-11-10
fix: jira source detail returning malformed typeMappings

### LINKED ISSUE #702: JIRA source detail api returning malform typeMapping field
## Describe the bug
Jira source detail api return a malform `typeMappings` causing `config-ui` to break if any status mapping were set

## To Reproduce
1. Add a `statusMapping` to any source
2. Click on that source on `config-ui`
3. A blank page with error message like following would appear
image

## Expected behavior
Should be able to operate on data source with or without `statusMappings` being set

## Screenshots
image

klesh:
Closes #702

Check issue #702 for detail

### DIFF
--- a/config-ui/src/pages/configure/settings/jira.jsx
+++ b/config-ui/src/pages/configure/settings/jira.jsx
@@ -150,7 +150,7 @@ export default function JiraSettings (props) {
     // }
     if (connection && connection.ID) {
       // Parse Type Mappings (V2)
-      parseTypeMappings(connection.TypeMappings)
+      parseTypeMappings(connection.typeMappings)
 
       // LEGACY TYPE MAPPINGS (Disabled)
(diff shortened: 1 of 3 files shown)


config-ui/src/pages/configure/settings/jira.jsx
```diff
@@ -150,7 +150,7 @@ export default function JiraSettings (props) {
     // }
     if (connection && connection.ID) {
       // Parse Type Mappings (V2)
-      parseTypeMappings(connection.TypeMappings)
+      parseTypeMappings(connection.typeMappings)
```
e2corporation, line 153: Current API Response from BE on `main` branch is showing that the response key starts with uppercase T `TypeMappings` .... This feature was already working without changes to this line, is the problem only backend related in that case?

Screen Shot 2021-11-10 at 8 12 20 AM

klesh: The BE also been updated to lower case t.

PR #705, opened 2021-11-10
fix: all plugins apis returning 201

### LINKED ISSUE #704: All plugin apis returning status 201
## Describe the bug
http status code represents `CREATED` which is only suitable for method like `POST` , and yet now all plugin apis returning 201 even for `GET` requests

## To Reproduce
1. Send requests to any jira plugin api
2. All apis returning status code 201

## Expected behavior
Returning 201 if resources created successfully, and 200 for other requests.
image

## Screenshots
image

klesh:
Closes #704

Please check issue for detail

### DIFF
--- a/plugins/jira/api/jira_issue_status_mapping.go
+++ b/plugins/jira/api/jira_issue_status_mapping.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"fmt"
+	"net/http"
 
 	"github.com/go-playground/validator/v10"
 	lakeModels "github.com/merico-dev/lake/models"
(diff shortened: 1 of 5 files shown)


plugins/jira/api/jira_issue_status_mapping.go
```diff
@@ -135,7 +135,7 @@ func PostIssueStatusMappings(input *core.ApiResourceInput) (*core.ApiResourceOut
 	if err != nil {
 		return nil, err
 	}
-	return &core.ApiResourceOutput{Body: jiraIssueStatusMapping}, nil
+	return &core.ApiResourceOutput{Body: jiraIssueStatusMapping, Status: 201}, nil
```
narrowizard, line 138: Why not using `http.StatusCreated` but a constant value `201`? As used above.

klesh: Good point

PR #706, opened 2021-11-10
New GitHub enrichment

joncodo:
- lead time, assignee, priority and type on github issues and enriched domain issues.

I think we also need this for gitlab....

@Startrekzky


joncodo:
FIxes: https://github.com/merico-dev/lake/issues/690

PR #729, opened 2021-11-12
feat: github plugin proxy support

mindlesscloud:
# Summary
Support setting proxy for Github plugin

### Key Points

### Description
The proxy server address should be set in the `.env` file with the key `GITHUB_PROXY`. If the key is empty or any other invalid url, no proxy applied. Only `http` and `socks5` protocol supported for now.

https://github.com/merico-dev/lake/issues/720


klesh (approved):
LGTM

Issue #771, opened 2021-11-15
Config UI  - Webpack / NGINX / devServer Proxy + OSX Network Errors


e2corporation:
## Describe the bug
The recent change to Webpack's dev configuration so that the `/api` endpoint is now accessed via proxy is not suitable for bare metal mode on OSX environment. The result is that the Config UI App operates in "offline" mode due to the fact that all api calls fail.

PR #772, opened 2021-11-15
fix: #771 use compressed format ipv6 loopback address

e2corporation:
### Config-UI / Webpack / devServer Proxy

- [x] fix: use compressed format ipv6 loopback address

### Description
Update the `devServer` Configuration for Webpack and change the `localhost` reference to compressed format `[::1]`

#771

### Current Behavior
When pulling `main` branch OSX users are unable to fetch data from `/api` because the devServer proxy fails to work as configured resulting in network 404s.

### New Behavior
Using the compressed format for localhost this should be more compatible with more environments including OSX so developers don't have to apply a patch to the configuration in order to reach `DEVLAKE_ENDPOINT`

### DIFF
--- a/config-ui/webpack.config.js
+++ b/config-ui/webpack.config.js
@@ -118,7 +118,7 @@ module.exports = (env = {}) => {
       host: '0.0.0.0',
       historyApiFallback: true,
       proxy: {
-        '/api/': { target: 'http://localhost:8080', pathRewrite: { '^/api': '' }, changeOrigin: true },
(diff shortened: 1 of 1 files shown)


e2corporation:
@klesh Please verify that this does not cause breakages on your end. Another approach in the future is that we use injection with webpack and inject a constant `DEVLAKE_ENDPOINT` at build time based on environment. Additionally, devServer proxy is only applicable to development mode and not production mode.

e2corporation:
This should not be bundled with current release `iter4` if there's any issues, but considered for after `iter5`.

klesh (approved):
LGTM

PR #785, opened 2021-11-15
docs: update README for setup

kevin-kline:
### Description
This PR attempts to address some of my concerns with the setup flow in the README. The primary concern is in the Dev Setup, there is no reference to the Config-UI, and no guidance on how to configure plugins. This PR adds the config UI setup steps to the Dev Setup, and gives stronger reference to the links that take the user to the Plugin Configuration docs which have a lot of information they need to get the setup right.

No.


kevin-kline:
Also note that this removes a reference to the "Main Configuration" page which no longer exists.

Startrekzky (commented):
LGTM

hezyin (approved):
LGTM

PR #820, opened 2021-11-18
Refactor: remove type assertion from jira issue collector

### LINKED ISSUE #563: Jira convertIssues uses too many type assertion, should be refactor
## Description
Jira convertIssues is using too many type assertion which is problematic, because it tedious and hard to understand, and debug.

Create structs for those known fields, and use `map.Decode` to extract them.
For `customfield`, apply the type assertion with caution.

No

klesh:
# Summary

A long overdue refactor.

### Key Points

### Description

1. type assertion are error prone and hard to debug, replaced with
`mapstructure.Decode` which is easy to understand and maintain.
2. replace `sql.NullTime` with `*time.Time` for maintainability.
3. unify struct naming to help distinguish Api structs and Db structs

Closes #563

### DIFF
--- a/plugins/core/iso8601time.go
+++ b/plugins/core/iso8601time.go
@@ -3,9 +3,12 @@ package core
 import (
 	"database/sql"
 	"fmt"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
+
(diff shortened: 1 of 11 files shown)


joncodo:
LGTM

PR #821, opened 2021-11-18
refactor: unify nullable datetime to `*time.Time`

### LINKED ISSUE #599: Use `*time.Time` instead of `time.Time` for null value handling
## Description
Now we have most of our datetime fields as type `time.Time`, and using `IsZero` to check if fields were `null`. That is due to documentation we found indicated the way to handle null value in database should be wrapped inside `sql.NullTime`.

According to @mindlesscloud , gorm can handle `*time.Time` being `nil` and represent null value in database correctly.
And seems like `json.Unmarshal` could also handle the pointer correctly. 

So, it appears that using `*time.Time` is a better way to handle the null value situation. We should verify these information, and consider other factors, and see if we opt for this way of nullable time.Time handling.

1. Verify if `json` lib and `gorm` can `*time.Time` as expected
2. Change all `nullable time.Time` type to `*time.Time`  if everything work out 

No

Stay with `time.Time` with `IsZero` which is less elegant and error-prone

klesh:
# Summary

Unify nullable datetime to type `*time.Time` by replacing all `sql.NullTime`

### Key Points

### Description

1. `*time.Time` is neater and easy to work with.
2. Handle nullable time in different ways creates confusion for new team member and contributor (if any)
3. By removing `sql.NullTime` we can remove those conversion logic as well, those are liability. I would like to remove `Iso8601Time` code if we could find a better solution, like some reliable lib.
4. It's easy to do so, 15mins of coding, although collecting data for testing tooks me more than an hour thanks to GFW 😂

Closes #599


joncodo:
Now that I see this, I think you are right. Cool

PR #829, opened 2021-11-19
Handle url query strings with net/url package

kevin-kline:
### Description
This PR replaces our old method of putting together query strings with a more robust way that is less error prone. Now we are using the net/url package to resolve any inconsistencies for example with '/' at the end of the string or no '/'. This PR also includes:

- Unit testing for URI formatting
- GitHub plugin refactor to pass query params separate from relative path string
- GitLab plugin refactor for the same as GitHub

https://github.com/merico-dev/lake/issues/804


joncodo:
LGTM

Issue #827, opened 2021-11-19
Tasks deduplication


klesh:
## Description
When user trigger pipeline with identical tasks list multiple times, devlake would run those duplicated tasks in parallel, which is a big waste and took longer time for each pipeline to complete

## Describe the solution you'd like
When creating pipeline, run dedupe check on existing tasks, and reuse those tasks.

## Has the Feature been Requested Before?
No

## Additional context
EE need this to prevent situation like:
1. User click on "sync jira" and EE send out "collect jira -> compute jira/gitlab metric"
2. User then click on "sync gitlab" and EE send out "collect gitlab -> compute jira/gitlab metric" right away
3. Causing jira/gitlab got computed twice

joncodo:
This is handled in this taks: https://github.com/merico-dev/lake/issues/775

If a new task is triggered while another one of the same type is running, cancel all running tasks of the same type and only execute the new one.

klesh:
@joncodo I suggest we reopen this one, and close #775 instead.
for automatical cancelling, we need to consider more factors.

joncodo:
@klesh feel free to change anything you need. I trust your ideas :)

klesh:
Now we had 2d array pipeline, this is no longer an issue. close for now.

PR #830, opened 2021-11-19
feat: cancel duplicate tasks

joncodo:
https://github.com/merico-dev/lake/issues/775

This makes it so when you trigger a task, if there are other tasks of the same type, they will be cancelled.

### DIFF
--- a/api/task/task.go
+++ b/api/task/task.go
@@ -27,12 +27,35 @@ func Post(ctx *gin.Context) {
 	// Return all created tasks to the User
 	ctx.JSON(http.StatusCreated, tasks)
 
+	// cancel all previous runing tasks of the same type
+	cancelAllDuplicateTasks(tasks)
+
 	go func() {
 		err := services.RunAllTasks(data, tasks)
 		if err != nil {
 			_ = ctx.AbortWithError(http.StatusInternalServerError, err)
 		}
 	}()
+}
+
+func cancelAllDuplicateTasks(tasks [][]models.Task) error {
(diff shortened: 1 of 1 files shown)


api/task/task.go
```diff
@@ -35,6 +38,25 @@ func Post(ctx *gin.Context) {
...
+func cancelAllDuplicateTasks(tasks [][]models.Task) {
+	for i := 0; i < len(tasks); i++ {
+		for j := 0; j < len(tasks[i]); j++ {
+			task := tasks[i][j]
+			// get task by id
+			var taskModel models.Task
+
+			db := models.Db.Model(&models.Task{}).Where("id = ?", task.ID)
+			_ = db.Debug().Find(&taskModel)
+
```
joncodo, line 50: Need to run the cancel command too.

func CancelTask(taskId uint64) error {
	if cancel, ok := runningTasks[taskId]; ok {
		logger.Info("cancel task ", taskId)
		cancel()
		delete(runningTasks, taskId)
	} else {
		return fmt.Errorf("unable to cancel task %v", taskId)
	}
	return nil
}

klesh (commented):
Hold on, we need to consider more on this.
This implementation would cause previous tasks fail, which is not ideal.
Please take a closer look to #827, from the "Additional context" section.
If we just cancelled all previous tasks, that make previous "sync" operation failed on EE side because the status would became "CANCELLED" (btw, this status should be recorded in `services/task.go`).

joncodo:
Changing this to WIP since it is on hold

PR #839, opened 2021-11-22
add lint to ci for Frontend

joncodo:
Fixes https://github.com/merico-dev/lake/issues/618
Screen Shot 2021-11-22 at 9 41 31 AM

Note the new CI step

### DIFF
--- a/config-ui/package.json
+++ b/config-ui/package.json
@@ -7,7 +7,7 @@
     "build-production": "webpack --config webpack.production.config.js --mode production",
     "start": "webpack serve --progress",
     "start-production": "npm run build-production;node server.js",
-    "lint": "eslint ./src"
+    "lint": "eslint ./src --fix"
   },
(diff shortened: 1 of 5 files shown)


config-ui/package.json
```diff
@@ -7,7 +7,7 @@
     "build-production": "webpack --config webpack.production.config.js --mode production",
     "start": "webpack serve --progress",
     "start-production": "npm run build-production;node server.js",
-    "lint": "eslint ./src"
+    "lint": "eslint ./src --fix"
```
e2corporation, line 10: I think there should be 2 scripts here, one that runs lint without fixing so developer can see changes before code is automatically touched with fix, and one that runs lint+fixing.

```
"lint": "eslint ./src",
"lint:fix": "eslint ./src --fix"
```

e2corporation (approved):
LGTM

Issue #804, opened 2021-11-17
GitHub endpoint must handle url with or without '/'


mindlesscloud:
## Describe the bug
The endpoint must end with `/` in config-ui, otherwise it would result in a DNS `no such host` error.

When a user edits this in config ui, they have to set end the url with trailing slash. The backend should handle this.

## Screenshots
![image](https://user-images.githubusercontent.com/8455907/142129736-a8a58fc2-eb3c-4b52-bbb1-c60bd2b04ea7.png)

## Additional context
a better way to concatenate url path elements:
```go
endpoint := "https://api.github.com" // This is the value that comes from config-ui
repo := "repo/merico-dev/lake"
u, err := url.Parse(endpoint)
if err != nil{
    // check err
}
u.Path = path.Join(u.Path, repo)
u.String() // the result url
```

klesh:
I think we should do it on iter5

joncodo:
## Assumptions

1. I think we should do this for all urls in all plugins since this could cause problems for other plugins too.
2. There should a core util library for stuff like this so we can share code across plugins?

## Questions

1. I do not see this problem for me locally. When I add a trailing slash in postman, the queries fail. How can I validate this work locally to prove that this is the right way to do things? @mindlesscloud 

FAILS:
<img width="530" alt="Screen Shot 2021-11-17 at 8 24 59 AM" src="https://user-images.githubusercontent.com/3011407/142200197-c60f1c53-9274-4c62-8c85-e65d016f6d7f.png">

WORKS:
<img width="613" alt="Screen Shot 2021-11-17 at 8 25 39 AM" src="https://user-images.githubusercontent.com/3011407/142200265-63570cba-be79-4291-82ab-bc350b44177a.png">

joncodo:
The actual issue we need to solve

- [ ] github and gitlab need to pass url and query params into the DO method
- [ ] The DO method needs to handle this with the "net/url" package


Something like this:

```
package main

import (
    "fmt"
    "log"
    "net/url"
)

func main() {
    
    baseUrl := "https://some-domain.com"
    base, err := url.Parse(baseUrl)
    if err != nil {
        log.Fatal(err)
    }

relativeUrl := "/getsomething/"
    u, err := url.Parse(relativeUrl)
    if err != nil {
        log.Fatal(err)
    }

    queryString := u.Query()
    queryString.Set("id", "1")
    u.RawQuery = queryString.Encode()


    fmt.Println(base.ResolveReference(u))
}

```

PR #840, opened 2021-11-22
fix: handle / or no / on baseUrl

kevin-kline:
### Description
This PR adds a method which allows ApiClient to handle base URLs with or without '/' at the end. It also adds unit tests to prove functionality. This means it is now OK to store endpoint URLs in the DB with or without relative paths attached (for example: ```mysite.com/rest``` or ```mysite.com``` are acceptable.

**Bottom Line** this reverts some changes to Jira plugin where ```rest/``` was added to relative paths. It is now fine to go back to the way we had it by storing ```/rest``` in the DB as the Jira endpoint. In fact, we will need ```/rest``` in the DB for Jira to work.

related to #834 
related to #829 
completes #804 

### Current Behavior
If endpoint (from .env or DB) does not end with '/' then whatever is written after the last '/' will be removed from the endpoint. 

### New Behavior
If endpoint does not end with '/' then we will append a '/' on the end in order to keep what is written after the previous '/'

### DIFF
--- a/plugins/core/apiclient.go
+++ b/plugins/core/apiclient.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
@@ -178,6 +179,7 @@ func UnmarshalResponse(res *http.Response, v interface{}) error {
 }
 
 func GetURIStringPointer(baseUrl string, relativePath string, queryParams *url.Values) (*string, error) {
+	AddMissingSlashToURL(&baseUrl)
 	base, err := url.Parse(baseUrl)
 	if err != nil {
 		return nil, err
@@ -195,4 +197,12 @@ func GetURIStringPointer(baseUrl string, relativePath string, queryParams *url.V
 	}
 	uri := base.ResolveReference(u).String()
 	return &uri, nil
+}
+
+func AddMissingSlashToURL(baseUrl *string) {
+	pattern := `\/$`
+	isMatch, _ := regexp.Match(pattern, []byte(*baseUrl))
+	if !isMatch {
+		*baseUrl += "/"
+	}
 }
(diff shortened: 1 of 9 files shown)


plugins/jira/tasks/jira_board_collector.go
```diff
@@ -29,7 +29,7 @@ type JiraApiBoard struct {
 }
 
 func CollectBoard(jiraApiClient *JiraApiClient, source *models.JiraSource, boardId uint64) error {
-	res, err := jiraApiClient.Get(fmt.Sprintf("rest/agile/1.0/board/%v", boardId), nil, nil)
+	res, err := jiraApiClient.Get(fmt.Sprintf("agile/1.0/board/%v", boardId), nil, nil)
```
joncodo, line 32: can we revert this now then? And have the old value?

plugins/core/apiclient.go
```diff
@@ -196,3 +198,11 @@ func GetURIStringPointer(baseUrl string, relativePath string, queryParams *url.V
 	uri := base.ResolveReference(u).String()
 	return &uri, nil
 }
+
+func AddMissingSlashToURL(baseUrl *string) {
```
joncodo, line 202: I think this is the best solution so far.

joncodo:
LGTM but let's get @klesh to approve it too

klesh:
I prefer using standard lib function as it's a standard behave from https://stackoverflow.com/questions/53692326/convert-relative-to-absolute-urls-in-go
especial the second solution seems more appropriate.

joncodo:
We tried that. It only works in standard lib if you have a trailing slash. If you don't, it cuts off the url. Feel free to try for yourself. If the base url is http://foo.com, it renders http://foo.com. If foo.com/bar, it renders foo.com. If foo.com/bar/, it renders foo.com/bar/

joncodo:
This is 100% needed. Merging

klesh:
You are right

PR #841, opened 2021-11-22
feat: test connection

joncodo:
Fixes https://github.com/merico-dev/lake/issues/667

## Questions

So the first question here is: "Do we make it mandatory to include a /test route?"

1. Yes since every plugin needs it
2. No, just allow it to be added or not. Since the front end can still call this route and if a 404 comes back, the UI just disables the test connection button and gives a message to the user that says "Not implemented, please contact the plugin maintainer"

Can we enforce the plugin maintainer to make this route?

No. Since they can just write a route that always returns true to bypass this rule.

## Design

So I have decided that the only thing we need to do for this ticket is to simply add 

```
func (plugin Github) ApiResources() map[string]map[string]core.ApiResourceHandler {
	return map[string]map[string]core.ApiResourceHandler{
		"test": {
			"GET": api.TestConnection,
		},
```

To all current plugins. @klesh says that the architecture may change anyway and this will solve the current problem.

### DIFF
--- a/config-ui/src/hooks/useConnectionManager.jsx
+++ b/config-ui/src/hooks/useConnectionManager.jsx
@@ -50,43 +50,44 @@ function useConnectionManager ({
   const [saveComplete, setSaveComplete] = useState(false)
   const [deleteComplete, setDeleteComplete] = useState(false)
 
-  const testConnection = (connection) => {
+  const testConnection = () => {
     setIsTesting(true)
     setShowError(false)
     ToastNotification.clear()
-    const testResponse = {
-      success: false,
-      connection: activeProvider,
-      errors: []
-    }
+    // TODO: run Save first
     const runTest = async () => {
-      let queryParams = ''
-      if (activeProvider.multiSource && connection.id){
-        queryParams += `?sourceId=${connection.id}`
+      let queryParams = ``
+      switch (activeProvider.id) {
+        case Providers.JENKINS:
+          queryParams = `?username=${username}&password=${password}&endpoint=${endpointUrl}`
+          break
+        case Providers.GITLAB:
+          queryParams = `?auth=${token}&endpoint=${endpointUrl}`
+          break
+        case Providers.GITHUB:
+          queryParams = `?auth=${token}&endpoint=${endpointUrl}`
+          break
+        case Providers.JIRA:
+          queryParams = `?auth=${token}&endpoint=${endpointUrl}`
+          break
       }
       let testUrl = `${DEVLAKE_ENDPOINT}/plugins/${activeProvider.id}/test`
       let getUrl = testUrl + queryParams
       console.log('INFO >>> GET URL for testing: ', getUrl);
       let res = await request.get(getUrl)
-      if (res.data && res.status === 200) {
-        testResponse.success = true
-      } else {
-        testResponse.errors.push(new Error("Test failed. Please check provider configuration."))
-      }
-    }
-    runTest()
-    setTimeout(() => {
-      if (testResponse.success) {
+      console.log('res.data', res.data);
+      if (res?.data?.Success && res.status === 200) {
         setIsTesting(false)
         setTestStatus(1)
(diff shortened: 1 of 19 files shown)


klesh:
I personally think it's ok to implement this via plugin api.

Because either way we have to have plugin api anyway for many other requirements, might as well put this optional feature into plugin api scope. Not all plugin need testconnection, not even for data source, like push api we were talking about lately, this kind of data sources, they dont connect to anywhere, instead receiving posted data from somewhere.

Another ongoing proposal is to let plugin implement specific interface if they supported a certain feature, and managed by framework, and framework would expose a common api endpoint for feature like `testConnection`, yes, this is doable, i just don't see the neccessity of it. The point of this proposal is being developer can implement plugin by this approach more easily because they just need to implement the the interface, and all plugins share the same api endpoint for same functionality. If we go down this way, we have bunch of interfaces to be defined, and most importantly, if we didn't conver every single possible scenario, we end up with a set of incomplete interfaces, even worst, it's hard to redefine those interfaces because many plugins already depended on some of those interfaces...And it will eventually become a complex code base like some good old JAVA projects. I quited c# because it was hard, not because of c# itself, but the complexity of OOP, it's hard to do it right, especially when there were just too many objects to be modeled, it's hard to find developer can get hold of the whole picture, not to mention it's impossible to have most developers share the same opinion on how should objects be modeled, and people end with arguing this should be modeled into 2 or 3 objects/interfaces.  It's dangerous for a system to become complex, think about windows , I pity the windows system developers, they have look back on each step, in order to make sure previous program would work properly, they have to becareful, each of every published API can not be updated, and they should stay the same behave forever.

However, this is a subject under discussion.

@liangjingyang @hezyin

joncodo:
The honest truth of this is that we don't have third party developers at this point. I do think that we need to think about that use case but it should not be our only concern.

If you want an interface defined, we can't enforce that interface since not all plugins need to provide testConnection.

If you have a concrete example of how to do this, please make a PR to show me what you mean. I have tried it both ways and I think this is the best way.

hezyin:
@klesh @joncodo In the long run, I'd vote for the interface approach, but this shouldn't block @joncodo from implementing this via plugin API. Either way, the majority of the work is shared.

And here's why I'd vote for the interface approach:

For developers who would want to add a data source for Dev Lake, they'll need to know how to write a plugin, specifically, what "interface" they need to implement? This "interface" is the contract between the framework and the plugins. In the simplest form, this "interface" only consists of one method called `Execute`. But as our functionalities grow, we may ask more from plugins and these new asks need to be included in the contract as well. This `TestConnection` is an example.

If we can agree on why we need this "interface" / contract, then let's talk about another issue. What if this contract only applies to a subset of plugins? For example, `TestConnection` should only apply to pull-based data source plugins and should never apply to subsequent plugins or push-based data source plugins.

```go
type Plugin interface {
}

type ConnectionTester interface{
    Connect(ctx context.Context, arg *Arg) (bool, error)
}

type PullBasedDataSourcePlugin interface{
    Plugin
    ConnectionTester
}
```

joncodo:
I really like the idea of classifying plugins. This means we can teat domain layer plugin different from tool layer. We can also detect if they pull data or not.

kevin-kline:
@joncodo pls review! Thx

kevin-kline:
UPDATE: after discussion with @joncodo @e2corporation , we decided to re-design this a little bit. I need to change the functionality from using stored params to query params.

Next steps:
- [ ] Extract query params from GET request
- [ ] Validate query params
- [ ] Use query params to make request

kevin-kline:
> UPDATE: after discussion with @joncodo @e2corporation , we decided to re-design this a little bit. I need to change the functionality from using stored params to query params.
> 
> Next steps:
> 
> * [x]  Extract query params from GET request
> * [x]  Validate query params
> * [x]  Use query params to make request
> * [x]  Send query params from frontend for each plugin
> * [ ]  Disable "Save connection" button until connection configuration has passed the test. (or some variation on this UX)

I'll leave the last one for another ticket. We should merge this.

Issue #800, opened 2021-11-16
macOS - DevLake `ERROR`: >>> Mysql failed to connect


e2corporation:
## Lake / Docker / Mac OS + MySQL
Users / Developers in a Mac OS Environment are unable to start **Lake** with `docker-compose up` due to a bad MySQL TCP Connection string. To resolve this issue for Mac OS and potentially other environments, please see the steps below.

**SOLUTION STEPS** 
- [ ] **Update MySQL Docker Configuration** in `docker-compose.yml` SET `container_name`
  - Define a **CONTAINER NAME** For the MySQL Instance eg. `container_name: lakeMYSQL`
- [ ] **Update Dev Lake's DB_URL Connection String** to use _Container Name_ and NOT _localhost_ 
  - `DB_URL=merico:merico@tcp(lakeMYSQL:3306)/lake?charset=utf8mb4&parseTime=True`

<img width="774" alt="Screen Shot 2021-11-16 at 10 47 11 AM" src="https://user-images.githubusercontent.com/1742233/142018169-f3724e34-8cb4-44e0-bfea-3b52ba1135a3.png">

## To Reproduce
1. Go to **DevLake** project folder (`lake`) running `main` branch
2. Run `docker-compose up` to start services
3. Notice the Go Init Model fails to connect `/app/models/init.go:31`
4. DevLake Issues **Panic Error** and exits with Code 2

## Expected behavior
DevLake should startup successfully and connect to the MySQL Service instance when running `docker-compose up` instead of throwing panic errors.

## Screenshots

**BEFORE** 
<img width="787" alt="Screen Shot 2021-11-16 at 10 20 49 AM" src="https://user-images.githubusercontent.com/1742233/142017359-2c057ea2-38f9-456b-b84b-38624ee7cc3c.png">

**AFTER PROPOSED SOLUTION**
<img width="787" alt="Screen Shot 2021-11-16 at 10 32 47 AM" src="https://user-images.githubusercontent.com/1742233/142017388-9e425ba5-1a64-4e64-90bb-3e9ff0c57f98.png">

## Additional context
Please see a similar issue below on StackOverflow for addtional context with regard to **Go** and **Docker**.
https://stackoverflow.com/questions/52504318/unable-to-connect-to-mysql-server-with-go-and-docker-dial-tcp-127-0-0-13306

e2corporation:
```
macOS Catalina 10.15.7 (19H2)
Docker version 20.10.8, build 3967b7d
[...]

PR #846, opened 2021-11-23
fix: #800

yihong0618:
fix: bug docker-compose can not up in some env like `mac` or `wsl` 

- [x] I have read through the [Contributing](../CONTRIBUTING.md) Documentation
- [x] This PR is using a `label` (bug, feature etc.)
- [ ] My code is has necessary documentation (if appropriate)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,7 @@ services:
     volumes:
       - mysql-storage:/var/lib/mysql
     restart: always
+    container_name: lakeMYSQL
     ports:
       - 127.0.0.1:3306:3306
     environment:
@@ -34,6 +35,8 @@ services:
     restart: always
     volumes:
       - ./.env:/app/.env
+    environment:
+      - DB_URL=merico:merico@tcp(lakeMYSQL:3306)/lake?charset=utf8mb4&parseTime=True
     depends_on:
       - grafana
   config-ui:


klesh:
@yihong0618 Thanks for your contribution.
Issue #800 was classified as a isolated incident, if you ran into same situation, please describe your situation on issue comment.
We have to understand why it's happening, and how common it's among our users, before we can approve PR regarding this issues,

yihong0618:
@klesh yes I had run into the same issue. And my env is `win10` `wsl`
Because the `docker network` problem mysql could not connect using docker-compose
e.g.
image

yihong0618:
And I search the isssues and try to rename the `DB_URL` the problem solved

klesh:
@yihong0618 I see, can u post the origin configuration which can reproduce the problem?

yihong0618:
I just follow the `doc`

```
git clone https://github.com/merico-dev/lake.git devlake
cd devlake
cp .env.example .env
docker-compose up # without -d
```

klesh:
I wonder why `mysql` was resolved to `127.0.0.1`? 
Can u do a `ping mysql` both on your host machine and `config-ui` container and post the output here?

yihong0618:
> I wonder why `mysql` was resolved to `127.0.0.1`? 
> 
> Can u do a `ping mysql` both on your host machine and `config-ui` container and post the output here?

will so it later.

https://stackoverflow.com/questions/52504318/unable-to-connect-to-mysql-server-with-go-and-docker-dial-tcp-127-0-0-13306

yihong0618:
> @yihong0618 If the issue you are describing is similar to #800, you should not have to add an `environment` block to docker-compose.yml
> 
> In your `.env` file the `DB_URL` is set there, the TCP connection string uses a hostname reference of `mysql`. So you just need to define the `container_name` property in your local docker-compose.yml that matches the name in the connection string, ie `container_name=mysql`. If you opted to use `lakeMYSQL` like suggested on ticket #800, you have to make the change in both places as well.
> 
> Also ensure you don't have another MySQL service running on your host machine on port 3306.

Agreed but IMO users who use docker-compose want to quick start the repo，and take a look. 
So I change only the compose file. About .env I think the may not only use for compose-yml

e2corporation:
@yihong0618 Thanks again for the PR proposal, our team will need to do more QA/testing with a **Windows** / `win10` environment + wsl to determine the best default configuration that works out of the box for all environments. For now we'll have to decline this PR as it doesn't fully address the underlying issue which is that the `DB_URL` connection string will vary if you are starting lake as a `User` vs `Developer`.

Also please note that our READMEs on the project mention that there is a different value for the MySQL hostname whether you are running lake as a _user_ vs _developer_. (`mysql:3306` vs `127.0.0.1:3306`)

For your `local` environment feel free to continue using your workaround mentioned on this ticket until our team has had a chance to review issues regarding windows environments in general.

@joncodo @klesh

PR #850, opened 2021-11-23
feat: only compile the plugins you want

joncodo:
Fixes: https://github.com/merico-dev/lake/issues/487

## Problem

On my computer, it takes me a long time to compile all plugins. I need a faster way to develop. Most times I only work in one plugin at a time and the re-comilation takes up too much of my day.

## Solution 

Everything will work as before here. All you have to do is run `make dev`

However, if you are only working in one plugin, you can now update this line in `scripts/compile-plugins.sh`

`PLUGIN_TO_COMPILE=all` for all plugins as before

`PLUGIN_TO_COMPILE=github` for only github

Screen Shot 2021-11-23 at 11 17 38 AM

### DIFF
--- a/scripts/compile-plugins.sh
+++ b/scripts/compile-plugins.sh
@@ -2,7 +2,9 @@
 # Change this if you only want to complile the plugin you are working on
 
 # If you want to use this, you neec to run `LOCAL=true make dev`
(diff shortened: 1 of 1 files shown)


joncodo:
See issue for context

PR #851, opened 2021-11-23
feat: dynamic JSON for triggers page

### LINKED ISSUE #774: Populate "Triggers" JSON dynamically
## Description
Many problems can arise from our current "Triggers" page. Perhaps the biggest problem is that users may trigger plugins that they have not yet activated.

We should pre-populate the "Triggers" page JSON based on which plugins have been activated by the user. We can do this by checking which connections exist.

No.

We should redesign the "Triggers" page completely, and hide all JSON from the User. 

## Additional context
This solution is a temporary minimum and a good start to addressing the long term Usability concerns.

kevin-kline:
### Description
This PR creates Utils that allow us to get example request body JSON ready to display to the user or send to the /task API based on an array of plugin names. For example, send "jira" and receive:

```
[[{
Plugin: "jira",
Options: {...}
}, {
Plugin: "jiradomain",
Options: {...}
}]]
```

This is a pretty flexible solution for customizing JSON for the user. It also implents:
- unit tests
- Util function for getting sources from the backend to tell us which plugins have existing connections
- Triggers Page updates sample JSON based on plugins with existing connections

closes #774 

### Current Behavior
Single hard-coded value for sample JSON on Triggers Page.

### New Behavior
Dynamic updates for sample JSON on Triggers Page based on which plugins have active source data from the backend.

### DIFF
--- a/config-ui/src/data/availablePlugins.js
+++ b/config-ui/src/data/availablePlugins.js
@@ -1,0 +1,3 @@
+const AVAILABLE_PLUGINS = ['gitlab', 'jira', 'jenkins', 'github']
+
+module.exports = AVAILABLE_PLUGINS
--- a/config-ui/src/data/collectionDefaults.js
+++ b/config-ui/src/data/collectionDefaults.js
@@ -1,0 +1,40 @@
+const collectionDefaults = {
+  github: {
+    options: {
+      repositoryName: 'lake',
+      owner: 'merico-dev'
+    },
+    domainPlugin: {
+      name: 'github-domain',
+      options: {}
+    }
+  },
+  gitlab: {
+    options: {
+      projectId: 8967944
+    },
+    domainPlugin: {
+      name: 'gitlab-domain',
+      options: {}
+    }
+  },
+  jira: {
+    options: {
+      boardId: 8,
+      sourceId: 1,
+    },
+    domainPlugin: {
+      name: 'jiradomain',
+      options: {}
+    }
+  },
+  jenkins: {
+    options: {},
+    domainPlugin: {
+      name: 'jenkinsdomain',
+      options: {}
+    }
+  },
+}
+
+module.exports = collectionDefaults
--- a/config-ui/src/utils/sourcesUtil.js
+++ b/config-ui/src/utils/sourcesUtil.js
@@ -1,0 +1,26 @@
+import { DEVLAKE_ENDPOINT } from './config.js'
+import request from './request'
+import PLUGINS from '../data/availablePlugins'
+
+const SourcesUtil = {
+  getPluginSources: async () => {
+    const pluginsToSet = []
+    const errors = []
+    for (const plugin of PLUGINS) {
+      try {
+        const res = await request.get(`${DEVLAKE_ENDPOINT}/plugins/${plugin}/sources`)
+        if (res?.data?.length > 0) {
+          pluginsToSet.push(plugin)
+        }
+      } catch (error) {
+        errors.push(error)
+      }
+    }
+    if (errors.length > 0) {
+      console.log('errors', errors)
+    }
+    return pluginsToSet
+  }
+}
+
+export default SourcesUtil
--- a/config-ui/src/utils/triggersUtil.js
+++ b/config-ui/src/utils/triggersUtil.js
@@ -1,0 +1,33 @@
+const COLLECTION_DEFAULTS = require('../data/collectionDefaults')
+const PLUGINS = require('../data/availablePlugins')
+
+const TriggersUtil = {
+  getCollectionJson: (plugins = []) => {
+    const arrayOfCollectors = []
+    const arrayOfDomainConverters = []
+
+    for (const plugin of plugins) {
+      if (PLUGINS.includes(plugin)) {
+        const collectorJson = module.exports.getCollectorJson(plugin)
+        const domainJson = module.exports.getDomainJson(plugin)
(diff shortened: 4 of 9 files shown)


config-ui/src/utils/sourcesUtil.js
```diff
@@ -0,0 +1,26 @@
+import { DEVLAKE_ENDPOINT } from './config.js'
+import request from './request'
+import PLUGINS from '../data/availablePlugins'
```
e2corporation, line 3: Aliased imports are preferred, 
import from `@/data/availablePlugins`, '@/utils/request` etc

config-ui/src/utils/triggersUtil.js
```diff
@@ -0,0 +1,33 @@
+const COLLECTION_DEFAULTS = require('../data/collectionDefaults')
```
e2corporation, line 1: Switch to aliased imports `@/data....`

e2corporation: Also we are using `imports` everywhere in config-ui as opposed to require.

config-ui/src/data/collectionDefaults.js
```diff
@@ -0,0 +1,40 @@
...
+  jenkins: {
+    options: {},
+    domainPlugin: {
+      name: 'jenkinsdomain',
+      options: {}
+    }
+  },
+}
+
+module.exports = collectionDefaults
```
e2corporation, line 40: Switch to `export default collectionDefaults`

config-ui/src/utils/triggersUtil.js
```diff
@@ -0,0 +1,33 @@
...
+  },
+  getDomainJson: (name) => {
+    return {
+      Plugin: COLLECTION_DEFAULTS[name]?.domainPlugin?.name,
+      Options: COLLECTION_DEFAULTS[name]?.domainPlugin?.options,
+    }
+  }
+}
+
+module.exports = TriggersUtil
```
e2corporation, line 33: Switch to `export default TriggersUtil`

config-ui/test/json.js
```diff
@@ -0,0 +1,28 @@
+var assert = require("assert");
```
e2corporation, line 1: Switch to `import` statements

config-ui/src/data/availablePlugins.js
```diff
@@ -0,0 +1,3 @@
+const AVAILABLE_PLUGINS = ['gitlab', 'jira', 'jenkins', 'github']
```
e2corporation, line 1: We have a top-level data file being referenced for supported integrations (plugins) already, see `@/data/Providers`, this file can be removed and the `Providers` constant referenced instead.

config-ui/src/utils/triggersUtil.js
kevin-kline: Hey Julien! 

I'm aware of require/import syntax. Import/export syntax is not compatible with Mocha for testing, and these utils need testing. If you'd like to propose an alternative testing solution, that's fine, otherwise I recommend we leave it as is since it does not affect performance. Import syntax is compatible with module.exports so I think it's a non-issue.

kevin-kline: I've been informed Mocha may be compatible with ```import``` syntax: https://stackoverflow.com/questions/57004631/mocha-tests-with-es6-style-imports

I will try to implement it.

config-ui/src/data/availablePlugins.js
kevin-kline: What about 'null'? This is not a provider.

image

config-ui/src/utils/triggersUtil.js
e2corporation: @kevin-kline Ahh I see -- can you also test that this has no impact to the build process if we do have to leave the requires in?

config-ui/src/data/availablePlugins.js
e2corporation: It's there as a placeholder provider, perhaps I was thinking about the Integrations.js constant instead. Reasoning is when we add a new provider we'll have to address this constant as well and add another key entry here in AVAILABLE_PLUGINS. I think leaving this file can work for now but will refactor later.

config-ui/src/utils/triggersUtil.js
kevin-kline: Attempt failed. I think it's a big rabbit hole for little gain. Anyone else wanna try? Again...it doesn't affect performance, merely cosmetic. Unless I'm mistaken.

kevin-kline: > 

How would I test this build process? Pls advise.

e2corporation: Simply do a `npm run build-production` that will run a build to ensure there are no warnings.

e2corporation (approved):
Approving PR given the `require` exception for Mocha. As long as build process is tested and no new errors are generated in the stack.

config-ui/src/utils/triggersUtil.js
kevin-kline: > Simply do a `npm run build-production` that will run a build to ensure there are no warnings.
> 
> 
> 
> 

I'll do that tomorrow to make sure.

kevin-kline: Done this. No errors.

PR #858, opened 2021-11-24
Ae plugin

joncodo:
Fixes https://github.com/merico-dev/lake/issues/448

From @hezyin 
- Maybe we only update the dev EQ and leave the rest to github or gitlab. This way no fields are shared by two plugins

UPDATE: Done and ready for review @klesh 

See the readmes in this PR for full details

Screen Shot 2021-11-29 at 10 47 36 AM


plugins/merico-analysis-engine/tasks/ae_api_client.go
```diff
@@ -0,0 +1,232 @@
...
+	if queryParams == nil {
+		queryParams = &url.Values{}
+	}
+	total, _, err := aeApiClient.getTotal(path, queryParams)
+	if err != nil {
+		return err
+	}
+
+	// not all api return x-total header, use step concurrency
+	if total == -1 {
```
klesh, line 134: AE is using trino as database, which is good at big chunck of data but bad at concurrency.
AFAIK, AE implemented pagination by caching query result, and has no locking mechanism at this point, so fetching data in parallel would likely make it slower.
We should remove the parallel fetching for Ae, and fetch with big amount of records for each page like 2000 at this point.

plugins/merico-analysis-engine/tasks/ae_commit_collector.go
```diff
@@ -0,0 +1,66 @@
...
+	HexSha      string `json:"hexsha"`
+	AnalysisId  string `json:"analysis_id"`
+	AuthorEmail string `json:"author_email"`
+	DevEq       int    `json:"dev_eq"`
+}
+
+func CollectCommits(projectId int, scheduler *utils.WorkerScheduler) error {
+	aeApiClient := CreateApiClient()
+	relativePath := fmt.Sprintf("projects/%v/commits", projectId)
+	return aeApiClient.FetchWithPaginationAnts(scheduler, relativePath, SetQueryParams(1, 100), 100,
```
klesh, line 27: No fetching in parallel at this point

joncodo:
@klesh This is done and ready to review

klesh (approved):

Issue #799, opened 2021-11-16
`config-ui` reporting 503 before devlake ready


klesh:
## Description
`devlake` takes longer to boot up, and `config-ui` get ready almost instantly. User would get 503 errors before `devlake` ready when trying to set system up on `config-ui`.
It would be a plus if we can solve this issue before release.

## Describe the solution you'd like
1. Tweak `docker-compose` to start `config-ui` only when `devlake` is ready(in sense of truly ready for requests), not sure if doable
2. Display a WATING DEVLAKE BOOT UP screen before actually showing `config-ui` screen, and keep probing the devlake port, for maximum like 1min or so. Display  a failure message when timeouted.

## Has the Feature been Requested Before?
No

hezyin:
@e2corporation Please prioritize designing the trigger and pipeline pages before this ticket as many other tickets depend on that two pages.

PR #873, opened 2021-11-26
fix: add docker wait on hosts feature

e2corporation:
### Config UI / Docker / Service Waiting

- [x] Add **Docker Compose Wait** Package (`docker-compose-wait`)
  - Add `WAIT_HOSTS` to docker compose configuration
- [x] Create **Offline Mode View** fallback on UI
  -  Automatically **poll** API Endpoint for `7s` up to `100` attempts.
- [x] Create **Offline Mode Hook** (`@hooks/useNetworkOfflineMode.jsx`)
  - Implemented by _Connection Manager Hook_. Routes to `/offline` mode if a `504` is returned from Fetching All connections.

> ⚠️ **"Config-UI" Docker Image Rebuild Required!** This PR updates the `Dockerfile` for `config-ui`. This image must be rebuilt so the changes are referenced by the root project's `docker-compose.yml` (ie. a newer image version for config-ui)

### Description
This PR adds a service wait feature to **Docker** Compose so that service containers that depend on one another are allowed to complete start before the config-ui container is started. The user is also routed to an "**Offline Mode**" view if a `504` (Gateway Timeout) is returned from fetching sources for a Data Provider.

#799 

### Current Behavior
The `depends_on` flag in the docker configuration is not sufficient and was not designed to actually wait on services to startup before proceeding with container startup. This means that if Config-UI is accessed too soon by the web browser error notices will be displayed stating that the Dev Lake API is not available.

### New Behavior
Config-UI should delay start and **WAIT** on dependent container services, primarily `devlake:8080` before proceeding to launch. In addition a fallback/offline view mode (alert or notice) will be added to the application if the API is not reachable. 

### Screenshots
Screen Shot 2021-11-30 at 10 50 48 AM
Screen Shot 2021-11-30 at 10 51 30 AM


e2corporation:
@joncodo @klesh Ready for review & merge (Config-UI Docker Image rebuild required). Thanks!

e2corporation:
To **test** the `Dockerfile` in `config-ui` without rebuilding the _mericodev_ image, apply this patch to `docker-compose.yml` to build with local context:
Screen Shot 2021-11-30 at 12 31 00 PM

Then issue `--build` flag when running docker.

klesh (approved):

PR #922, opened 2021-12-02
fix: remove `wait` so offline page can show its charm

klesh:
# Summary

  `wait` will block nginx, so user wont be able to open `config-ui`,
  thus, user may never see the `/offline` page at all. this fixes the
  problem by removing the `wait` and make nginx be able to start even
  `devlake` is not ready.

# Step to verify
I have already push the latest build to dockhub. so you can verify it by:

1. chage `config-ui` image tag to latest and remove `depends_on` devlake part for `docker-compose.yml`,
2. stop all by `docker-compose down` and launch `config-ui` only by `docker-compose up -d config-ui`
3. open `config-ui` on browser, config-ui should pop up, and click into plugin, wait a few seconds `offline` page should be pop up
4. better to show `offline` page even if user didn't do anything, we can do this on next iter if @e2corporation think this is a good idea.

### Key Points

### DIFF
--- a/config-ui/nginx.conf
+++ b/config-ui/nginx.conf
@@ -1,6 +1,3 @@
-upstream devlake {
-    server ${DEVLAKE_ENDPOINT} fail_timeout=5s max_fails=5;
-}
 server {
   listen 80;
   server_name localhost;
@@ -11,8 +8,13 @@ server {
   }
 
   location /api/ {
+    resolver 127.0.0.11 valid=300s;
+    resolver_timeout 1s;
+    set $target "${DEVLAKE_ENDPOINT}";
     rewrite /api/(.*) /$1  break;
-    proxy_pass http://devlake;
+    proxy_send_timeout 2s;
+    proxy_read_timeout 2s;
+    proxy_pass http://$target;
   }
 
   location /grafana {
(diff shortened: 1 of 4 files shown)


config-ui/nginx.conf
```diff
@@ -11,8 +8,13 @@ server {
   }
 
   location /api/ {
+    resolver 127.0.0.11 valid=300s;
```
e2corporation, line 11: Is this address intentionally ending with `11` as the last IP block? `127.0.0.11`

config-ui/src/hooks/useNetworkOfflineMode.jsx
```diff
@@ -12,10 +12,11 @@ function useNetworkOfflineMode (offlineStatus = 504, offlineRoute = '/offline')
   }, [])
 
   useEffect(() => {
-    if (status && response && status === offlineStatus) {
+    console.log('status', status)
```
e2corporation, line 15: Other Effects are logging the status related to offlineMode, this can be removed.

e2corporation (approved):
@klesh Approved w/ some minor feedback.

config-ui/nginx.conf
klesh: intentionally, this is the dns server address of  docker network

PR #923, opened 2021-12-02
feat: remove old binaries when running make dev

joncodo:
Fixes https://github.com/merico-dev/lake/issues/883

This simply allows users to run `make dev` and clear out old binaries. 

This solves the problem where a user built binaries on a plugin before such as `jiradomain` and now that plugin no longer exists.

This is safe to always run


mindlesscloud (approved):
LGTM

PR #944, opened 2021-12-08
Push API

### LINKED ISSUE #428: Add Domain Layer Push api (Create only)
## Use Case

1. Push api makes it more flexible to write plugins
2. It provides a better user experience for async data outside our system. IE: Real time data

## Why do we want this?

1. There are companies that use their own software that we can't make a plugin for. If they want that data, they have to make it and its only for them.
2. Not every company has golang devs. They need to interface with a REST api for our domain layer. 
3. The api we expose does not handle all cases. 
4. Push apis allow 3rd party services to push data to us when the time is right.

## What should it do?

1. Exposes CRUD access to all models in the domain layer.
2. Let sql constraints validate the input. The api returns actual sql debugging information
3. This should be a service separate from lake.

## How secure is it?

1. Prevent some queries from breaking data?
2. Prevent security threats?

## Questions

1. Why do we maintain a plugin architecture AND a REST api? ANSWER: We need both
2. Do we want to maintain both? ANSWER: Yes

## Notes

- This is simply a REST api and not a message queue for real time events
- This should be written in golang

kevin-kline:
### Key Points

### Description
This PR offers an optional push API service for Users to insert data via a generic endpoint. 

## Refactor Note (*** THIS IS NO LONGER NEEDED, AND HAS BEEN REVERTED ***)

Please notice there is a small refactor to LoadConfigFile() which adds the input parameter "envPath" for the env file. That is why we have LoadConfigFile("") in many places where it is not needed to specify. But it was needed to access the .env file from a sub-directory.

### TODO

- [x] Add SQL error responses
- [x] Remove Dockerfile
- ** NO LONGER NEEDED ** Add to Docker Compose YAML as separate service
- [x] Add sanitation to SQL string

Closes #428


push-api/main.go
```diff
@@ -0,0 +1,47 @@
...
+	POST /api/:tableName
+	[
+		{
+			"id": 1,
+			"sha": "osidjfoawehfwh08"
+		}
+	]
+*/
+
+func main() {
```
klesh, line 21: Why we have separated api server for this? any particular reason it should not be a part of our existing api server?

kevin-kline: The reason to keep it separate is to offer it as an optional service to Users who want it. Some Users may not want it, or choose to activate it only sometimes.

Am I right? @joncodo

joncodo: Yes. We want it disabled in some cases

klesh: I see, but it doesn't necessary be a standalone server, it will be impossible or complicated to allow `config-ui` be able to toggle this functions.
How about we put it under a certain url path, and add a middleware whether let through or block the requests based on some `config.V` settings?
Here is an article I found regarding `gin middleware` https://sosedoff.com/2014/12/21/gin-middleware.html, it works pretty much lik `expressjs` middleware.

joncodo: Ok. Sounds good to me

kevin-kline:
New ToDo:

- [ ] Integrate with main api (instead of separate service)

PR #947, opened 2021-12-09
feat: disable auth in grafana

joncodo:
No one has asked for this but I stumbled upon this solution while I was working on something else.

Right now, we always ask users to log into grafana. However since they are hosting it, we don't really care. If they want login, they can enable it.

This allows users to go to grafana directly without login.

I want this in the code but its up to @Startrekzky and @hezyin


joncodo:
We need this for team code.

PR #948, opened 2021-12-09
feat: fix local mode for view dashboards button

joncodo:
This makes a local mode in webpack. Then we can use variables from webpack in the JS.

Fixes https://github.com/merico-dev/lake/issues/907

### DIFF
--- a/config-ui/package.json
+++ b/config-ui/package.json
@@ -16,6 +16,7 @@
     "axios": "^0.21.4",
     "babel-plugin-module-resolver": "^4.1.0",
     "dotenv": "^10.0.0",
+    "dotenv-webpack": "^7.0.3",
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "react-router-dom": "^5.3.0",
--- a/config-ui/src/utils/config.js
+++ b/config-ui/src/utils/config.js
@@ -1,2 +1,4 @@
 export const DEVLAKE_ENDPOINT = '/api'
-export const GRAFANA_ENDPOINT = '/grafana/d/0Rjxknc7z/demo-homepage?orgId=1'
+export const GRAFANA_BASE_URL = '/grafana'
+export const LOCAL_BASE_URL = 'http://localhost:3002'
+export const GRAFANA_ENDPOINT = '/d/0Rjxknc7z/demo-homepage?orgId=1'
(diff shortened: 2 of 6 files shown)


klesh:
I would prefer make the `dev server` act like `nginx` rather than `config-ui` using different url based on the context.
But this could work,  I'm ok with it,  it's up to you now @e2corporation

config-ui/package.json
```diff
@@ -16,6 +16,7 @@
     "axios": "^0.21.4",
     "babel-plugin-module-resolver": "^4.1.0",
     "dotenv": "^10.0.0",
+    "dotenv-webpack": "^7.0.3",
```
e2corporation, line 19: It would be great to add the updated `package-lock.json` file changes, but you may be better off deleting your local package-lock.json, removing node_modules, and re-running npm install, then adding that rebuilt `package-lock.json` to this PR.

kevin-kline (approved):
Tested and works locally.

config-ui/src/utils/config.js
```diff
@@ -1,2 +1,4 @@
 export const DEVLAKE_ENDPOINT = '/api'
-export const GRAFANA_ENDPOINT = '/grafana/d/0Rjxknc7z/demo-homepage?orgId=1'
+export const GRAFANA_BASE_URL = '/grafana'
+export const LOCAL_BASE_URL = 'http://localhost:3002'
+export const GRAFANA_ENDPOINT = '/d/0Rjxknc7z/demo-homepage?orgId=1'
```
e2corporation, line 4: Is it possible to do the ternary check for `process.env.LOCAL` here? That way we only need to export the `GRAFANA_ENDPOINT` and we can keep the same existing import signature elsewhere and remove the process env checks from the component files.

e2corporation (approved):
Approved with some minor feedback @joncodo

config-ui/package.json
kevin-kline: Done

e2corporation:
> I would prefer make the `dev server` act like `nginx` rather than `config-ui` using different url based on the context.
> But this could work, I'm ok with it, it's up to you now @e2corporation

I'm good with whatever solution @joncodo wants to apply, approved w/ some minor feedback.

joncodo:
> > I would prefer make the `dev server` act like `nginx` rather than `config-ui` using different url based on the context.
> > But this could work, I'm ok with it, it's up to you now @e2corporation
> 
> I'm good with whatever solution @joncodo wants to apply, approved w/ some minor feedback.

Fixed

Issue #575, opened 2021-10-28
Store all fields from jira, and extract customfield on enrichment stage


klesh:
## Description
Now we extract customfield on collection stage, which causing a full re-collection if didn't set the customfield name correctly, like setting `JIRA_ISSUE_EPIC_KEY_FIELD` to a wrong value.

## Describe the solution you'd like
Store all fields into into a json field, and extract data on enrichment stage

PR #955, opened 2021-12-10
feat: collect all jira issue field in db

likyh:
#575

# Summary

collect all jira issue fields to db

### Key Points

### Description
create a type `JsonColumn` and add it in `JiraIssue` to collect all jira issue fields to db

[Please mention the issues here.](https://github.com/merico-dev/lake/issues/575)

### Screenshots
image

### DIFF
--- a/plugins/jira/tasks/jira_issue_collector.go
+++ b/plugins/jira/tasks/jira_issue_collector.go
@@ -215,6 +215,7 @@ func convertIssue(source *models.JiraSource, jiraApiIssue *JiraApiIssue) (jiraIs
 		workload, _ = jiraApiIssue.Fields[source.StoryPointField].(float64)
 	}
 	jiraIssue = &models.JiraIssue{
+		AllFields:          jiraApiIssue.Fields,
 		SourceId:           source.ID,
 		IssueId:            id,
 		ProjectId:          projectId,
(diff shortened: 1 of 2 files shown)


models/common/base.go
```diff
@@ -23,3 +26,36 @@ var (
 func IsDuplicateError(err error) bool {
 	return err != nil && DUPLICATE_REGEX.MatchString(err.Error())
 }
+
+// JsonColumn json column in db 数据库json字段
+type JsonColumn struct {
```
klesh, line 31: Can `V` be of type `map[string]interface{}`? Than we don't need to `type assertion` before we can access the `map`.
And please update the comment to English for the sake of other developers.

plugins/jira/tasks/jira_issue_collector.go
```diff
@@ -215,6 +216,7 @@ func convertIssue(source *models.JiraSource, jiraApiIssue *JiraApiIssue) (jiraIs
 		workload, _ = jiraApiIssue.Fields[source.StoryPointField].(float64)
 	}
 	jiraIssue = &models.JiraIssue{
+		AllFields:          common.JsonColumn{fields},
```
klesh, line 219: please fix the linting issue.

PR #956, opened 2021-12-10
feat: ae supports canceling

klesh:
# Summary

Make ae plugin supports canceling, both collector and converter.

### Key Points

### Description
The idea is simple:
1. make `api client` supports cancel, so before/after each request, task can be canceled right away as long as `ctx.Context` got passed into `api client` by `SetContext` method correctly.
2. add `cancel checking` in the begining of each loop, so conversion task can be canceled.

image

### DIFF
--- a/plugins/ae/ae.go
+++ b/plugins/ae/ae.go
@@ -40,19 +40,19 @@ func (plugin AE) Execute(options map[string]interface{}, progress chan<- float32
 	}
 
 	progress <- 0.1
-	if err := tasks.CollectProject(projectIdInt); err != nil {
+	if err := tasks.CollectProject(projectIdInt, ctx); err != nil {
 		return fmt.Errorf("could not collect project: %v", err)
 	}
 
 	progress <- 0.25
 
-	if err := tasks.CollectCommits(projectIdInt); err != nil {
+	if err := tasks.CollectCommits(projectIdInt, ctx); err != nil {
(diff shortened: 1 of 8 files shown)


joncodo:
From @klesh 

Hi, @Jonathan ODonnell  @Kevin Kline , I won't have much time to coding on  canceling feature recently, would you guys help me on it, I have put up a sample PR https://github.com/merico-dev/lake/pull/956
The idea is simple:
the work_scheduler already support canceling, but not every task utilizing work_scheduler
make ae api client supports canceling (maybe we could move this feature to api client once for all), so before/after each request, task can be canceled right away, even for collectors that didn't utilize the work_scheduer but just api client.
add cancel checking in the begining of each loop for converter, so conversion task can be canceled.
What needs to be done:
make sure all plugins implement the idea above
make sure all returned canceling error is the exact instance of core.TaskCanceled, this is for TASK_CANCELED status in database on next step, using same instance is the most reliable way for error checking.
channel and context.Conext is very important for our system in the long term, make sure you understand how they work.
@Julien Chinapen maybe we could add a canceling button for  Triggers page before `pipeline run` is ready.

joncodo:
@klesh we can work on this for sure but we are in the middle of the DB schema changes and migrations. @Startrekzky Please assign this ticket a priority label

klesh:
you can work on db schema migration first.

mindlesscloud (approved):

PR #967, opened 2021-12-14
Jira issue commits relationship

### LINKED ISSUE #490: map issue and commit entities in Jira tool layer
## Description
Map issue_id (from issue management domain) and commit_id (from issue management domain) to calculate metrics such as `bug count per commit`.

This is a requirement proposed by our EE team to observe the delivery quality of their work.

Users have 2 logical ways to map the 2 entities: issue and commit.

1. associate commit(s) to an issue, no matter it's the id, url, or value that contains the distinct identifier of a commit. Eg. use a field in Jira issue to store commit url.
2. associate issue(s) to a commit, no matter it's the key, url, or value that contains the distinct identifier of an issue. Eg. add an issue key like "EE-2581" in a commit message in Gitlab.

The EE team uses [Gitlab for Jira](https://docs.gitlab.com/ee/integration/jira/connect-app.html), a Jira plugin that enables commits, branches, and MR in Gitlab projects to be synced to Jira.

When developers submit a commit with a commit message that includes an issue key like "EE-5315", the Gitlab url of this commit will automatically be stored in the `web link` field of a Jira issue. E.g. merico.gitlab.com/xxx/~/commit-abcdefg.

We can parse the value of  `web link` to get the commit id, `abcdefg`, by parsing the web link that matches the commit url rule.

image

However, this approach is only used by our EE team. Other users may use various personalized ways to log the commit identifier in a field of an issue. Therefore, we should 

- [ ] allow users to specify the field to be parsed, no matter in which domain the field is
- [ ] hardcode the parsing rule

## ER Relation between Issue and Commit
Issue:commit  = n:n
An issue can map to multiple commits, a commit can map to multiple issues.

## Questions
1. It seems different users may use different methods to link Jira issues to Gitlab commits, this may lead to a result that Commit url may appeared in Jira issue's comment section or remote link or other fields.

klesh:
# Summary

1. jira issue remotelinks collection
2. create jira issue / git commit sha relationship based on remotelink url pattern

### Key Points

### Description

Lately, we received some usabiltiy issues:
1. User have to recollect all data if they set a wrong value for customfield, #575 
2. Conversion takes so long if table was too big

By developing this feature, I also try to solve these issues:
1. Collection stage store only primary key and `RawJson`, and extract needed field on Enrichment stage. So, for customfield, users don't have to recollection if they set it wrong. This is similar to previous js POC implementation.
2. Adopt `Pre Allocated Array` and `Batch Insert` for Conversion tasks which introduced by @mindlesscloud , this technique reduced conversion time dramatically.

I would like everybody @mindlesscloud  @hezyin @joncodo @kevin-kline to take a close look at this PR, and post your thought regarding the approach, and we may gradually refactor all our tasks if we all concured this is a better way to go.

Closes #490

### DIFF
--- a/plugins/jira/tasks/jira_remotelink_enricher.go
+++ b/plugins/jira/tasks/jira_remotelink_enricher.go
@@ -1,0 +1,122 @@
+package tasks
+
+import (
+	"encoding/json"
+	"regexp"
+
+	lakeModels "github.com/merico-dev/lake/models"
+	"github.com/merico-dev/lake/plugins/jira/models"
+	"gorm.io/gorm/clause"
+)
+
+func EnrichRemotelinks(source *models.JiraSource, boardId uint64) (err error) {
+	// prepare for issue_commits enrichment
+	var commitShaRegex *regexp.Regexp
+	if source.RemotelinkCommitShaPattern != "" {
+		commitShaRegex = regexp.MustCompile(source.RemotelinkCommitShaPattern)
+	}
+	// clean up issue_commits relationship for board
+	lakeModels.Db.Exec(`
+	DELETE ic
+	FROM jira_issue_commits ic
+	LEFT JOIN jira_board_issues bi ON (bi.source_id = ic.source_id AND bi.issue_id = ic.issue_id)
+	WHERE ic.source_id = ? AND bi.board_id = ?
+	`, source.ID, boardId)
+
+	var remotelink *models.JiraRemotelink
+	var issueCommit *models.JiraIssueCommit
+	// select all remotelinks belongs to the board, cursor is important for low memory footprint
+	cursor, err := lakeModels.Db.Model(&models.JiraRemotelink{}).
+		Select("jira_remotelinks.*").
+		Joins("left join jira_board_issues on jira_board_issues.issue_id = jira_remotelinks.issue_id").
+		Where("jira_board_issues.board_id = ? AND jira_board_issues.source_id = ?", boardId, source.ID).
+		Rows()
+	if err != nil {
+		return err
+	}
+	defer cursor.Close()
+
+	// save in batch, this could improvement performance dramatically
+	size := 1000
+	i := 0
+	j := 0
+	batch := make([]models.JiraRemotelink, size)
+	icBatch := make([]models.JiraIssueCommit, size)
+	saveBatch := func() error {
+		err := lakeModels.Db.Clauses(clause.OnConflict{
+			UpdateAll: true,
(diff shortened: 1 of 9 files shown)


joncodo:
I 100% agree. Collection should collect all the data it can. Never ask users to re-collect. Instead, just re-enrich. If we don't do this and we want a new property of the original JSON, we will have to re-collect it all. That is our bottle neck.

plugins/jira/tasks/jira_remotelink_enricher.go
```diff
@@ -0,0 +1,122 @@
...
+		Select("jira_remotelinks.*").
+		Joins("left join jira_board_issues on jira_board_issues.issue_id = jira_remotelinks.issue_id").
+		Where("jira_board_issues.board_id = ? AND jira_board_issues.source_id = ?", boardId, source.ID).
+		Rows()
+	if err != nil {
+		return err
+	}
+	defer cursor.Close()
+
+	// save in batch, this could improvement performance dramatically
```
joncodo, line 39: nice

joncodo:
> I 100% agree. Collection should collect all the data it can. Never ask users to re-collect. Instead, just re-enrich. If we don't do this and we want a new property of the original JSON, we will have to re-collect it all. That is our bottle neck.

This is one of the reasons I wanted to use Mongo for collection and mysql for relational

kevin-kline:
> 

This is my thought too. Why use MySQL to store JSON when we could use MongoDB? If we're going this route, seems like we should go back to keeping them separate.

kevin-kline (approved):
It LGTM. See comments for more info. I think gathering all the raw data is good; same as how we had with JS version. I'm just not sure if we need MySQL to store JSON blobs instead of Mongo (with separate collection layer from enriched layer), but if you still believe it's better to keep 1 DB, I'm still on board with it.

klesh:
@joncodo @kevin-kline MongoDB is hard to maintenance when data set grow too big, and we didn't use any particular feature from MongoDB, so most of us agreed that we better remove it to keep things simple, lower the bar of trying out our system, this is the same reason why we changed the postgres to mysql, most of us actually prefer postgres rather than mysql, but we all admitted mysql is more widely spreaded than postgres, it would be easier for most of our users to get started with our system.

@hezyin @mindlesscloud post your thought please.
@joncodo please help making sure we all align on this, and then we can merge this one and create issues for other plugins.

joncodo:
LGTM. Please merge

joncodo (approved):

Issue #971, opened 2021-12-15
Add dev and user profiles to docker-compose file


ggcannard:
## Description
Add separate compose profiles to the project's docker-compose file that separate user and development use-cases. The profiles feature of docker-compose allows users to run commands against a subset of services defined in the compose file. The end result will be that users and developers will use one of the following commands to run the project:

* `docker compose --profile user up -d`
* `docker compose --profile dev up -d`

## Proposed Implementation
Define one additional service in the compose file:
* grafana-dev: Rather than run Grafana from a pre-built image where the dashboards are copied into the container, run one where these are mounted as local volumes. This will save a lot of time, as updates to the Grafana charts will appear instantly in the browser rather than having to rebuild the image.

Include the following services in each profile:
* default (always started):
    * mysql
* user
    * grafana
    * devlake
* dev
    * grafana-dev

## Scope of Change
This change will update the docker-compose file at the root of the project, and require updating the readme to document this new workflow.

ggcannard:
Created a PR for this. The actual implementation differs a bit from what was proposed, see description in the PR.

PR #973, opened 2021-12-15
feat: adds compose profiles

ggcannard:
Implements #971 

The actual implementation differs slightly from what was proposed. In order to keep the user setup as easy as possible, the `COMPOSE_PROFILES` env variable has been added to `.env` and is set to `user`. This makes it that no change to the user setup is necessary. The readme instructs developers to switch the profile by changing this variable to `dev`.


klesh (commented):
LGTM

Issue #806, opened 2021-11-17
Upgrade for Existing users/DB migrations for existing data


Startrekzky:
## Description
Now we dont actually support upgrade devlake to newer version at this point, things like data schema change would cause devlake not working properly, and grafana dashboards may not updated after first initialization(to be verified), we need to support upgrade to newer version for deployed instance. 

1. Establish a mechanism to ensure user can migrate database for deployed instance during upgrade operation
2. Make sure user can get latest dashboards on deployed instance when user chose to upgrade

## Describe the solution you'd like

Solutions are to be consider, we need to investigate the situation futher more. like:
1. How is `gorm`s `AutoMigrate` behave, what is the limitation
2. Does `grafana` copys prebuilt dashboards on `initialization` and never again? 

## Has the Feature been Requested Before?
No

## Additional context
@runjinz was unable to upgrade his deployed instance while keeping all his data. had to delete tables and re-sync everything.

joncodo:
?

klesh:
@joncodo  issue detail is filled.

joncodo:
You have to use Migration from Gorm. https://gorm.io/docs/migration.html

klesh:
We should try this feature next time we have to do some breaking migration, set up an example of how to do so.

joncodo:
@mindlesscloud I am going to take this ticket. Hopefully that is ok with you

joncodo:
# PLAN OF ATTACK


We need tests to make sure we can insert every model type as of now
We need to migrate from scratch
All migrations should be done with SQL
All test should still pass after refactor
Tests will help to document the data model to our users

joncodo:
To investigate: https://sqitch.org/

joncodo:
Currently running into this problem: https://github.com/golang-migrate/migrate/issues/282

joncodo:
PR ready

joncodo:
## Questions to solve

1. How do we force the tests to use the test DB? - Use a flag for TEST in the init method
2. How do users auto run the migration scripts on each run of the service?
3. Have we tested this new DB from a new user perspective from docker-compose?
4. How do we run these tests in CI? - Prevent the models tests in CI. Only run them locally for now
5. How do we remove the auto migrate lines and make the migrations run against lake for CI and docker-compose and local dev?
6. How do plugins run migrations?

joncodo:
Closing in favour of https://github.com/merico-dev/lake/issues/979

PR #974, opened 2021-12-15
feat: insert tests for some domain level DB models - WIP

kevin-kline:
### Description
This is a WIP PR.

The purpose is to create tests on all our DB models prior to our migrations refactor in ticket #806 
This is step 1.
We will run the tests again AFTER the refactor to make sure all models behave as expected.

Relates to: https://github.com/merico-dev/lake/issues/972

### Other Information

- lake_test DB must be created manually to run tests.
- tests are currently EXPECTED to fail
- The plan is to create test DB programmatically.
- Only some of the models are tested here. This needs more work, I'm just putting it up for transparency, feedback, and/or someone else to continue with this.


joncodo:
Taking this over for today. We can pair again on it when you get back tomorrow my dude @kevin-kline

PR #981, opened 2021-12-17
Kk db tests

kevin-kline:
### Description
This PR adds more model tests for domain layer entities.

related to (read for more details):
- https://github.com/merico-dev/lake/issues/972
- https://github.com/merico-dev/lake/issues/979

### Other Information
Experiencing some unexpected behavior with migrations. Some result from "dirty database" problems:
https://github.com/golang-migrate/migrate/blob/master/FAQ.md#what-does-dirty-database-mean

image


joncodo:
The way to fix the above error is documented in the Migrations.md file

```
## How to reset your DB. (https://github.com/golang-migrate/migrate/issues/282#issuecomment-530743258)

NOTE: Use this when you get an error like this "error: Dirty database version 16. Fix and force version."

1. Delete your DB
2. `migrate -path db/migration -database "mysql://root:admin@tcp(localhost:3306)/lake" -verbose force 1`
3. `migrate -path db/migration -database "mysql://root:admin@tcp(localhost:3306)/lake" -verbose down`
4. `migrate -path db/migration -database "mysql://root:admin@tcp(localhost:3306)/lake" -verbose up`
```

Issue #993, opened 2021-12-21
Document how to step into plugins with debugger


ggcannard:
## Documentation Scope
Help developers to setup debugging for plugins code. Document will describe passing appropriate options to the build command, and provide example configurations for VS Code. It will also link to open bug reports relevant to this topic.

PR #995, opened 2021-12-21
docs: Adds help document for debugging

ggcannard:
Adds debugging help documentation described in #993

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ hello:
 build-plugin:
 	@sh scripts/compile-plugins.sh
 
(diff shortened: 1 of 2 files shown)


klesh:
I think this should be placed to wiki rathar than code base because it is not specifically related to our code base, this technique or knowlegde may change over time depends on golang/vscode evolve.

joncodo:
Moved to wiki

joncodo:
https://github.com/merico-dev/lake/wiki/How-to-enable-a-debugger-in-Go

PR #1003, opened 2021-12-22
"/env" Set and Get refactor

### LINKED ISSUE #1001: Refactor - Config variables
## Description
Currently, when we send data to the "/env" endpoint, we are unmarshalling JSON data into a struct with fixed properties and setting them each manually. This means we have to maintain a list of all ENV variables, which makes it difficult to change.

I'd like to refactor the Set method to generically accept any keys and set them in config by looping over the keys/values.

No.

We could keep it the same.

## Additional context 

image

## Tags

@joncodo

kevin-kline:
### Description
This PR contains a small refactor to make handling ENV variables through the /env endpoint easier. The solution provided here allows us to read/write without adhering to a specifically determined set of variables. This means we have less maintenance overhead and changes are easier to make. 

Closes #1001

### Current Behavior
Env vars are read/written using hardcoded lists we need to maintain.

### New Behavior
Env vars are read/written dynamically; no list is needed.

### Other Information
- @e2corporation My search through the code suggested we are not using the "/env" endpoint directly from any active pages on the config-ui (only "offline" index page and "configure" index page, both deprecated). Please confirm that this endpoint is not used and no frontend changes are needed. 
- Jenkins sources was refactored a little due to a change to the GetConfigJson() util method.
- "TEST_VAR" added to .env.example to make Config Read/Write unit test pass.
- Test updated to pass as well

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -78,3 +78,8 @@ GRAFANA_ENDPOINT=http://localhost:3002
 AE_APP_ID=
 AE_SECRET_KEY=
 AE_ENDPOINT=
+
+#########################
+# Testing
+#########################
+TEST_VAR=false
--- a/api/env/env.go
+++ b/api/env/env.go
@@ -4,7 +4,6 @@ import (
 	"net/http"
 
 	"github.com/gin-gonic/gin"
-	"github.com/gin-gonic/gin/binding"
 	"github.com/merico-dev/lake/config"
 	"github.com/merico-dev/lake/logger"
 	"github.com/spf13/viper"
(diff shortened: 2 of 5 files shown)


e2corporation:
@kevin-kline Correct, I don't think we are sending `POST` requests to `/env` at the moment, however the **Offline Manager** does need that endpoint and will poll `/env` with a `GET` request in order to determine the Lake API status.

klesh:
Right now we have nginx loading this .env file directly, allowing user to set arbitrary environment variables is a potential security risk, we have to think it through.

klesh:
We need the strong typed configuration in some way, make it a string dictionary is too general and hard to maintain (settings are scattered into multiple places).

In order to avoid duplicated effort maintaining `api/env/env.go`, I think a better approach is:
1. decode json request to a `map[string]interface{}`
2. read keys from `config.V.AllKeys()`
3. loop through all keys and compare to request json
4. update setting if key existed in `AllKeys`
5. write config file

and the whole configuration must be redesigned because the framework right now awares specific settings of each and every plugins, which is not ideal.

joncodo:
@klesh are you suggesting that we get rid of the .env file in favour of yaml or json?

PR #1013, opened 2021-12-29
style: tidy `go.mod` correctly

Juneezee:
# Summary

### Key Points

### Description
There should be two `require` directives only: one for direct dependencies, and another one for indirect dependencies. To maintain consistency, it is best to tidy them nicely.

N/A

### Current Behavior
The current `go.mod` file has 4 `require` directives. At Go 1.17 or higher, it is only necessary to have two `require` directives to enable the [module graph pruning](https://go.dev/ref/mod#graph-pruning).

### New Behavior
`go.mod` file only has two `require` directives, with direct and indirect dependencies separated.

### Screenshots
N/A

### Other Information
N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,42 +3,34 @@ module github.com/merico-dev/lake
 go 1.17
 
 require (
+	github.com/bndr/gojenkins v1.1.0
+	github.com/faabiosr/cachego v0.15.0
+	github.com/fastwego/feishu v1.0.0-beta.4
+	github.com/gin-contrib/cors v1.3.1
+	github.com/gin-gonic/gin v1.7.4
+	github.com/go-playground/validator/v10 v10.9.0
+	github.com/golang-migrate/migrate/v4 v4.15.1
+	github.com/magiconair/properties v1.8.5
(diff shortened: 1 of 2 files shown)


mindlesscloud (approved):

Startrekzky:
Thank you for your contribution to DevLake. We really appreciate it! To better serve our users, do you think you could share your DevLake use case with us?

Juneezee:
@Startrekzky Sure! We have a self-hosted GitLab instance, and we do not want to purchase their premium plan at the moment because we are a still fairly small team. At the same time, we would like use the productivity analytics feature (which is only available on GitLab subscription plan) in our weekly review meeting.

Startrekzky:
> @Startrekzky Sure! We have a self-hosted GitLab instance, and we do not want to purchase their premium plan at the moment because we are a still fairly small team. At the same time, we would like use the productivity analytics feature (which is only available on GitLab subscription plan) in our weekly review meeting.

Cool. Productivity analytics is exactly what we're working on. What metrics are you looking at?

Issue #1044, opened 2022-01-07
Paging in UI wording is off when there are no records to display.


joncodo:
URL: http://localhost:4000/pipelines

![Screen Shot 2022-01-07 at 9 37 40 AM](https://user-images.githubusercontent.com/3011407/148551907-e184cb16-22a6-4adb-a29d-c3a4b5443c07.png)

e2corporation:
@joncodo Will get this hot fixed.

e2corporation:
@joncodo PR ready. https://github.com/merico-dev/lake/pull/1047

PR #1047, opened 2022-01-07
`Hotfix` manage pipelines : fix pagination displayed count & range

e2corporation:
### Config UI / Pipelines / Manage Pipelines

- [x] `Fixed` Display count should be 0 when no data
- [x] `Fixed` Incorrect Max Page Detection (Prefer `ceil` not `floor`)
- [x] `Fixed` Display count when filtered count less-than-or-equal to `perPage` 
- [x] `Fixed` Handle Last Page not accessible scenario
- [x] `Fixed` Handle Overpaging
- [x] `Fixed` Add special `PAGE 2` Offset Condition
- [x] Add Current Page Tooltip
- [x] Test Paging Options & Behavior

### Description
This Hotfix PR resolves an inaccurate Pagination display counts due when the filtered result set contains 0 records, along with required improvements to pagination slicing with respect to start and end offsets. Max-page detection has also been corrected.

#1044 

 
### Screenshots

Screen Shot 2022-01-07 at 8 51 56 AM

Screen Shot 2022-01-07 at 2 22 46 PM
Screen Shot 2022-01-07 at 2 22 25 PM
Screen Shot 2022-01-07 at 2 24 35 PM

### DIFF
--- a/config-ui/src/pages/pipelines/activity.jsx
+++ b/config-ui/src/pages/pipelines/activity.jsx
@@ -362,9 +362,9 @@ const PipelineActivity = (props) => {
                                     <p>Are you Sure you want to cancel this <strong>Run</strong>?</p>
                                     <div style={{ display: 'flex', width: '100%', justifyContent: 'flex-end' }}>
                                       <Button
-                                        text='CANCEL' minimal
+                                        text='NO' minimal
                                         small className={Classes.POPOVER_DISMISS}
-                                        stlye={{ marginLeft: 'auto', marginRight: '3px' }}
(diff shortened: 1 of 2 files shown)


e2corporation:
This is ready and should get merged with release. @joncodo @klesh

e2corporation:
I need to update the for when the record count is less than the per page count. Making another small adjustment.

e2corporation:
This is ready for re-verification and merge @joncodo @klesh

e2corporation:
I still need to improve the *End Set Count* for cases where the last page contains less records then the perPage value, but this is safe to merge for now as it resolves the original 0 count problem.

PR #1055, opened 2022-01-10
fix: replaced /env with /ping

kevin-kline:
### Description
Replaces /env endpoint with /ping endpoint, since we no longer want /env.

Closes https://github.com/merico-dev/lake/issues/1033


e2corporation (approved):
LGTM for `config-ui`

PR #1056, opened 2022-01-10
fix: label color change

kevin-kline:
### Description
Changes label colors to:

Success = Green
Warning = Orange
Danger = Red

Closes https://github.com/merico-dev/lake/issues/797

### Screenshots
image

### Other Information
Without refactoring, this is the quickest way to achieve the result. The coupling is due to the reliance on blueprintJS concept of "labelIntent". Otherwise I would've just added new colours.

I realize this will change these defaults in other areas. That's up to the team to decide if it's OK design wise.
@joncodo @e2corporation @Startrekzky

### DIFF
--- a/config-ui/src/pages/configure/settings/jira.jsx
+++ b/config-ui/src/pages/configure/settings/jira.jsx
@@ -184,7 +184,7 @@ export default function JiraSettings (props) {
 
       <MappingTag
         labelName='Requirement'
-        labelIntent='primary'
+        labelIntent='success'
         typeOrStatus='type'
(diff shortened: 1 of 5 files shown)


e2corporation:
@kevin-kline I think it would be better to use inlined local style overrides for these tags as these global changes would change the tone elsewhere throughout the UI.

e2corporation:
Alternatively setting the 3rd label to intent='primary' and the first label to intent='warning would be another workaround.

kevin-kline:
> 

Ok.

e2corporation (approved):
LGTM

Issue #1061, opened 2022-01-12
Missing API Validation Layer - Results in lost data


kevin-kline:
## Description:

- API validation is needed to ensure response data from APIs are consistent with what we expect
- Every entity we request from an API gets converted from a response to a Go struct prior to saving in our DB
    - Ex. Commits from GitHub
- If the responses from the API change in any way from what we expect to get back, then we risk losing data
- This can happen silently because we don’t have validation to ensure the data matches what we expect
- API responses can vary, or be changed intentionally by the provider, so we must be prepared for this
- We have discovered this problem exists in our system in the GitHub plugin, so it may exist in others as well. If not now, it could in the future.

## Example (tested, proven):
    - In the GitHub plugin, only some of our response data contained author data and committer data
    - This is because some users are not verified on GitHub
    - Some commits were not being saved because of the failed conversion
    - We expected certain data, but the responses were inconsistent leading to lost data
    - Without validation we have no way to know this is happening, since this error occurs at runtime

## Screenshots:
1. Commit w/ author
![image](https://user-images.githubusercontent.com/27032263/149185974-2ab58e3b-164c-4fcb-a8d4-c3486af27cc3.png)

2. Commit w/out author
![image](https://user-images.githubusercontent.com/27032263/149186090-52cf253d-e470-4a07-8172-44c4ef5dbcec.png)

3. Proof of why author / committer is not found (no valid user account on GitHub)
![image](https://user-images.githubusercontent.com/27032263/149186219-8adaacab-1ba0-4411-b884-64aded026131.png)

4. Problem in the code
![image](https://user-images.githubusercontent.com/27032263/149186323-e089134d-0b5f-46d6-9f97-4891303b91c4.png)

## Possible Solutions:
- Add validation layer
- Maybe this package: https://pkg.go.dev/gopkg.in/go-playground/validator.v8#section-readme  

@joncodo Please comment / edit with your understanding if needed

Startrekzky:
root-reason: no error swallowing so the bug doesn't expose in development and QA stage

PR #1062, opened 2022-01-13
Fix Github commits data lost

klesh:
# Summary

Fix issues #1061

### Key Points


mindlesscloud (approved):

joncodo:
I know that this is a good thing to merge but I want to make sure it satisfies the problems that @hezyin brought up to me yesterday about a more general fix since we will need many "if" checks across the system under these circumstances. Having said that, I think it might be the only way.

joncodo:
- [ ] QA this
- [ ] Check for error messages in the logs to be sure they make sense

Issue #984, opened 2021-12-20
Refactor enrichers/converters from all plugins to adopt cursor and batch insertion approach


klesh:
## Description
When dataset grows too big, we have two things to consider:
1. OOM might occur if we keep loading all rows from table into memory
2. Enrichment/Conversion become super slow

## Describe the solution you'd like
1. All data reading should use cursor instead of `gorm.DB.Rows` method to avoid OOM
2. Insert/Update should be operated in `Batch Mode`

Example implementation can be seen here:  https://github.com/merico-dev/lake/pull/967/files#diff-323d48ffd286c2e4453d8c77076e3ceeaf8d8f23a7757357697a3821bc3f7e43R12

## Has the Feature been Requested Before?
No

klesh:
Addressed by #1287

PR #1063, opened 2022-01-13
Kk 984 batch insert

joncodo:
WIP


kevin-kline:
Update:

GitLab Commits collector provides a working example of the following concept:
1. To save in batches, we want to append response data from the API to slices instead of saving to DB on each response.
2. Once all data has been appended to slice(s), we save the data in batches by using GORM's native Create() functionality.

I've ran collection multiple times without seeing any data loss, however @hezyin has advised that we be careful about read/writing to a slice from different goroutines because they are not thread-safe: https://stackoverflow.com/questions/44152988/append-not-thread-safe

I will think more on how to solve this. If anyone has ideas, please chime in @klesh @joncodo

kevin-kline:
> 

Another Update:

Found a solution here: http://dnaeon.github.io/concurrent-maps-and-slices-in-go/

I've implemented this on GitLab Commits collection. Works perfectly fine. 

The other way to do it is via channels, which might be a better overall solution. @hezyin provided a link that speaks more about the two options we have: https://www.oreilly.com/library/view/concurrency-in-go/9781491941294/ch04.html

origin_img_v2_39cc5eb6-51c9-40d2-aeb3-28db7ca1ea8g

I have opted to implement option #1 (synchronization primitives for sharing memory). The only thing I don't like about it is we have to write a lot of new types to get them to save with GORM in batches, since GORM reads the type from the slice. The example uses an interface, but GORM won't understand that. (unless I am missing something, please correct me if I'm wrong!)

@klesh @joncodo Please feel free to weigh in on this decision. I think option #1 is perfectly fine, and I will continue with it. Option #2 might be better overall but a little more tricky to refactor.

klesh (commented):
The #984 is saying to work on enrichers/converters, why we are working on `collectors` instead?
The ROI of converting `collectors` is low, because the bottleneck of collector is Network I/O, not database, and we are already fetching data in parallel, I don't think we can gain much out of this.

kevin-kline:
After discussing with @klesh we've decided to work only on enrichers/converters for now. Closing for now.

Issue #950, opened 2021-12-10
Resolve the association between PR and commit and store it in the domain layer


Startrekzky:
## Describe the solution you'd like
There are two ways to resolve this issue:
1. Parse the PR number from the commit message.
2. Fetch the association from GitHub/GitLab REST API:
  - https://docs.github.com/en/rest/reference/pulls#get-a-pull-request
  - https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr

## Belongs to issue
#949

PR #1107, opened 2022-01-23
feat(github plugin): resolve github commits and pull requests

warren830:
# Summary
1. Extracted func CollectPullRequestCommits from CollectChildrenOnPullRequests as a standalone task
2. Resolved the relationship between pr and commit and stored as a new table pr_commit

### Screenshots
image

Will close issue #950


klesh:
pr/commits relationship is reprsented in comma separated format? it doesn't seem right.

warren830:
> pr/commits relationship is reprsented in comma separated format? it doesn't seem right.

Fixed in the new pr

plugins/github/tasks/github_pull_request_commit_convertor.go
```diff
@@ -0,0 +1,35 @@
...
+
+func PrCommitConvertor() (err error) {
+	githubPullRequestCommit := &models.GithubPullRequestCommit{}
+
+	cursor, err := lakeModels.Db.Model(&githubPullRequestCommit).Rows()
+	if err != nil {
+		return err
+	}
+	defer cursor.Close()
+	// iterate all rows
```
klesh, line 18: i think we should delete pr/commits relationship before insertion considering pr might be force pushed.

warren830: I deleted the old relationship when collecting, is that right?

klesh (approved):
LGTM

Issue #951, opened 2021-12-10
Extract type and component from GitHub PR labels, and store them in the pull_requests table


Startrekzky:
## Purpose
Standardize the types and modules of GitHub PR according to the rules set by the user, so that `commits` can be analyzed in these 2 dimensions.

## Describe the solution you'd like
Solution: Implemented in 2 stages (P0 and P1)
Affected plugin: GitHub

P0: configuration via `.env`
The user needs to configure the GitHub Plugin in 2 ways by entering a regular expression that matches the label
- `PR_TYPE:/*type.*$/`
- `PR_COMPONENT:/*component.*$/`

After user configuration, when GitHub Plugin does the tool layer -> domain layer enrichment of the PR entity:
- If there is no label under PR that meets PR_TYPE, PR_COMPONENT, then set the `type` and `component` of table.pull_requests to null (if go is not null, then it is an empty string)
- Set the 1st label under PR that satisfies the PR_TYPE, PR_COMPONENT regularity as the value of the `type` and `component` of table.pull_requests. (Why: In some edge cases, users may tag the same PR with multiple components, e.g. component/plugins, component/config-ui)
<br>
P1 (For Later): Configuration via 'config-UI'.

## Belongs to issue
#949

Startrekzky:
Already fixed.

PR #1108, opened 2022-01-24
feat(plugins-github): enrich pull request with labels

warren830:
Users can config regular expression for PR_TYPE and PR_COMPONENT
And the github plugin will enrich pull requests with labels which meet the expression configured

### Screenshots
type and component have been enriched:
image

### issue
will close issue #951

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,6 +65,10 @@ FEISHU_APPSCRECT=
 GITHUB_ENDPOINT=https://api.github.com/
 GITHUB_AUTH=***
 GITHUB_PROXY=
+# GITHUB_PR_TYPE=type/(.*)$ the program will extract the value in (), in this example, you will get "refactor" from "type/refactor"
+GITHUB_PR_TYPE=
+# GITHUB_PR_COMPONENT=component/(.*)$ the program will extract the value in (), in this example, you will get "plugins" from "component/plugins"
+GITHUB_PR_COMPONENT=
 
 ##########################
 # ConfigUI configuration #
--- a/plugins/github/github.go
+++ b/plugins/github/github.go
@@ -99,24 +99,25 @@ func (plugin Github) Execute(options map[string]interface{}, progress chan<- flo
 	}
 	if len(tasksToRun) == 0 {
 		tasksToRun = map[string]bool{
-			"collectRepo":          true,
-			"collectCommits":       true,
-			"collectIssues":        true,
-			"collectIssueEvents":   true,
-			"collectIssueComments": true,
-			"collectIssueLabels":   true,
-			"collectPrReviews":     true,
-			"collectPrLabels":      true,
-			"collectPrCommits":     true,
-			"collectPrComments":    true,
-			"enrichIssues":         true,
-			"convertRepos":         true,
-			"convertIssues":        true,
-			"convertPrs":           true,
-			"convertCommits":       true,
-			"convertPrCommits":     true,
-			"convertNotes":         true,
-			"convertUsers":         true,
+			"collectRepo":                true,
+			"collectCommits":             true,
+			"collectIssues":              true,
+			"collectIssueEvents":         true,
+			"collectIssueComments":       true,
+			"collectIssueLabels":         true,
+			"collectPullRequestReviews":  true,
+			"collectPullRequestLabels":   true,
+			"collectPullRequestCommits":  true,
+			"collectPullRequestComments": true,
+			"enrichIssues":               true,
+			"enrichPullRequests":         true,
+			"convertRepos":               true,
+			"convertIssues":              true,
+			"convertPullRequests":        true,
+			"convertCommits":             true,
+			"convertPullRequestCommits":  true,
+			"convertNotes":               true,
+			"convertUsers":               true,
 		}
 	}
 
@@ -133,7 +134,6 @@ func (plugin Github) Execute(options map[string]interface{}, progress chan<- flo
 		}
 		tasks.CollectChildrenOnCommits(ownerString, repositoryNameString, repoId, scheduler, githubApiClient)
 	}
-
 	if tasksToRun["collectIssues"] {
 		progress <- 0.1
 		fmt.Println("INFO >>> starting issues / PRs collection")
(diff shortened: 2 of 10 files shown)


.env.example
```diff
@@ -78,3 +82,11 @@ GRAFANA_ENDPOINT=http://localhost:3002
 AE_APP_ID=
 AE_SECRET_KEY=
 AE_ENDPOINT=
+
+
+########################
+# GitHub PullRequest configuration #
+########################
+
+PR_TYPE="/*type/(.*$)"
```
klesh, line 91: dupcated to line 69
value should be `type/(.*)$`
should add `GITHUB_` as prefix

.env.example
```diff
@@ -78,3 +82,11 @@ GRAFANA_ENDPOINT=http://localhost:3002
...
+PR_COMPONENT="/*component/(.*$)"
```
klesh, line 92: same problems as previous one

plugins/github/github.go
```diff
@@ -98,16 +98,17 @@ func (plugin Github) Execute(options map[string]interface{}, progress chan<- flo
...
-			"convertCommits": true,
-			"convertNotes":   true,
-			"convertUsers":   true,
+			"collectRepo":            true,
+			"collectCommits":         true,
+			"collectIssues":          true,
+			"enrichIssues":           true,
+			"convertRepos":           true,
+			"convertIssues":          true,
+			"pullRequestLabelEnrich": true,
```
klesh, line 107: convetion: verb before name, `enrichPrs` is more preferable

klesh: our purpose is to enrich pr, right now we utilize `label` only, doesn't mean we can use it only. 
from user perspective, they care the outcome, not the source.

plugins/github/github.go
```diff
@@ -178,6 +179,13 @@ func (plugin Github) Execute(options map[string]interface{}, progress chan<- flo
 			return err
 		}
 	}
+	if tasksToRun["pullRequestLabelEnrich"] {
+		progress <- 0.6
+		err = tasks.EnrichGithubPullRequestWithLabel()
```
klesh, line 184: we can name this method specifically like this internally, no problem.
just remember to change the `tasksToRun` naming because it is comming from client side request.

plugins/github/tasks/github_pull_request_label_enricher.go
```diff
@@ -0,0 +1,70 @@
...
+func setPullRequestLabel(label string, pr *githubModels.GithubPullRequest) {
+	var labelRegex *regexp.Regexp
+	// if pr.Type has not been set nad prType is set in .env, process
+	if prType != "" && pr.Type == "" {
+		labelRegex = regexp.MustCompile(prType)
+	}
+	if labelRegex != nil {
+		labelString := labelRegex.Find([]byte(label))
+		if len(labelString) > 0 {
+			pr.Type = string(labelString)
```
klesh, line 54: use sub pattern instead of full match

plugins/github/tasks/github_pull_request_label_enricher.go
```diff
@@ -0,0 +1,70 @@
...
+			return
+		}
+	}
+
+	// if pr.Component has not been set and prComponent is set in .env, process
+	if prComponent != "" && pr.Component == "" {
+		labelRegex = regexp.MustCompile(prComponent)
+	}
+	if labelRegex != nil {
+		labelString := labelRegex.Find([]byte(label))
```
klesh, line 64: same as above

plugins/github/tasks/github_pull_request_enricher.go
```diff
@@ -0,0 +1,69 @@
...
+	// iterate all rows
+	for cursor.Next() {
+		err = lakeModels.Db.ScanRows(cursor, githubPullRequst)
+		if err != nil {
+			return err
+		}
+		var pullRequestLabels []string
+
+		lakeModels.Db.Table("github_issue_labels").
+			Where("issue_id = ?", githubPullRequst.GithubId).Order("updated_at ASC").
```
warren830, line 29: not sure if we should add .Oder("updated_at ASC"), in the issue, it said if there are more than one, we need to extract the first one.

plugins/github/github.go
warren830: ok, fixed in the new version

plugins/github/github.go
warren830: ok, fixed in the new version

.env.example
warren830: sorry for this careless problem

plugins/github/tasks/github_pull_request_enricher.go
klesh: ## generally spearking
we should avoid `order by`  non-index field. 
`updated_at` and `created_at` are filled by us during database manipulation, they are not from user data source. nor do they imply any particular order because we are fetching data in parallel anywhere possible. So, most likely it make no sense to order by these two fields.
it may make sense to order by `github_updated_at`, but if that was the case, we have to consider adding index to this field at the same time.

## this case
as for this particular case, no proof shows we have to consider the timing factor, i dont think we need to `order` by any field.

plugins/github/tasks/github_pull_request_enricher.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+	}
+	return nil
+}
+
+func setPullRequestLabel(label string, pr *githubModels.GithubPullRequest) {
+	var labelRegex *regexp.Regexp
+	// if pr.Type has not been set and prType is set in .env, process the below
+	if prType != "" && pr.Type == "" {
+		labelRegex = regexp.MustCompile(prType)
```
klesh, line 50: `labelRegex` is local variable, everytime `setPullRequestLabel` got invoked, regexp got recompiled.
we should make it global and initialize it in `init` function once to avoid unneccessary compiliation

plugins/github/tasks/github_pull_request_enricher.go
```diff
@@ -0,0 +1,71 @@
...
+		groups := labelRegex.FindStringSubmatch(label)
+		if len(groups) > 0 {
+			pr.Type = groups[1]
+			return
+		}
+	}
+
+	// if pr.Component has not been set and prComponent is set in .env, process
+	if prComponent != "" && pr.Component == "" {
+		labelRegex = regexp.MustCompile(prComponent)
```
klesh, line 62: should use another global variable for `component`

warren830: Fixed, thx!

klesh (approved):
LGTM

PR #1116, opened 2022-01-25
feat: link tool layer github issues to commit

joncodo:
Fixes #https://github.com/merico-dev/lake/issues/1113


klesh (commented):
the fix is legit, but the commit message doesn't seem right!
the code is linking to repository, but the commit message saying "to commit"?

klesh (approved):
LGTM

PR #1142, opened 2022-01-28
fix: fix too many connections

mindlesscloud:
# Summary
fix the "too many connections" error by setting up a connection pool 

### Key Points


klesh (approved):
LGTM

PR #1172, opened 2022-01-30
fix: remove build section from release yml

klesh:
# Summary

remove build section from release yml and make sure `docker-compose up` can work

### Key Points


Startrekzky (approved):
LGTM

PR #1181, opened 2022-02-07
refactor(domain layer,plugins): indicate which sub task failed

### LINKED ISSUE #1050: Improve pipeline debugging experience
## Description
Currently, pipeline contains only error message itself, which is not so helpful for user to understand what went wrong in most cases, @narrowizard suggested we may add more information to help both user and us to identify which step went wrong.

1. Add field `FailedSubTask` to `tasks` table to indicate which step failed
2. Define a `SubTaskError` struct with fields `SubTaskName` along side error message to hold information during runtime
3. All `plugin.Execute` return `SubTaskError` when it failed on sub task with specific `SubTaskName`
4. `services/task.RunTask` check if returned error was a `SubTaskError` and save `SubTaskName` as `FailedSubTask` into `tasks` table.

By doing so, user and us can know better which step did error ocurr.

No

## Additional context
EE need this for a better user experience.
image
dont know which step failed, not helping

warren830:
# Summary

To improve pipeline debugging experience, this change will indicate which sub task failed
1. added an extra field(failed_sub_task) in table.tasks
2. let all plugin.execute return SubTaskError if sub task failed
3. in services/task.RunTask, check if error is SubTaskError

###Todo
let other plugins.execute to return SubTaskError

closes #1050

### Screenshots
image

### DIFF
--- a/errors/sub_task_error.go
+++ b/errors/sub_task_error.go
@@ -1,0 +1,16 @@
+package errors
+
+var _ error = (*SubTaskError)(nil)
+
+type SubTaskError struct {
+	SubTaskName string
+	Message     string
+}
+
+func (e *SubTaskError) Error() string {
+	return e.Message
+}
+
+func (e *SubTaskError) GetSubTaskName() string {
+	return e.SubTaskName
+}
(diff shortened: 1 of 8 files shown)


klesh (approved):
LGTM

PR #1184, opened 2022-02-08
`Docs`: Use curl instead of git-clone to prepare the configuration files

daniel-hutao:
Signed-off-by: Daniel Hu <tao.hu@merico.dev>

# Summary

Use curl instead of git-clone to prepare the `.env & docker-compose.yml` files

### Key Points

### Description

It's too heavy to let users clone the whole repo to get the two configuration files, just `curl` them.

### Screenshots

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,12 +194,11 @@ NOTE: After installing docker, you may need to run the docker application and re
 
 #### Commands to run in your terminal<a id="user-setup-commands"></a>
 
-1. Clone repository:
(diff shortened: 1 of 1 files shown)


klesh:
@daniel-hutao you can work on released version, like https://github.com/merico-dev/lake/tree/main/releases/lake-v0.7.0 
released versions use fixed docker image tag, itself can function properly without repo.

Issue #1199, opened 2022-02-11
User reported bug of jenkins


klesh:
## Describe the bug
![cc796f49f3f68e5436fcc4cff1d5b2a](https://user-images.githubusercontent.com/61080/153572351-2e58d3e7-07a0-40cd-b726-908289fc654e.jpg)

Looks like `SaveBuilds` from https://github.com/merico-dev/lake/blob/8af58c0af5e43d361f87c4e3f507c8ad7ab5c9f5/plugins/jenkins/tasks/storage.go#L57 unable to handle empty slice

I suggest we go through this plugin throughly, to prevent this kind of bug altogether, since this is very similar to what user reported last time (build has empty job slice)

## To Reproduce
No sure how to reproduce

## Expected behavior
1. jenkins should be able to handle all kinds of empty slices

warren830:
Plan of attack:

Startrekzky:
Resolved.

PR #1204, opened 2022-02-13
Fix jenkins save builds

### LINKED ISSUE #1207: extract related commit sha for jenkins build
## Description

Looks like jenkins keeps built commit sha, and it can be collected from api
image
It's valuable for user to obtain this information for analysis.

Collect this information and convert it to domain layer accordingly

Related to #894

warren830:
# Summary

Refactor jenkins sub tasks, to be the same structure as other plugins

### Description
694303d6e4def5c1896c638f45b56c9e8d8f4255:
Instead of old method, this pr make sub tasks more clear, we can know which file do which sub task directly
Also, after several tests, createInBatches can handle empty slice errors(create an empty slice, and pass it into createInBatches)

2c93653b5db18d505a3879073b375bc601c5ecac:
store commit sha

694303d6e4def5c1896c638f45b56c9e8d8f4255 will close issue #1199, 
2c93653b5db18d505a3879073b375bc601c5ecac will close #1207

### Other Information
I checked the total number of builds, the number in db equals to that in the jenkins server
auto init gitlab - 
db: 66
server: 66
auto_test - 
db: 402
server: 402
Create_license - 
db: 203
server: 203
Create_license_new - 
db: 10
server: 10
Deploy k8s single_new - 
db: 10
server: 10
Deploy k8s test -  
db: 2173
server: 2173


plugins/jenkins/tasks/jenkins_build_collect.go
```diff
@@ -0,0 +1,80 @@
...
+			// so we need to filter it manually
+			// timestampHalfYearAgo := time.Now().AddDate(0, -6, 0).Unix() * 1000
+			// var filteredData = make([]models.JenkinsBuildProps, 0)
+			// for _, item := range builds.Builds {
+			// 	if item.Timestamp >= timestampHalfYearAgo {
+			// 		filteredData = append(filteredData, item)
+			// 	}
+			// }
+
+			var filteredData = builds.Builds
```
klesh, line 47: We should rename the variable if it wasn't actually got filtered

klesh (commented):
LGTM overall, some minor fixes are needed
Can we also add a small section onto README to explain the relationship between `build` and `job`? I keep forgetting which belongs to which

plugins/jenkins/tasks/jenkins_build_collect.go
klesh: second thought, we should filter it, I think, because once build is done, it is done, it makes no sense to update a existing build record.

warren830: I have changed the code to filter all builds after the last build in db

plugins/jenkins/tasks/jenkins_build_collect.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/merico-dev/lake/logger"
+	lakeModels "github.com/merico-dev/lake/models"
+	"github.com/merico-dev/lake/plugins/jenkins/models"
+	"github.com/merico-dev/lake/utils"
+	"gorm.io/gorm/clause"
+	"time"
+)
+
+func CollectBuilds(worker *JenkinsApiClient, scheduler *utils.WorkerScheduler) error {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
warren830: ok, will fix it right now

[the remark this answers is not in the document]
warren830: > I suspect this `ctx` should be using the one got passed in at `plugin.Execute` method , so that the canceling could function?
FIxed in 6758bda560d659c038eb7a6a06dc849e19b134c6

plugins/jenkins/tasks/jenkins_build_collect.go
warren830: Also updated doc in the newest push

plugins/jenkins/tasks/jenkins_build_enricher.go
```diff
@@ -0,0 +1,62 @@
...
+		err = lakeModels.Db.ScanRows(cursor, build)
+
+		if build.CommitSha != "" {
+			continue
+		}
+
+		err = scheduler.Submit(func() error {
+			logger.Debug("(enrich build) Submit", build.DisplayName)
+
+			job, err := apiClient.jenkins.GetJob(ctx, build.JobName)
```
klesh, line 34: it doesn't seem right to access api inside a Enricher, an Enricher should only depends on database, not external data

plugins/jenkins/tasks/jenkins_build_enricher.go
```diff
@@ -0,0 +1,62 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to get jenkins job: %v", err)
+			}
+			rawBuild, err := job.GetBuild(ctx, build.Number)
```
klesh, line 38: We should move this collecting operation to collecor instead

plugins/jenkins/tasks/jenkins_build_enricher.go
warren830: I fixed, cause there are some other problems in this way

plugins/jenkins/tasks/jenkins_build_enricher.go
warren830: Sorry, already moved

klesh (approved):
LGTM

PR #1210, opened 2022-02-16
fix(refdiff): fix progress

warren830:
# Summary
When calculating progress, use index + 1 to calculate as below: 
progress <- float32((i + 1) / lenCommitPairs);
set the initial progress to 0.00;
otherwise, 1% will appear first, then 0%


klesh (approved):
LGTM

Issue #1069, opened 2022-01-17
Support sprint metrics based on sprint participants


Startrekzky:
## Description
As a PM, I want to see each sprint participant's schedules and the completion rate of a sprint, such as how many codes have been committed, how many issues have been assigned and accomplished, etc.

## Metrics to be supported
- [ ] `Dev Eq`: Dev Eq of each sprint participant
     Definition: Find all the issues that the sprint participant participated in under the sprint and the time period in which they participated, and calculate the sum of the issue-associated commits whose authored_date belongs to this time period.
- [ ] `Issue(C)`: Count of completed issues in which the sprint participant has participated
- [ ] `Issue(P)`: Count of participated issues in which the sprint participant has participated
- [ ] `Story Point(C)`: Sum of the story points of completed issues in which the participant has participated
- [ ] `Story Point(P)`: Sum of the story points of issues in which the participant has participated
- [ ] `Worktime(A)`: Actual work logged by each sprint participant
- [ ] `Fixed Bug`: Count of fixed bugs in which the sprint participant has participated
- [ ] `Fixed Incident`: Count of fixed incidents in which the sprint participant has participated
- [ ] `Avg Incident lead time`: Avg fixing time of fixed incidents that sprint participant has participated
- [ ] `Avg Bug lead time`: Avg fixing time of fixed bugs that sprint participant has participated

Note: 
1. keep `worktime(A)` and remove `worktime(P)`
2. remove metric `workload` as it's too vague

## Describe the solution you'd like
Please make sure the domain layer schema supports the metrics above. 
There might be few things to check 

- [ ] If the raw data for those metrics exist in the domain layer. If not, collect it.
- [ ] Is the existing schema easy to calculate those metrics. If not, design middle tables such as `issue_assignee_history` to make the metrics easier to query.

Startrekzky:
Already done.

PR #1211, opened 2022-02-16
fix(plugins): change assignee id

warren830:
# Summary

Use domain layer user id instead of tool layer in issue_assignee_history
Used onConflict(updateAll : true) when insert table.assignee_history, table.sprints_history, table.status_history
 
### Description

Will not close issue #1069 , but a fix

### Current Behavior
Store tool layer raw jira_issues.assignee_id in table.issue_assignee_history

### New Behavior
Changed to domain layer user id 

### Screenshots
image
### Other Information


klesh (approved):
LGTM

PR #1224, opened 2022-02-17
fix(github): make sure of data integration

### LINKED ISSUE #1222: github pr commits collection should flush record before fetching not inbetween
## Describe the bug
The following code seems like we will keep only the last page of records
image

## To Reproduce
Spotted by code reading, no existing reproduciable step available.

## Expected behavior
All records from all pages should be saved into database.

warren830:
# Summary

To fix issue #1222 

### Key Points

### Description
Moved deleting part out from return func

Closes #1222 

### Current Behavior
Flush records during fetching data by pages

### New Behavior
Flush data before fetching


klesh (approved):
LGTM

PR #1263, opened 2022-02-22
fix: multithread read empty config

mappjzc:
# Summary
fix: multithread read empty config
refactored configuration file processing

### Key Points

### Description
Fix multithreading unsafe issue of flushing global variables in LoadConfigFile().
Fix the problem that the use of local variables in init() of config.go may cause invalid operation.

Delete the ConfigJson data structure
Remove LoadConfigFile() function
Add the GetConfig() function to privatize the Config global variable
Some Chinese annotations were killed because of klesh
All the unified calling functions for Config requirements in the adjustment code are GetConfig()

close https://github.com/merico-dev/lake/issues/1243

### Current Behavior

### New Behavior

### Screenshots

### Other Information

### DIFF
--- a/e2e/database.go
+++ b/e2e/database.go
@@ -10,8 +10,8 @@ import (
 )
 
 func InitializeDb() (*sql.DB, error) {
-	V := LoadConfigFile()
-	dbUrl := V.GetString("DB_URL")
+	v := LoadConfigFile()
+	dbUrl := v.GetString("DB_URL")
 	db, err := sql.Open("mysql", dbUrl)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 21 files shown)


e2e/database.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 func InitializeDb() (*sql.DB, error) {
-	V := LoadConfigFile()
-	dbUrl := V.GetString("DB_URL")
+	v := LoadConfigFile()
```
klesh, line 13: I think this one should use `GetConfig` instead?

klesh (approved):
LGTM

PR #1273, opened 2022-02-23
Add gitlab commits' parent and fix a bug

likyh:
# Summary

1. Collect commits' parent  sha in gitlab;
2. fix bug in collect gitlab tags:
  2.1. delete old data before save new tag
  2.2. add `project_id` column in table `gitlab_tags`

### Key Points

no

### Screenshots
image
image

### DIFF
--- a/plugins/gitlab/models/gitlab_commit.go
+++ b/plugins/gitlab/models/gitlab_commit.go
@@ -18,6 +18,7 @@ type GitlabCommit struct {
 	CommitterEmail string
 	CommittedDate  time.Time
 	WebUrl         string
+	ParentIdsStr   string
 	Additions      int `gorm:"comment:Added lines of code"`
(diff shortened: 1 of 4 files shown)


klesh (commented):
DevLake use commit sha as primary for all entities at this moment.
`auto increment id` / `uuid`  are bad for idempotent data synchronization.
We have to discuss before switching

plugins/gitlab/models/gitlab_commit.go
```diff
@@ -18,6 +18,7 @@ type GitlabCommit struct {
 	CommitterEmail string
 	CommittedDate  time.Time
 	WebUrl         string
+	ParentIdsStr   string
```
klesh, line 21: Can we store this relationship into a new table like `gitlab_commit_parents` ?

klesh:
total outdated

PR #1275, opened 2022-02-24
fix(plugins): fix concurrent problems

warren830:
# Summary

when pass the pointer to apiclient.GetAsync, met some concurrent problems

### Key Points

### Description
We should new an object and pass this object's pointer to the goroutine func in every iteration

### Current Behavior
Concurrent problems

### New Behavior
generated an new object and pass the pointer to go func for every iteration

### Screenshots

### Other Information


klesh (approved):
LGTM

Issue #1288, opened 2022-03-01
Backend API does not save `Proxy` field for JIRA, GitLab and Jenkins


e2corporation:
## Config-UI / Data Integrations / Connection **Proxy**
A feature ticket in development has requested that `Proxy` be added for all Data Providers, however only **GitHub** is actually saving the proxy field to backend database. We can limit this field to just one or more providers as well. Waiting on BE Team response on how to proceed.

## Expected behavior
All Data Providers should support a Connection Proxy.

## Screenshots
<img width="466" alt="Screen Shot 2022-02-28 at 11 01 19 PM" src="https://user-images.githubusercontent.com/1742233/156102477-45479d29-18dd-4529-93ff-c29db32d281e.png">

e2corporation:
@klesh Do we need Proxy for all Providers or just GitHub?

e2corporation:
Relates to #1223

klesh:
> @klesh Do we need Proxy for all Providers or just GitHub?

github/gitlab/jira are must, as for jenkins, I think we can add that as well.

klesh:
@e2corporation jira did support `proxy`, but it was capitalized, I put up a PR to make it lowercase, https://github.com/merico-dev/lake/pull/1292

Please take a look at it

klesh:
@e2corporation 
as of the other plugin, I saw sth weird, why we had body of `PUT` different than the one came from GET response?
![image](https://user-images.githubusercontent.com/61080/156133880-c0fafcfd-49bf-4039-bbb4-9a3bf8c0163c.png)
Can we make them identical?

e2corporation:
@klesh We need to have a discussion regarding this, we can't keep changing/modifying the API contract randomly during development without discussing how it can break the UI, the system already expected 'GITLAB_ENDPOINT' and 'GITLAB_AUTH' why would we change these now?

klesh:
> @klesh We need to have a discussion regarding this, we can't keep changing/modifying the API contract randomly during development without discussing how it can break the UI, the system already expected `GITLAB_ENDPOINT` and `GITLAB_AUTH` why would we change these now?

Sure thing, this is under discussion, nothing have been done yet. So, you prefer they being different?

klesh:
> @klesh We need to have a discussion regarding this, we can't keep changing/modifying the API contract randomly during development without discussing how it can break the UI, the system already expected `GITLAB_ENDPOINT` and `GITLAB_AUTH` why would we change these now?

[...]

PR #1292, opened 2022-03-01
fix: jira proxy should be lowercase

klesh:
# Summary

Change proxy field into lowercase, related to #1288

156133192-fc6a02f0-d742-4d08-88c2-3532a016c4ee

### Key Points


klesh:
@e2corporation Please help reviewing

PR #1335, opened 2022-03-15
Refactor git extractor

hezyin:
# Summary

This is a minor refactor PR to get the git extractor conformed to the new `SubTaskEntryPoint` interface. It also removes usage of progress channel and global logger to prepare for the migration to Temporal.

## Verification

After the refactor, I truncated the related tables and re-run the plugin for verification. The number of rows for each table is summarized below (data collected against DevLake git repo):

| Table     | COUNT |
| ----------- | ----------- |
| refs      | 20       |
| commits   | 2757       |
| repo_commits | 2757 | 
| commit_files | 17810 | 
| commit_parents | 3460 |

### DIFF
--- a/plugins/gitextractor/tasks/git_repo_collector.go
+++ b/plugins/gitextractor/tasks/git_repo_collector.go
@@ -36,13 +36,13 @@ func (o GitExtractorOptions) Valid() error {
 	return nil
 }
 
-func CollectGitRepo(taskCtx core.SubTaskContext) error {
-	db := taskCtx.GetDb()
+func CollectGitRepo(subTaskCtx core.SubTaskContext) error {
+	db := subTaskCtx.GetDb()
 	storage := store.NewDatabase(db)
 	defer storage.Close()
-	op := taskCtx.GetData().(GitExtractorOptions)
-	ctx := taskCtx.GetContext()
-	p := parser.NewLibGit2(storage)
+	op := subTaskCtx.GetData().(GitExtractorOptions)
+	ctx := subTaskCtx.GetContext()
+	p := parser.NewLibGit2(storage, subTaskCtx.GetLogger())
 	var err error
 	if strings.HasPrefix(op.Url, "http") {
 		err = p.CloneOverHTTP(ctx, op.RepoId, op.Url, op.User, op.Password, op.Proxy)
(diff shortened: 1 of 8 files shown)


plugins/gitextractor/tasks/git_repo_collector.go
```diff
@@ -0,0 +1,58 @@
...
+	return nil
+}
+
+func CollectGitRepo(taskCtx core.SubTaskContext) error {
+	db := taskCtx.GetDb()
+	storage := store.NewDatabase(db)
+	defer storage.Close()
+	op := taskCtx.GetData().(GitExtractorOptions)
+	ctx := taskCtx.GetContext()
+	p := parser.NewLibGit2(storage)
```
klesh, line 45: Maybe we should pass down `taskCtx` as a field of `LibGit2`
and we need to call `SubTaskContext.IncProgress` for each `commit` we collected
and we dont wnat `libgit2.go` to depend on `github.com/merico-dev/lake/logger` on line 7

plugins/gitextractor/tasks/git_repo_collector.go
```diff
@@ -0,0 +1,58 @@
...
+	var err error
+	if strings.HasPrefix(op.Url, "http") {
+		err = p.CloneOverHTTP(ctx, op.RepoId, op.Url, op.User, op.Password, op.Proxy)
```
klesh, line 48: We can remove `ctx` and use the shared `taskCtx` field because it seems like very single method needed it.

hezyin: Decided to keep this as is since `LibGit2` is better off just taking what it needs (`context.Context` and `core.logger`) and staying unaware of the concept of task or subtask. I could move `context.Context` into a field of `LibGit2` but since it's mostly cosmetic, I'll skip it for now.

plugins/gitextractor/tasks/git_repo_collector.go
hezyin: Dependency on `github.com/merico-dev/lake/logger` has been removed

hezyin:
@klesh All feedbacks have been addressed except batch save. Made sure `plugins/gitextractor/main.go` still works after refactor and also added docs for this standalone entry to the plugin functionality.

klesh (approved):
LGTM

PR #1353, opened 2022-03-18
feat: feishu collector&extractor

abeizn:
image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,6 +57,7 @@ JENKINS_PASSWORD=
 
 FEISHU_APPID=
 FEISHU_APPSCRECT=
+FEISHU_ENDPOINT=https://open.feishu.cn/open-apis/vc/v1/
 
 ########################
 # GitHub configuration #
--- a/plugins/feishu/feishu.go
+++ b/plugins/feishu/feishu.go
@@ -1,24 +1,16 @@
 package main
 
 import (
-	"context"
-	"encoding/json"
 	"fmt"
-	"net/http"
-	"net/url"
-	"os"
-	"strconv"
-	"strings"
-	"time"
-
-	"github.com/faabiosr/cachego/file"
-	"github.com/fastwego/feishu"
+	"context"
+	"github.com/mitchellh/mapstructure"
+	"github.com/merico-dev/lake/errors"
 	"github.com/merico-dev/lake/config"
-	"github.com/merico-dev/lake/logger"
+	"github.com/merico-dev/lake/plugins/helper"
 	lakeModels "github.com/merico-dev/lake/models"
 	"github.com/merico-dev/lake/plugins/core"
-	"github.com/merico-dev/lake/plugins/feishu/apimodels"
 	"github.com/merico-dev/lake/plugins/feishu/models"
+	"github.com/merico-dev/lake/plugins/feishu/tasks"
 	"github.com/merico-dev/lake/utils"
 )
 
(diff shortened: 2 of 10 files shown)


plugins/feishu/feishu.go
```diff
@@ -156,7 +126,7 @@ func main() {
 	go func() {
 		err := PluginEntry.Execute(
 			map[string]interface{}{
-				"numOfDaysToCollect": 80,
+				"numOfDaysToCollect": 8.0,
```
warren830, line 129: I don't think we should use float here

plugins/feishu/models/feishu_meeting_top_user_item.go
```diff
@@ -12,4 +12,5 @@ type FeishuMeetingTopUserItem struct {
 	MeetingDuration string `json:"meeting_duration"`
 	Name            string `json:"name"`
 	UserType        int64  `json:"user_type"`
+	common.RawDataOrigin
```
warren830, line 15: please use common.NoPKModel instead of RawDataOrigin

plugins/feishu/tasks/feishu_api_client.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/merico-dev/lake/utils"
+	"github.com/merico-dev/lake/config"
+	"github.com/faabiosr/cachego"
+	"github.com/faabiosr/cachego/file"
+)
+
+type FeishuApiClient struct{
+	core.ApiClient
+}
+var ServerUrl  = "https://open.feishu.cn"
```
warren830, line 23: According to other plugins, we'd better pass this value rather than define it here

plugins/feishu/tasks/feishu_api_client.go
```diff
@@ -0,0 +1,153 @@
...
+
+type getRefreshRequestFunc func() *http.Request
+type DefaultAccessTokenManager struct {
+	Id                    string
+	GetRefreshRequestFunc getRefreshRequestFunc
+	Cache                 cachego.Cache
+}
+
+// 防止多个 goroutine 并发刷新冲突
+var getAccessTokenLock sync.Mutex
```
warren830, line 33: please also add English comment

plugins/feishu/tasks/feishu_meeting_top_user_item_collector.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+const RAW_MEETING_TOP_USER_ITEM_TABLE = "feishu_meeting_top_user_item"
+
+var _ core.SubTaskEntryPoint = CollectMeetingTopUserItem
+
+func CollectMeetingTopUserItem(taskCtx core.SubTaskContext) error{
+	logger := helper.NewDefaultTaskLogger(nil, "feishu")
+	data := taskCtx.GetData().(*FeishuTaskData)
+	pageSize := 100
```
warren830, line 20: We don't need to define here

.env.example
```diff
@@ -57,6 +57,7 @@ JENKINS_PASSWORD=
 
 FEISHU_APPID=
 FEISHU_APPSCRECT=
+FEISHU_ENDPOINT=https://open.feishu.cn/open-apis/vc/v1/
```
warren830, line 60: we don't need to give this a value

plugins/feishu/tasks/feishu_meeting_top_user_item_collector.go
```diff
@@ -0,0 +1,67 @@
...
+			query.Set("start_time", strconv.FormatInt(input.PairStartTime.Unix(), 10))
+			query.Set("end_time", strconv.FormatInt(input.PairEndTime.Unix(), 10))
+			query.Set("limit", strconv.Itoa(pageSize))
+			query.Set("order_by", "2")
+			return query, nil
+		},
+		ResponseParser: func(res *http.Response) ([]json.RawMessage, error) {
+			
+			body := &apimodels.FeishuMeetingTopUserItemResult{}
+			err := core.UnmarshalResponse(res, body)
```
warren830, line 54: we need to discuss with @klesh if we should save this res directly or save the arrays of the object

plugins/helper/iterator.go
```diff
@@ -45,3 +45,43 @@ func (c *CursorIterator) Close() error {
...
+
+type DatePair struct{
+	PairStartTime time.Time
+	PairEndTime time.Time
+}
+
+func (c *DateInterator) HasNext() bool{
+	return c.Current < c.Days
+}
+
```
warren830, line 65: because the structs above are used only by feishu, please define them inside feishu pkg

warren830 (approved):

PR #1359, opened 2022-03-21
Move `Plugin.Execute` to framework -- I KNOW, THIS IS A HUGE PR, DONT HATE ME! 

klesh:
# Summary

Move `Plugin.Execute` to framework
Decopule global variables

### Key Points


plugins/core/plugin_migration.go
```diff
@@ -0,0 +1,5 @@
+package core
+
+// TODO: Implement this interface to register database migration-scripts for your plugin
+type PlugibMigrations interface {
```
warren830, line 4: rename to pluginMigration

.env.example
```diff
@@ -15,6 +17,12 @@ MODE=release
 NOTIFICATION_ENDPOINT=
 NOTIFICATION_SECRET=
 
+API_TIMEOUT=10
+API_RETRY=3
+API_REQUESTS_PER_HOUR=10000
```
warren830, line 22: I'm thinking about adding a retry interval, If we retry without an interval, it might cause rate limit exceed

warren830 (approved):
LGTM, fantastic design, great way to manage sub tasks. And for the future contributors, it will be much easier to build a new plugin

.env.example
klesh: retry was moved to `ApiAsyncClient`, and using `GetAsync` method, it was under control of rate limit

mindlesscloud (approved):

PR #1391, opened 2022-03-25
fix(github): move regex compile to func

warren830:
# Summary

moved regex compile into func
Already tested enrich after changing config-ui

### Screenshots
image

### Other Information


klesh:
this fixed #1394

klesh (approved):

PR #1398, opened 2022-03-27
Improve demo dashboards

hezyin:
# Summary

This PR does two things:

1. Make the dashboards display assignee name instead of assignee id for better user experience
2. In the detailed commit info page, add back contributors who haven't contributed in the last month or the month before last by simulating outer join with a union of a left join and a right join


klesh (approved):
LGTM

Startrekzky:
This can solve part of the problem, however, if we are joining over 3 tables and only the middle tables have not-null values, the records in those tables will not be shown in the final result.

PR #1392, opened 2022-03-25
feat(pipeline): support pipeline plan

warren830:
## 

blueprint

Summary

modify create pipeline to support pipeline plan
create patch to enable/disable pipeline plan

Users can set schedule on config-ui, then frontend will parse the user input to crontab format and send back to backend.
Please check cron time format in https://crontab.guru/
One more thing, if it's interval, the format should like this "*/5 * * * ", add '' before '/'

Important

As this pr changed id&createdAt&updatedAt api output to camelCase, this will impact "task" in frontend

Other Information

In the last commit, I divided api pipelines to pipeplines and pipelianplans as blow:
r.GET("/pipelines", pipelines.Index)
r.GET("/pipelines/:pipelineId", pipelines.Get)
r.GET("/blueprints", blueprints.Index)
r.GET("/blueprints/:blueprintId", blueprints.Get)
r.POST("/pipelines", pipelines.Post)
r.POST("/blueprints", blueprints.Post)
r.DELETE("/pipelines/:pipelineId", pipelines.Delete)
r.DELETE("/blueprints/:blueprintId", blueprints.Delete)
r.PATCH("/blueprints/:blueprintId", blueprints.PUT)
r.GET("/pipelines/:pipelineId/tasks", task.Index)
r.GET("/ping", ping.Get)
r.POST("/push/:tableName", push.Post)

POST /blueprints

    Request
    {
    	"name": "COLLECT BLUEPRINT",
    	"tasks": [
    		[
    			{
    				"plugin": "github",
    				"options": {
    					"repo": "lake",
    					"owner": "merico-dev"
    				}
    			}
    		]
    	],
    	"enable": true,
    	"cronConfig": "103 13 /13 * *"
    }
    Response
    {
    	"id": 7,
    	"name": "COLLECT BLUEPRINT",
    	"tasks": [
    		[
    			{
    				"plugin": "github",
    				"options": {
    					"owner": "merico-dev",
    					"repo": "lake"
    				}
    			}
    		]
    	],
    	"enable": true,
    	"cronConfig": "103 13 /13 * *",
    	"createdAt": "2022-03-27T10:16:20.046+08:00",
    	"updatedAt": "2022-03-27T10:16:20.046+08:00"
    }

GET /blueprints

    Response
    {
    	"count": 2,
    	"blueprints": [
    		{
    			"name": "COLLECT BLUEPRINT",
    			"tasks": [
    				[
    					{
    						"plugin": "github",
    						"options": {
    							"repo": "lake",
[...]

PR #1403, opened 2022-03-28
Feat create cron jobs

warren830:
# Summary

Created cron jobs in init.go
Related pr: #1392


services/pipeline.go
```diff
@@ -31,15 +31,18 @@ func init() {
 	db.Model(&models.Pipeline{}).Where("status = ?", models.TASK_RUNNING).Update("status", models.TASK_FAILED)
 }
 
-func CreatePipeline(newPipeline *models.NewPipeline) (*models.Pipeline, error) {
+func CreatePipeline(name string, tasks [][]*models.NewTask, pipelinePlanId uint64) (*models.Pipeline, error) {
```
klesh, line 34: why change the function signature? we can add this new field into `NewPipeline`, right?

services/pipeline_plan.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+func init() {
+	v := config.GetConfig()
+	var notificationEndpoint = v.GetString("NOTIFICATION_ENDPOINT")
+	var notificationSecret = v.GetString("NOTIFICATION_SECRET")
+	if strings.TrimSpace(notificationEndpoint) != "" {
+		notificationService = NewNotificationService(notificationEndpoint, notificationSecret)
+	}
+	db.Model(&models.Pipeline{}).Where("status = ?", models.TASK_RUNNING).Update("status", models.TASK_FAILED)
```
klesh, line 20: i think this one should be moved to service/init.go

services/pipeline_plan.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/merico-dev/lake/errors"
+	"github.com/merico-dev/lake/logger"
+	"github.com/merico-dev/lake/models"
+	"gorm.io/gorm/clause"
+	"strings"
+)
+
+func init() {
+	v := config.GetConfig()
+	var notificationEndpoint = v.GetString("NOTIFICATION_ENDPOINT")
```
klesh, line 15: I dont think we need notification service in `pipeline_plan`

services/pipeline_plan.go
```diff
@@ -0,0 +1,101 @@
...
+func GetPipelinePlan(pipelinePlanId uint64) (*models.PipelinePlan, error) {
+	pipelinePlan := &models.PipelinePlan{}
+	err := db.Find(pipelinePlan, pipelinePlanId).Error
+	if err != nil {
+		return nil, err
+	}
+	return pipelinePlan, nil
+}
+
+func ModifyPipelinePlan(newPipeline *models.NewPipeline, pipelinePlanId uint64) (*models.PipelinePlan, error) {
```
klesh, line 70: We should use PipelinePlan struct instead of `NewPipeline` for this function

services/pipeline_plan.go
warren830: Sorry, I should understand this code clearly before copy&paste

services/pipeline_plan.go
warren830: Ah, understood

warren830: But we need to use newPipeline to modify pipelinePlan in service

services/pipeline.go
warren830: fixed

services/pipeline_plan.go
warren830: fixed

PR #1417, opened 2022-03-29
feat: dbt plugin

abeizn:
image
image


plugins/dbtplugin/tasks/dbt_task_data.go
```diff
@@ -0,0 +1,14 @@
+package tasks
+
+// import (
+// 	"github.com/merico-dev/lake/plugins/helper"
+// )
+
+type DbtOptions struct {
+	SelectedModels string   `json:"selectedModels"`
```
klesh, line 8: What if `SelectedModels` was omitted? Will this plugin run all models ?

klesh: Can we optionaly specify a `modelPath` ?

klesh (commented):
Can we support progress?
Maybe by parsing the process `stdout` and extract the information?

plugins/dbtplugin/tasks/dbt_converter.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/merico-dev/lake/plugins/core"
+	
+)
+
+func DbtConverter(taskCtx core.SubTaskContext) error {
+	data := taskCtx.GetData().(*DbtTaskData)
+	models := data.Options.SelectedModels
+	current_dir, _ := os.Getwd()
+	os.Chdir(current_dir+"/plugins/dbtplugin")
+	cmd := exec.Command("bash", "-c", "dbt run --select " + models)
```
klesh, line 15: exec.Command("dbt","run","--select", models...)

mindlesscloud (approved):

PR #1419, opened 2022-03-29
refactor: move `RunPipeline` and `RunTask` logic to `runner`

klesh:
# Summary

  The RunTask logic can be reused between standalone and temporal mode,
  so it is moved into runner, the standalone should be working without
  problem at this point. Next step will be implementing temporal mode

### Key Points


abeizn:
LGTM

abeizn (approved):

PR #1467, opened 2022-04-08
Chore add comments refdiff

warren830:
# Summary

added some comments and changed a func name

### DIFF
--- a/plugins/refdiff/tasks/refs_commits_diffs_calculator.go
+++ b/plugins/refdiff/tasks/refs_commits_diffs_calculator.go
@@ -33,10 +33,12 @@ func CalculateCommitsDiff(taskCtx core.SubTaskContext) error {
 	}
 	commitPairs := make([][4]string, 0, len(pairs))
 	for i, refPair := range pairs {
+		// get new ref's commit sha
 		newCommit, err := ref2sha(refPair.NewRef)
 		if err != nil {
 			return fmt.Errorf("failed to load commit sha for NewRef on pair #%d: %w", i, err)
 		}
+		// get old ref's commit sha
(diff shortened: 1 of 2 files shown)


plugins/refdiff/tasks/refs_commits_diffs_calculator.go
```diff
@@ -68,7 +70,7 @@ func CalculateCommitsDiff(taskCtx core.SubTaskContext) error {
 		if err != nil {
 			return fmt.Errorf("failed to read commit from database: %v", err)
 		}
-		commitNodeGraph.AddSide(commitParent.CommitSha, commitParent.ParentCommitSha)
+		commitNodeGraph.BuildParents(commitParent.CommitSha, commitParent.ParentCommitSha)
```
mappjzc, line 73: Here create two node (if not be created) and only one parent.
I am not sure whether BuildParents is a good name.

plugins/refdiff/tasks/refs_commits_diffs_calculator.go
```diff
@@ -114,7 +116,7 @@ func CalculateCommitsDiff(taskCtx core.SubTaskContext) error {
 			continue
 		}
 
-		lostSha, oldCount, newCount := commitNodeGraph.CalculateLost(pair[1], pair[0])
+		lostSha, oldCount, newCount := commitNodeGraph.CalculateDiffSha(pair[1], pair[0])
```
mappjzc, line 119: lost not diff
because we only caculate what the new have and the old don't have.
we ignore the old have but the new don't have.
so it is not the diff about the new and the old
it is only caculate what the old lost from the new.

mappjzc (approved):

Issue #1501, opened 2022-04-12
v0.10.0 gitlab test


warren830:
### Plan of attack

Dividing test tasks into two categories, each of them has different workflow:

1. plugin test
2. framework test

#### Plugin Assignement:

- gitlab

For each of every plugin, the following operations are required, and must be presented as **Checklist** in **specific testing issue**:

- [x] Main readme and plugin readme are up to date
- [x] Trigger plugin with only required options and record following information
  - [x] Which board/repo you collected against? Project-Id = 116***
  - [x] How many subtasks are executed by default? 22
  - [x] How long did it take to finish the whole process? 13 minutes
  - [x] How many records did each subtasks generated, and how many shoud it be according to data source if available?
  - [x] What do you think of the plugin status? Is good enough for you? performance-wise, data-wise.
             We can consider converting pipeline to build
- [x] Trigger plugin with `tasks` field for those `EnabledByDefault` set to `false`(gitlab does not have optional tasks)
  - [x] Which board/repo you collected against?
  - [x] What subtasks did you specify?
  - [x] How long did it take to finish the whole process?
  - [x] How many records did each subtasks generated, and how many shoud it be according to data source if avaiable?
  - [x] What do you think of the plugin status? Is good enough for you? performance-wise, data-wise.
- [x] Screen-shots or test-snipet that can show plugin works properly.
![image](https://user-images.githubusercontent.com/39366025/163744631-2ef1d908-d86d-4c5d-aa8f-fa6f7c1160e5.png)

- [x] Create issue for each and every bug you found during the whole process, and add link to **plugin testing issue**
- [x] Fill out table on #1500 to what domain layer entities could a plugin provide, `default` means it was produced by default, `optional` means it was produced only when specified by using `tasks` field (aka. `EnabledByDefault=false`

| subtasks  | api generated   | data source | 
| -------------------- | ---- | ------ |
| 1. CollectProjectMeta               |   1   |   1     | 
| 2. ExtractProjectMeta        |   1   |   1     | 
| 3. CollectCommitsMeta        |   7556   |    7556   |
| 4. ExtractCommitsMeta:       |   9794(including both _raw_gitlab_api_commit and _raw_gitlab_api_merge_request_commits)    |   748   |
| 5. CollectTagMeta        |   156   |   156    |
| 6. ExtractTagMeta:       |   156   |   156    |
| 7. CollectApiMergeRequestsMeta:      |   2062   |   2062    |
[...]

PR #1541, opened 2022-04-13
fix(framework): return the result of retry instead of assign it to err

### LINKED ISSUE #1506: Nil pointer error after "context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
## Describe the bug
If we failed to request from remote server, the system reported nil pointer error.
This happened because the handler would be called twice if retry of api request occurred. 
When the handler is called at the second time, the res is still nil and the err is also nil.

## How to fix
We directly return the result of retry.

## Screenshots
image

warren830:
# Summary
We return the result of retry instead of assign it to err

Close #1506 

### Other Information
relate to #1501


klesh (approved):
LGTM

PR #1589, opened 2022-04-15
fix(gitlab): fix pipeline

### LINKED ISSUE #1575: Gitlab pipeline collector work improperly
## Describe the bug
1. From datasource, we can see there are more than 1000 pipelines, but we only collect 100

## Expected behavior
The number of pipelines collected should be the same as datasource

## Screenshots
image

warren830:
# Summary
1. delete GetTotalPages in pipeline collector.
As there is no "X-Total-Pages" in response.header, but we set GetTotalPages for collector, so the request will be sent only once
### Key Points

### Description

close #1575

### Current Behavior

### New Behavior

### Screenshots
image

### Other Information


klesh (approved):
LGTM

PR #1612, opened 2022-04-15
fix: all pipelines triggered by different blueprints have same name

### LINKED ISSUE #1607: Names of pipelines triggered by blueprint doesn't seem right (seemingly backend problem)
## Describe the bug

blueprints i created this morning
image

pipeline runs that triggered afternoon 
image

problem
The timing seems to be alright, but the names doesnt, they should not be identical

## Expected behavior
Names of pipelines trigger by `hourly test` should begin with `hourly test` , not `specify time 2pm 15th april CST` which is the name of another blueprint

klesh:
Fixes #1607


hezyin (approved):
LGTM

Issue #1621, opened 2022-04-15
Config-ui: 'Create Blueprint' dialog doesn't disappear after 'Create Blueprint'


Startrekzky:
## Describe the bug
'Create Blueprint' dialog doesn't disappear after 'Create Blueprint'

## To Reproduce
1. Go to config-ui, click 'Create Pipeline Run'
2. Toggle on 'Github', input valid repo info
3. Toggle on 'Automate Pipeline', click 'Add Blueprint'
4. Log in blueprint detail, and click 'Save Blueprint'
5. See the result

![image](https://user-images.githubusercontent.com/14050754/163568839-0c874849-801a-42a9-b3f1-24b6fb6a2531.png)

## Expected behavior
The dialogue disappeared.

e2corporation:
@Startrekzky Yes it is still open but in "EDIT" Mode -- user can close and reopen the editor if needed to make changes....

klesh:
I agree with @Startrekzky , it is kind of counter-intuitive, I thought it wasn't saved.
Maybe close the dialog and jump to blueprints page would be helpful, user can then clearly see that blueprint got saved, and they can edit it if needed.

e2corporation:
Closed with PR #1627

PR #1627, opened 2022-04-15
`fix`: create pipeline: auto-close blueprint dialog on save

e2corporation:
### Config-UI / Pipelines / Create Pipeline

- [x] fix: auto-close blueprint dialog on save

### Description
This PR closes the Blueprint Dialog automatically after saving a blueprint.

#1621


hezyin (approved):
LGTM

Issue #1613, opened 2022-04-15
Config-ui: save connection failed in Github connection page


Startrekzky:
## Describe the bug
Save connection failed
Version: v0.10.0

## To Reproduce
1. Go to config-ui, click 'Github' 
2. Click on the default connection in the list
![image](https://user-images.githubusercontent.com/14050754/163546774-99edbf2d-1bbe-41d3-8fd8-517f83793328.png)

3. Enter valid endpoint and url, click 'Save Connection'. It'll show a success message.
![image](https://user-images.githubusercontent.com/14050754/163546992-17d0f25c-b42d-4ddc-9731-a20a08ad4a84.png)
![image](https://user-images.githubusercontent.com/14050754/163547198-883a30c2-cb71-496c-b60b-554889ee2b5f.png)

4. Scroll down to the bottom of the page, click 'Save Settings'. It'll show a success message.
5. Go back to the connection list, see connection 'status'. It's still disconnected.

## Expected behavior
The connection is successfully created after this process, as it's a common user journey to configure GitHub connection.

e2corporation:
@klesh GitHub provider is working as expected on my local `main` branch, if the auth token is bad then it would lead to a bad connection test. Also why was the status message changed in the code from 'offline' to 'disconnected' ?

PR #1639, opened 2022-04-15
`fix`: config-ui: connections - update active connection state after modify

e2corporation:
### Config-UI / Data Integrations / Configure Connection

- [x] Update **Active Connection** Object state after _Modification_
- [x] Test **GitHub** Settings
- [x] Test **JIRA** Settings ⚠️ (Known issues w/ base64 Auth)

### Description
This PR resolves an issue with the **Save Settings** routine for a Connection source where stale connection data is used as settings are saved/modified during the same session as when connection level settings are modified. This creates the undesirable situation of the user overriding previously saved connection data when the decoupled save settings action is used. The Active Connection object is now refreshed after modification so the connection payload is in-sync, preventing any data inconsistency.

#1613

### DIFF
--- a/config-ui/src/hooks/useConnectionManager.jsx
+++ b/config-ui/src/hooks/useConnectionManager.jsx
@@ -154,7 +154,9 @@ function useConnectionManager ({
         setShowError(false)
         setErrors([])
         ToastNotification.clear()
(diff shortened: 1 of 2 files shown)


hezyin (approved):
LGTM

Issue #1649, opened 2022-04-18
Pipelines seem stuck when running GitHub and Jira collection at the same time


hezyin:
## Describe the bug

Seems like lake has a problem with running multiple pipelines at the same time.

## To Reproduce

I encountered this issue when I built images from `686858a9c693371d23675462a6a3b4517440a589` on main.

1. Checkout to `686858a9c693371d23675462a6a3b4517440a589`
2. Run `docker-compose build` and `docker-compose up -d`
3. Setup GitHub and Jira data connection in Config UI
4. First create a GitHub pipeline to collect lake's own GitHub data
5. Then create a Jira pipeline (I used board 76 from our jira instance)
6. Observe that it's taking much longer than usual to collect data

<img width="2264" alt="Screen Shot 2022-04-17 at 5 38 57 PM" src="https://user-images.githubusercontent.com/2908155/163738335-50957f8c-3b20-43cd-804b-bb2859b0073d.png">
<img width="2274" alt="Screen Shot 2022-04-17 at 5 40 18 PM" src="https://user-images.githubusercontent.com/2908155/163738358-ae334b7a-11bc-4d2f-b67b-906c839bd49a.png">
<img width="2264" alt="Screen Shot 2022-04-17 at 5 40 27 PM" src="https://user-images.githubusercontent.com/2908155/163738362-209eec3e-6eaf-451e-aa15-00bcc0a1da47.png">

7. Try canceling the Jira pipeline and observe the cancellation doesn't work and the pipeline keeps running

## Expected behavior

I expect the GitHub pipeline to take about 25 mins and the Jira pipeline to take less than 3 mins. Also expects the pipeline canceling would work.

hezyin:
So it turns out my case is due to two separate issues:

1. The Jira stuck problem is fixed by https://github.com/merico-dev/lake/pull/1650
2. The Github stuck problem is because I used an invalid API token among all tokens.

Closing this issue now.

PR #1650, opened 2022-04-18
fix: jira stuck at collectChangelog when if Input is empty

klesh:
# Summary

Related to #1649

### Key Points


warren830 (approved):
LGTM

Issue #1683, opened 2022-04-18
API_TIMEOUT in .env missing uint


warren830:
## Describe the bug
[task service] task failed%!(EXTRA *fmt.wrapError=failed to create jira api client: failed to parse API_TIMEOUT: time: missing unit in duration "10")"
we don't have unit for API_TIMEOUT in .env.example. If users copy&paste the file to .env, they will fail to create api client

## Expected behavior
API_TIMEOUT should equal to 10s, not 10

PR #1686, opened 2022-04-18
fix(config file): add unit to API_TIMEOUT

warren830:
# Summary

Add unit to API_TIMEOUT in .env.example in root dir

relate to #1683


klesh (approved):
LGTM

Issue #1682, opened 2022-04-18
regression: unable to see error message on config-ui


klesh:
## Describe the bug

![image](https://user-images.githubusercontent.com/61080/163834732-2adfe139-2082-4937-aa5f-7e207c49afdb.png)

## To Reproduce
1. Go to github connection setting
2. Enter some invalid token
3. Can see error message promting which token was invalid
4. No can see error message on interface

## Expected behavior
Be able to see error message responsed from server

PR #1688, opened 2022-04-18
`fix`: use message prop from 400 response body

e2corporation:
### ⚠️ Config-UI / Data Integrations / Connection Test

- [ ] Use Error Message prop from 400 `response` `body`
- [ ] Fix invalid Custom Frequency validation rule

### Description
This PR uses the proper intended key for message error responses with regard to connection tests so that accurate information is displayed about the error. In specific cases such as GitHub multi-token, if one bad token is present the error message will contain details about the token error.

#1682

### Screenshots
Screen Shot 2022-04-18 at 12 02 07 PM


klesh (approved):
LGTM

Issue #1641, opened 2022-04-16
[Config UI] Jira issue type mapping: selected a single bug label, refreshed the page, and there were multiple bug labels


hezyin:
## Describe the bug

When configuring the Jira issue type mapping in config UI, some duplicate labels with the same name would appear after I refresh the page.

## To Reproduce

This is happening on the latest main (`9b4477fac`).

1. Run `docker-compose build` and `docker-compose up`
2. Visit config UI at localhost:4000
3. Successfully configure a Jira data source connection
4. Head on to select a label for issue type bug
<img width="1263" alt="Screen Shot 2022-04-15 at 4 54 31 PM" src="https://user-images.githubusercontent.com/2908155/163653766-68ad65e1-d633-4652-aaba-75896af1df6d.png">

5. Save the 
<img width="1180" alt="Screen Shot 2022-04-15 at 4 54 40 PM" src="https://user-images.githubusercontent.com/2908155/163653779-01ab4091-679d-4c9d-96aa-63b2ee61739e.png">
setting

6. Refresh the page and see the following
<img width="1164" alt="Screen Shot 2022-04-15 at 4 54 56 PM" src="https://user-images.githubusercontent.com/2908155/163653797-8ab129a2-492e-4538-b29b-c40db75c4a2a.png">

There are 4 bug labels in the issue type bug field, I'd expect only one.

PR #1694, opened 2022-04-19
`fix` `v0.10.0` jira provider - remove duplicate jira issue types from list

e2corporation:
### Config-UI / Data Integrations / JIRA / Configure Settings

- [x] Add a **Reducer** for de-duplicating **JIRA** IssueTypes fetched from `REST` `API`
- [x] **Test** Issue Types **Multi-tag Selector** List Render

### Description
[**V0.10.0** RELEASE TARGET] Ⓜ️ `main`
This PR removes duplicate **JIRA Issue Types** fetched from the `REST` `API` in cases where types of the same name such as "**Bug**" are unnecessarily duplicated within the response. This doesn't happen on all JIRA Instances, however in cases where the `Issuetype:name` property (ie `value` key) is repeated it will be removed from the list selection.

#1641

### Screenshots
Screen Shot 2022-04-19 at 10 18 13 AM
Screen Shot 2022-04-19 at 10 19 19 AM

### DIFF
--- a/config-ui/src/hooks/useJIRA.jsx
+++ b/config-ui/src/hooks/useJIRA.jsx
@@ -74,7 +74,9 @@ const useJIRA = ({ apiProxyPath, issuesEndpoint, fieldsEndpoint }) => {
   }
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


klesh (approved):
LGTM

PR #1399, opened 2022-03-28
fix: #1373

mindlesscloud:
# Summary

### Key Points

- [ ] This is a breaking change
- [ ] New or existing documentation is updated

### Description
Describe what this PR does, and aims to solve in a few sentences.

### Does this close any open issues?
#1373 

### Current Behavior
Describe the current behaviour of this issue, if relevant.

### New Behavior
Describe the new behaviour updated in this issue, if relevant.

### Screenshots
image

### Other Information
Any other information that is important to this PR.

PR #1695, opened 2022-04-19
fix(api_collector): fix dead lock

### LINKED ISSUE #1693: api_collector dead lock
## Describe the bug
When we are running two plugins at the same time, one of them might get block.
This is caused by dead lock. We call scheduler.submit inside scheduler.submit in api_async_client.DoAsync

## Expected behavior
System should report error or continue working

warren830:
close #1693

# Summary

Add a subpool for scheduler to hold retry requests
And to avoid rate limit, for main pool submit, we will sleep for a second if there are still some running goroutines in the subpool. 

### Key Points

### Description

close #1693 

### Current Behavior

### New Behavior

### Screenshots
image

image

### Other Information

### DIFF
--- a/plugins/helper/api_async_client.go
+++ b/plugins/helper/api_async_client.go
@@ -73,7 +73,7 @@ func CreateAsyncApiClient(
 		requests,
 		duration,
 	)
-	scheduler, err := NewWorkerScheduler(numOfWorkers, requests, duration, taskCtx.GetContext())
+	scheduler, err := NewWorkerScheduler(numOfWorkers, requests, duration, taskCtx.GetContext(), retry)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create scheduler: %w", err)
 	}
@@ -97,7 +97,8 @@ func (apiClient *ApiAsyncClient) DoAsync(
 	handler ApiAsyncCallback,
 	retry int,
 ) error {
-	return apiClient.scheduler.Submit(func() error {
+	var subFunc func() error
+	subFunc = func() error {
 		var err error
 		var res *http.Response
 		var body []byte
@@ -111,7 +112,8 @@ func (apiClient *ApiAsyncClient) DoAsync(
 		if err != nil {
 			if retry < apiClient.maxRetry && err != context.Canceled {
 				apiClient.logError("retry #%d for %s", retry, err.Error())
-				return apiClient.DoAsync(method, path, query, body, header, handler, retry+1)
+				retry++
+				return apiClient.scheduler.Submit(subFunc, apiClient.scheduler.subPool)
 			}
 		}
 		if err == nil {
@@ -122,7 +124,8 @@ func (apiClient *ApiAsyncClient) DoAsync(
 		// it is important to let handler have a chance to handle error, or it can hang indefinitely
 		// when error occurs
 		return handler(res, err)
-	})
+	}
+	return apiClient.scheduler.Submit(subFunc)
 }
 
 // Enqueue an api get request, the request may be sent sometime in future in parallel with other api requests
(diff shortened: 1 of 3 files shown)


plugins/helper/api_async_client.go
```diff
@@ -111,7 +112,8 @@ func (apiClient *ApiAsyncClient) DoAsync(
 		if err != nil {
 			if retry < apiClient.maxRetry && err != context.Canceled {
 				apiClient.logError("retry #%d for %s", retry, err.Error())
-				return apiClient.DoAsync(method, path, query, body, header, handler, retry+1)
+				retry++
+				return apiClient.scheduler.Submit(subFunc, apiClient.scheduler.subPool)
```
liangjingyang, line 116: What if mainPool.workerNum * maxRetry > subPool.workerNum? Will the worker of subPool run out?

klesh: I agree with @liangjingyang , this is effectively increasing the size of the pool, which can reduce the probability of dead locking, but not solving the root cause.

I think a better way to solve this is to use the same way to #1399

warren830: I set the subpool's size to four times of mainpool. So it will not run out of workers

warren830: Maybe I should set subpool's size to maxtry * size of mainpool

plugins/helper/worker_scheduler.go
```diff
@@ -34,6 +35,12 @@ func NewWorkerScheduler(workerNum int, maxWork int, maxWorkDuration time.Duratio
 		workerErrors = append(*pWorkerErrors, i.(error))
 		pWorkerErrors = &workerErrors
 	}))
+	subPool, err := ants.NewPool(workerNum*5, ants.WithPanicHandler(func(i interface{}) {
+		mux.Lock()
```
warren830, line 39: we can set the first param of newPool to workerNum * maxtry

plugins/helper/worker_scheduler.go
```diff
@@ -78,6 +92,11 @@ func (s *WorkerScheduler) Submit(task func() error) error {
 				panic(fmt.Errorf("%s\n%s", err, cf))
 			}
 		}()
+		if pool == nil {
+			for s.subPool.Running() != 0 {
+				time.Sleep(1 * time.Second)
+			}
+		}
```
warren830, line 99: Here, if there are still running workers in subpool, we can set a ticker here

plugins/helper/worker_scheduler.go
warren830: And I tried to let network disconnect until finish all retry, and it works

liangjingyang:
The key point is how to retry a task when error occurred, we shouldn't submit a new task within an uncompleted task recursively. 

There are 2 solutions to achieve this:

1. Let the ApiCollector control the retry, but I think this needs more changes than the second one. It require every handler has a channel to send message from task to ApiCollector, but seems only the handler inside `stepFetch` has a channel currently.
2. Retry by a normal loop inside the task instead of submitting a new task, maybe:
```
var err error
res, err = apiClient.Do(method, path, query, body, header)
for retry < apiClient.maxRetry - 1 && err != nil {
    retry = retry + 1
    res, err = apiClient.Do(method, path, query, body, header)
    ......
}
```

 I think the subpool solution has 2 issues:
1. Introduce a new pool which increases the complexity.
2. The `Sleep` could slow down the data collection when error occurred.

hezyin:
Submitting and waiting for a task within a task is a common cause of deadlocks. One good practice is to only submit tasks from one thread and not expose the pool itself to tasks executed by the pool. I'm leaning towards solution 2 in @liangjingyang's comment.

Issue #1606, opened 2022-04-15
Config UI - Jira Add Connection Input Warning


yumengwang03:
## Describe the bug
When adding a connection for Jira, the input fields are highlighted with warning, which can be misleading because the user hasn't entered anything.

## To Reproduce
1. Go to 'Data Integrations'
2. Click on 'Jira'
3. Click on 'Add Connection'

## Expected behavior
Text input fields should not be shown with warning.

## Screenshots
<img width="1315" alt="Screen Shot 2022-04-15 at 3 23 25 PM" src="https://user-images.githubusercontent.com/9535430/163538984-c5613b37-23c1-4510-a6b1-851fc543051d.png">

e2corporation:
@yumengwang03 @Startrekzky This is not a bug, the product/design team previously requested that errors be moved closer to the form inputs. This is standard inline form validation that lets the user know that required fields are needed.

PR #1697, opened 2022-04-20
`feat` `v0.10.0` connection inputs: add validate-on-focus support

e2corporation:
### Config-UI / Data Integrations / Connections

- [x] Add `validateOnFocus` Feature support to InputValidation Helper Component
- [x] Test Connection Form validation behavior

### Description
[**V0.10.0** RELEASE TARGET] Ⓜ️ `main`
This PR adds "**Validate On-Focus**" support to the primary **Connection Form Inputs** for JIRA and other Data Providers as well. In the case of JIRA, a multi-source provider, when Adding a New connection this will allow the input fields to appear with less "Severity" by default, only when focus is given to the primary Input element will the error states be fully activated.

#1606

### Screenshots
Screen Shot 2022-04-19 at 8 19 10 PM
Screen Shot 2022-04-19 at 8 45 38 PM
Screen Shot 2022-04-19 at 8 30 01 PM

### DIFF
--- a/config-ui/src/components/validation/InputValidationError.jsx
+++ b/config-ui/src/components/validation/InputValidationError.jsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState, useCallback } from 'react'
 import {
   Colors,
   Icon,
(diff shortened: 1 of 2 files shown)


klesh (approved):
LGTM

PR #1724, opened 2022-04-24
feat(framework): print log to both file and console

### LINKED ISSUE #1716: Print log to both file and stdOut
## Description
Can we save the log information to both file and stdOut?

When log.go is initing, we create a dir named "logs"
In the func of default_logger.go, we set log output to multiwriter which contains both file and stdout

For each subtask, we create a log file. The file name can be set to `plugin_task.log`

warren830:
# Summary

Add a logger pool to DefaultLogger, whenever we need a nested logger, we use file name to check if the pool has that logger, if so, we just return it to create a  new default logger, otherwise, we create a new logger

### Key Points

### Description

closes #1716

### Current Behavior

### New Behavior

### Screenshots
image
image
image
image

### Other Information

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -1,11 +1,13 @@
 package logger
 
 import (
+	"fmt"
 	"github.com/merico-dev/lake/config"
 	"github.com/merico-dev/lake/plugins/core"
 	"github.com/merico-dev/lake/plugins/helper"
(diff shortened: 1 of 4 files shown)


warren830:
can not be closed

mindlesscloud (approved):

PR #1726, opened 2022-04-24
feat: prompt user of potential correct jira endpoint url `TestConnect…

klesh:
# Summary

Before:
Snipaste_2022-04-24_16-15-02

After:
Snipaste_2022-04-24_17-03-31

### Key Points


Startrekzky (approved):
LGTM

PR #1740, opened 2022-04-25
1733 make sure every plugin sort data source records by created date to avoid data lost during collection

### LINKED ISSUE #1733: Make sure every plugin sort data source records by created date to avoid data lost during collection
## Description

When devlake is fetching paginated records from API, and someone is making changes onto data-source, like creating or deleting records that we are fetching, some records might be missed out due to records might shifting between pages. So, the most sensible strategy is to order records by `creation date` in ascending order.

However, this is easier said than done, it is hard to detect whether or not a `collector` adopt this strategy with some sort of automatic mechanism, until we figure out how to enforce this strategy, we shall do it manually.

Go through every single collector of every single plugin, make sure they adopt the `order by created_date` strategy
image

No

warren830:
# Summary

Modified feishu and github to make sure they collect data from the earliest one.
Gitlab collect data from the earliest one by default

closes #1733 

### Current Behavior

### New Behavior

### Screenshots

### Other Information


klesh:
As we discussed, this is unnecessary, close it for now, feel free to reopen

plugins/helper/iterator.go
```diff
@@ -46,40 +46,41 @@ func (c *CursorIterator) Close() error {
...
+func (c *DateInterator) Close() error{
 	return nil
 }
 
 func NewDateInterator(days int) (*DateInterator, error) {
 	endTime := time.Now().Truncate(24 * time.Hour)
 	return &DateInterator{
-		startTime: endTime,
-		endTime:   endTime.AddDate(0, 0, 1),
+		startTime: endTime.AddDate(0, 0, -days-1),
```
klesh, line 82: @abeizn Please confirm this is correct.

abeizn (approved):

Issue #1544, opened 2022-04-13
dbt plugin support postgres (duedate 28th April)


klesh:
## Description
Requirement from EE, pg support could help EE reduce the cost of maintenance.

## Describe the solution you'd like
Add `postgres` support to dbt plugin

## Has the Feature been Requested Before?
No

PR #1748, opened 2022-04-26
feat: dbt plugin support postgres

abeizn:
relate to #1544 

mysql final return result:
image

postgres final return result:
image


mindlesscloud (approved):
LGTM

Issue #1747, opened 2022-04-26
Add the version number to the config-UI


Startrekzky:
## Why
It helps:
- when users are reporting a bug and trying to provide version info.
- when users have deployed multiple instances and try to figure out the version of them.

## Describe the solution you'd like
- Add the version number to config-UI. Users can see the version number no matter what pages they are visiting.
- The value should be up-to-date. Is there a way to not forget to update version No. before a release.

klesh:
Backend part is ready

tanninghan:
@e2corporation Prototype is ready, with the name issue#1747  https://www.figma.com/file/qpUZN3dJqnOcXxPZs36wTj/Config-UI-update?node-id=65%3A1433

PR #1749, opened 2022-04-26
feat: add version information based on git tag/sha

klesh:
# Summary

Backend part of issues #1747

1. Version is generated based on tag/commit sha of current commit
2. devlake server/worker would print `Version` on start up
3. devlake server provides `/version` to return version information

### Key Points

### Description
Snipaste_2022-04-26_15-29-43

Snipaste_2022-04-26_15-25-56

Snipaste_2022-04-26_15-27-58

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,20 @@
 # https://stackoverflow.com/questions/920413/make-error-missing-separator
 # https://tutorialedge.net/golang/makefiles-for-go-developers/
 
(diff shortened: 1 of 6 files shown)


e2corporation:
@klesh Under development-mode I noticed that the endpoint just returns "dev", wouldn't it be more appropriate to return the 8-digit commit hash instead? Not a major issue.

hezyin (approved):
LGTM

PR #1769, opened 2022-04-28
feat: config ui - add devlake api version tag to sidebar

e2corporation:
### Config-UI / Sidebar / API Version Tag

- [x] Add Version Tag to sidebar (provided by `/version` endpoint)

### Description
This PR adds a **Version Tag** to Config-UI's Sidebar so that users are aware of the current release version and commit hash being served.

#1747

### Screenshots
Screen Shot 2022-04-28 at 1 12 06 AM

### DIFF
--- a/config-ui/src/components/Sidebar.jsx
+++ b/config-ui/src/components/Sidebar.jsx
@@ -29,7 +29,7 @@ const Sidebar = () => {
           console.log('>>> API VERSION ERROR...', e)
           setVersionTag('dev+error')
         })
-        setVersionTag(res?.data ? res.data?.version : 'dev+error')
+        setVersionTag(res?.data ? res.data?.version : '')
       } catch (e) {
         setVersionTag('dev+error')
       }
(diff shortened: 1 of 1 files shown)


e2corporation:
Depends on PR #1749

config-ui/src/components/Sidebar.jsx
```diff
@@ -41,8 +59,10 @@ const Sidebar = () => {
         <sup style={{ fontSize: '9px', color: '#cccccc', marginLeft: '-30px' }}>DEV</sup>LAKE
       </h3>
       <SidebarMenu menu={menu} />
-
-      <span className='copyright-tag'><strong>Apache 2.0 License</strong><br />&copy; 2021 Merico</span>
+      <span className='copyright-tag'>
+        <span className='version-tag'>{versionTag || 'dev+unknown'}</span><br />
```
hezyin, line 63: Same as above

hezyin (changes requested):
I suggest that we display an empty string as default when we can't get the version info from the API, otherwise this PR looks good.

hezyin (approved):
LGTM

e2corporation:
@hezyin This is simply a fallback value to indicate that the `/version` endpoint has a failure and it should alert the developer of a problem, if BE is operating normally only the tag value would show. This can be change to empty string if needed.

Issue #1662, opened 2022-04-18
Notify user of having invalid API tokens when saving GitHub data connection


hezyin:
## Description

If I try to save GitHub data connection with some invalid API tokens, DevLake saves the connection successfully and without any notification for the invalid API tokens.

<img width="2540" alt="Screen Shot 2022-04-18 at 12 52 43 AM" src="https://user-images.githubusercontent.com/2908155/163776232-d75bba80-22cb-47a0-bd2c-4cafd90199b8.png">

## Describe the solution you'd like

It's ok for DevLake to save the connection successfully, but maybe giving notification of having invalid API tokens would be helpful, especially considering invalid GitHub API tokens might result in stuck pipelines.

PR #1777, opened 2022-04-30
feat: connections - test connection on save for multi-token providers

e2corporation:
### Config-UI / Data Integrations / Configure Connection

- [x] Test Connection `onSave` for Multi-token Providers like **GitHub** and **JIRA**
- [x] `fix` set version tag to empty str if unavailable

### Description
This PR will perform a Connection test when a Connection Source is modified, but only for token based providers such as GitHub and JIRA. If a multi-token string is detected in the payload, a connection test will be performed afterwards to notify the user if there is a problem with one of the tokens.

#1662


klesh (approved):
LGTM

Issue #1385, opened 2022-03-24
Prompt users to set multiple tokens for GitHub configuration


tanninghan:
## Description
Due to GitHub API rate limit, it will take a lot of time to run data collection if only one token is configured.
![image](https://user-images.githubusercontent.com/102040651/159972988-0727c403-10e1-4e4b-ab33-b14b9d00a6e7.png)

## Describe the solution you'd like
Prompt users to set multiple tokens and tell them how to get these tokens.

## Design of the solution
Show an `Add Token` button and perhaps 5 input boxes on the `GitHub Add Connection` page, so that users can provide multiple tokens at once to speed up data collection.

tanninghan:
Prototypes of GitHub configuration
![GitHub Add Connection](https://user-images.githubusercontent.com/102040651/160452934-cc979c60-f993-4654-990d-e9e7a3bc670c.jpg)
![Manage Github Settings](https://user-images.githubusercontent.com/102040651/160452938-3ad4321a-03cd-486c-b278-750f8e098610.jpg)

tanninghan:
Prototype of GitHub configuration
![GitHub configuration](https://user-images.githubusercontent.com/102040651/160614521-99029c5b-9bd5-4957-a3b5-ac307acac4c1.jpg)
![Manage Github Settings](https://user-images.githubusercontent.com/102040651/160614528-fd7e7baf-0341-49a7-a5e2-41ef0ea33286.jpg)

e2corporation:
@Startrekzky Moving this ticket for enhancing multi-token input to `v0.10.0`

PR #1785, opened 2022-05-04
fix: use multi-line textarea on github token field

e2corporation:
### Config-UI / Data Integrations / GitHub Connection - **Token Field**

- [x] Use Multi-line **TextArea** Input for **GitHub Token** Connection Field
- [x] Move Tooltip to separate `info` icon next to "Auth Token" label

### Description
This PR Updates the Token form field for a GitHub Data Source Connection. A multi-line TextArea will be used for GitHub provider specifically so multiple tokens can be easier managed.

#1385

### Screenshots
Screen Shot 2022-05-04 at 2 24 11 PM
Screen Shot 2022-05-04 at 2 32 24 PM
Screen Shot 2022-05-06 at 8 56 56 PM

### DIFF
--- a/config-ui/src/data/Providers.js
+++ b/config-ui/src/data/Providers.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Tooltip } from '@blueprintjs/core'
+import { Tooltip, Icon } from '@blueprintjs/core'
 import { ReactComponent as GitlabProviderIcon } from '@/images/integrations/gitlab.svg'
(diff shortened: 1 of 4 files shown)


klesh:
Nice, even better if we can had line number

e2corporation:
> Nice, even better if we can had line number

@klesh This type of input/textarea doesn't support line numbers, also this isn't that practical for this type of field. Ideally it should be a multi-tag input field like we do with other inputs, but textarea is more suitable.

klesh (approved):
LGTM

PR #1811, opened 2022-05-07
feat: add icon_url for table issues

### LINKED ISSUE #1774: Add `icon` for table `issues`
## Description
When we list issues, an icon is useful for users to check type and source of the issue. eg: It is a bug from jira.  
We Suggest adding an `icon` column for `issues` entity, using original type icon of this issue.

## Additional context
image  
Or  
image

abeizn:
fix #1774

image
image
image


klesh:
@narrowizard

mindlesscloud (approved):

PR #1814, opened 2022-05-08
feat(github): change invalid `token` index

### LINKED ISSUE #1800: github test connection reporting invalid token with its `index`
## Description

The invalid token message is referring failed token by its `index` which start from `0`, it is not very user friendly.
image

We should use `number` which start from `1` instead.

No

warren830:
# Summary

add 1 to `token` index, so that it will show more understandable index in log

closes #1800

### Screenshots
image

### Other Information


abeizn (approved):

abeizn:
LGTM

PR #1877, opened 2022-05-14
fix: refdiff incompatible with postgres 

### LINKED ISSUE #1842: Postgres Support: some SQL include functions that are only available in Postgres
## Describe the bug
The function `GROUP_CONCAT` is not available in Postgres

## Screenshots
image
image

### LINKED ISSUE #1868: jira use empty connection data without any error
## Describe the bug
image
There are no Connection with ConnectionId 1.
So the jira get an empty struct
But here don't sent an error
The empty sturct are using continue.
It is unsafe.

## To Reproduce
1. Debug plugins/jira/jira.go with param “-s 1 -b 1” without the database data of ConnectionID 1.
2. break after db.Find and look at the err and connection.

### LINKED ISSUE #1873: refdiff (calculatePrCherryPick) is incompatible with postgres
## Describe the bug
ERROR: unterminated quoted identifier at or near """ (SQLSTATE 42601)
and
ERROR: invalid value, should be pointer to struct or slice

## To Reproduce
run github plugin (apache/incubator-devlake)
run gitextrator plugin(apache/incubator-devlake)
run refdiff (calculatePrCherryPick)

## Screenshots
image
image

abeizn:
fix #1868: refs_issues_diffs has no primary key 
fix #1873:  unterminated quoted identifier at or near """  and  invalid value, should be pointer to struct or slice 
fix #1842: some SQL include functions that are only available in Postgres

### DIFF
--- a/models/domainlayer/crossdomain/refs_issues_diff.go
+++ b/models/domainlayer/crossdomain/refs_issues_diff.go
@@ -3,11 +3,11 @@ package crossdomain
 import "github.com/merico-dev/lake/models/common"
 
 type RefsIssuesDiffs struct {
-	NewRefId        string `gorm:"type:varchar(255)"`
-	OldRefId        string `gorm:"type:varchar(255)"`
+	NewRefId        string `gorm:"primaryKey;type:varchar(255)"`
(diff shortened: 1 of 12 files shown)


klesh:
Covered by #1890

PR #1892, opened 2022-05-17
fix(lint): increase the logging detail in golangci-lint action

warren830:
# Summary

increase the logging detail in golangci-lint action


klesh (approved):
LGTM

PR #1907, opened 2022-05-18
fix: some panels of grafana can not work

### LINKED ISSUE #1905: grafana GithubReleaseQualityAndContributionAnalysis some panels can not work
## Describe the bug
bug1: when user collects muti-repos, and select all repos mode, then some papels can not show, the reason is some sql should be in ($repo_id) instead of = ($repo_id)

bug2: when user selected all repos mode, the value is equal to the last repo instead of the sum of all repos.

## To Reproduce
1. collect muti-repo
2. select all repo mode
3. then you can see some papel that show Bar charts requires a string field.

## Screenshots
bug1:
image

bug2:
image
image
image

abeizn:
### Description
change  `= ($repo_id)` to `in ($repo_id) `

fix #1905 

### Screenshots
image


klesh (approved):
LGTM

Issue #1911, opened 2022-05-18
Rename module name to `github.com/apache/incubator-devlake`


klesh:
## Description
Subtask of #1909 

## Describe the solution you'd like
1. Rename `module` name from `go.mod` file
2. Replace `github.com/merico-dev/lake` with ` github.com/apache/incubator-lake` of `import` section for all our source files

## Has the Feature been Requested Before?
No

likyh:
Finished

Issue #1910, opened 2022-05-18
All source code files must begin with the Apache License Header


klesh:
## Description
Subtask of #1909 

## Describe the solution you'd like
Add apache header to every single source code files
header can be found from all apache project, like https://github.com/apache/servicecomb-java-chassis/blob/master/core/src/main/java/org/apache/servicecomb/core/bootstrap/SCBBootstrap.java

## Has the Feature been Requested Before?
No

likyh:
ok, i'll add it

likyh:
ok, i'll add it

likyh:
finished

PR #1940, opened 2022-05-19
feat: add go header and other golint check

likyh:
# Summary

Add apache header and replace package name to apache package name;
Fix some bug for golangci-lint.

### Key Points

#1910 #1911

### Screenshots
image


klesh (approved):
LGTM

Issue #1939, opened 2022-05-19
grafana GithubReleaseQualityAndContributionAnalysis add some severity distribution dashboard and add all repos mode


abeizn:
## Description
1. The grafana GithubReleaseQualityAndContributionAnalysis plan to add some severity distribution so that users understand the severity distribution of bug in different tag.

2. when user use dashboard and select all repo, the function does not work.
![image](https://user-images.githubusercontent.com/101256042/169238731-59ef168a-8249-4c58-ad8e-930afb9098cf.png)
![image](https://user-images.githubusercontent.com/101256042/169238902-f0907e36-0a51-40b4-b7a9-0956e84d5bce.png)
![image](https://user-images.githubusercontent.com/101256042/169238952-4b0ba1ac-1b2f-48fa-9a6b-fc564df264da.png)

PR #1945, opened 2022-05-19
feat: grafana add some severity distribution and adapt muti-repos mode

abeizn:
### Description
fix: #1939 

### Screenshots
1. grafana add some severity distribution
image

3. grafana add  muti-repos mode and the final value is right
image


warren830 (approved):
LGTM

PR #1974, opened 2022-05-23
fix(runner): swallow the error form exec.Command.Run

### LINKED ISSUE #1973: exec.Command always return an error
## Describe the bug
In `DirectRun`, if `exec.Command.Run()` starts but does not complete successfully, the error is of
type *ExitError. Other error types may be returned for other situations.

For stty, in linux, it use -arg `-F`, but it always return error if -arg is `-F` in macOS

## Expected behavior
We cannot panic this error, we can panic for other types of error but not including `*ExitError`

warren830:
# Summary
swallow the error from exec.Command.Run()

closes #1973

### Screenshots

### Other Information

### DIFF
--- a/runner/directrun.go
+++ b/runner/directrun.go
@@ -64,10 +64,8 @@ func DirectRun(cmd *cobra.Command, args []string, pluginTask core.PluginTask, op
 		panic(err)
 	}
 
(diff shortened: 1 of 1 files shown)


mindlesscloud (approved):
LGTM

Issue #1983, opened 2022-05-24
 Refdiff to support another set of parameters tag regex and tag limit, and calculate diffs based on these 2 parameters


klesh:
## Description
Refdiff currently accept only `pairs` and calculate diffs for them, this is not very flexible, because most of the time, user would like to know diffs of latest releases. It's no point to calculate same set of pairs over and over again, especially with `pipelines` that triggered by `blueprint`

## Describe the solution you'd like
Beside `pairs`, refdiff can accept `tagsPattern` and `tagsLimit`, maybe evens `tagsOrder`(alphabetically , semver). and refdiff would filter `tags` by these parameters, sort them, take last `tagsLimit` tags, and calculates diffs for them.

## Has the Feature been Requested Before?
No

mappjzc:
Assignee to me.

Issue #1981, opened 2022-05-24
Multi-data connections support for AE


klesh:
## Description
To be filled

## Has the Feature been Requested Before?
Please link any other

klesh:
@mappjzc

Issue #1980, opened 2022-05-24
Multi-data connections support for Github


klesh:
## Description
To be filled

## Has the Feature been Requested Before?
Please link any other

klesh:
@likyh

likyh:
ok, I'll try to finish it.

likyh:
Finished

Issue #1979, opened 2022-05-24
Config UI for v1.0 - stage 1


klesh:
## Description
Stage 1 of #1862 

## Describe the solution you'd like

**NOTE: every `connection` should have `ratelimit` field**

### Backend tasks
- #1412
- #1978 
- #636 
- #1980
- #1981
- #2262
- #1983
- Gitlab to accept `transformation rules` from `task.options`
- #1986
- Feishu to accept `transofmration rules ` from `task.options`
- Blueprint Api design upgrade
  1. #1987
  2. #1982

Multi Connection Checklist:
- Replace `GetDb` with `GetDal`
- Configuration like `PrType` `IssueTypeBug` are now grouped as `TransformationRules` and will be passed in `task.options`, no need  to store them with `connection` anymore. and jira connection should be refactor and remove those settings as well.
- Remove `Concurrency` parameter
- One PR for one `entity`
- Write `e2e-test` for each `entity`
- Write `godoc` for every public method/function/struct
- Write `unit-test` for public util function if any
- Close the issue on last PR

### Frontend tasks (to be filled by @e2corporation )
- #1951 

## Has the Feature been Requested Before?
No

e2corporation:
@klesh can you please provide me the latest expected JSON configuration for Blueprint Create / `POST`? I need to make some changes to the request structure so that we send `boards` and `projects` as an array of **Board IDs** and **Repositories**. Backend can then split the string to extract `Owner` and `RepositoryName`.


# Requested Structure Updates
```
{
    "version": "1.0",
    "connections": [
        {
            "plugin": "github",
            "connectionId": 123,
            "scope": [
                {
                    "projects": ["merico-dev/lake", "merico-dev/stream"]
                    "transformation": {
                        "issueTypeBug": "^(bug|failure|error)$"
                    },
                    "entities": [
                        "code",
                        "crossdomain",
                        "devops",
                        "ticket",
                        "user"
                    ]
                }
            ]
        },
        {
            "plugin": "jira",
            "connectionId": 321,
            "scope": [
                {
                    "boards": [123],
                    "options": {
                        "since": "2022-1-2"
                    },
                    "transformation": {
                        "issueTypeBug": "??"
                    },
                    "entities": []
                }
            ]
        }
    ]
}
```


# Current Structure proposed on #1982
Dispatching `owner` and `repo` as 2 separate fields is not ideal, instead we should send an array of `projects`
```
{
    "version": "1.0",
    "connections": [
        {
[...]

PR #1994, opened 2022-05-24
refactor(jira): extract common tools

### LINKED ISSUE #2003: Extract common tools for data connections
## What and why to refactor
Most of our data plugins use a serial of common utils.

## Related issues
Please link any other

warren830:
# Summary

In this pr, we extract the following common tools for plugin connection:
`RefreshAndSaveConnection`
`FindConnectionByInput`
`GetConnectionIdByInputParam`
`DecryptConnection`

Then, when we need to do the similar operations for other plugins, we can call these funcs rather than creating a duplicate one.
And this pr rebase on #2002, so you can see some frontend changes here.
The reason I rebase this on #2002 is we need to verify if the common tools can fit to username/password pattern

fixes #2003 
also is part of #1979 
this is the base of multi connection pr:
- #1978 
- #636 
- #1980
- #1981
- #1983

### Screenshots
db migration for basic_auth_encoded to username/password
image
image

### Other Information


plugins/helper/common_models.go
```diff
@@ -0,0 +1,24 @@
...
+	Name string `gorm:"type:varchar(100);uniqueIndex" json:"name" validate:"required"`
+	common.Model
+}
+
+type BasicAuth struct {
+	Username string `mapstructure:"username" validate:"required" json:"username"`
+	Password string `mapstructure:"password" validate:"required" json:"password" encrypt:"yes"`
+}
+
+type Token struct {
```
klesh, line 15: Token is a bit vague, I would suggest to use `AccessToken` instead

plugins/helper/connection_util.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/apache/incubator-devlake/plugins/core"
+	"github.com/go-playground/validator/v10"
+	"github.com/mitchellh/mapstructure"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+	"reflect"
+	"strconv"
+)
+
+func RefreshAndSaveConnection(connection interface{}, data map[string]interface{}, db *gorm.DB) error {
```
klesh, line 15: Please add `godoc` to explain what would this method do, including its PATCH and VALIDATION nature

plugins/helper/connection_util.go
```diff
@@ -0,0 +1,160 @@
...
+		panic("entityPtr is not a pointer")
+	}
+
+	encryptCode, err := GetEncKey()
+	if err != nil {
+		return err
+	}
+	dataType := reflect.Indirect(dataVal).Type()
+	fieldName := getEncryptField(dataType, "encrypt")
+	fmt.Printf("fieldName : %s\n", fieldName)
```
klesh, line 43: I think this line should be removed.

plugins/jira/api/connection.go
```diff
@@ -52,11 +49,12 @@ func TestConnection(input *core.ApiResourceInput) (*core.ApiResourceOutput, erro
 	if err != nil {
 		return nil, err
 	}
+
 	// test connection
 	apiClient, err := helper.NewApiClient(
 		connection.Endpoint,
 		map[string]string{
-			"Authorization": fmt.Sprintf("Basic %v", connection.Auth),
+			"Authorization": "",
```
klesh, line 57: How does this work? I think this effectively makes all user/pass combination pass the test 😂

plugins/jira/api/connection.go
```diff
@@ -309,12 +203,12 @@ func ListConnections(input *core.ApiResourceInput) (*core.ApiResourceOutput, err
...
 		return nil, err
 	}
-	encKey, err := getEncKey()
+	encKey, err := helper.GetEncKey()
 	if err != nil {
 		return nil, err
 	}
 	for i := range jiraConnections {
-		jiraConnections[i].BasicAuthEncoded, err = core.Decrypt(encKey, jiraConnections[i].BasicAuthEncoded)
+		jiraConnections[i].Password, err = core.Decrypt(encKey, jiraConnections[i].Password)
```
klesh, line 211: Since we have already let helper encrypt fields for use, we should ask helper to decrypt them as well

plugins/helper/common_models.go
```diff
@@ -0,0 +1,24 @@
+package helper
```
klesh, line 1: The filename is vague, I suggest that we name it `connection.go` and merge `connection_util.go` altogether since they are highly coupling

plugins/jira/tasks/api_client.go
```diff
@@ -29,11 +30,13 @@ import (
 func NewJiraApiClient(taskCtx core.TaskContext, connection *models.JiraConnection) (*helper.ApiAsyncClient, error) {
 	// load configuration
 	encKey := taskCtx.GetConfig(core.EncodeKeyEnvStr)
-	auth, err := core.Decrypt(encKey, connection.BasicAuthEncoded)
+	decodedPassword, err := core.Decrypt(encKey, connection.Password)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to decrypt Auth Token: %w", err)
 	}
 
+	auth := utils.GetEncodedToken(connection.Username, decodedPassword)
```
klesh, line 38: It would make sense to let `BasicAuth` struct implement a method to return `encodeToken`

plugins/jira/api/connection.go
warren830: I remember jira does not require this to do the test. Will check again

warren830: my bad, already fixed

plugins/jira/api/connection.go
```diff
@@ -309,12 +203,12 @@ func ListConnections(input *core.ApiResourceInput) (*core.ApiResourceOutput, err
 	if err != nil {
 		return nil, err
 	}
-	encKey, err := getEncKey()
+	encKey, err := helper.GetEncKey()
 	if err != nil {
 		return nil, err
 	}
 	for i := range jiraConnections {
```
klesh, line 210: Looks like this logic should be moved to helper function

plugins/jira/api/proxy.go
```diff
@@ -48,7 +49,14 @@ func Proxy(input *core.ApiResourceInput) (*core.ApiResourceOutput, error) {
 		return nil, err
 	}
 	encKey := cfg.GetString(core.EncodeKeyEnvStr)
-	basicAuth, err := core.Decrypt(encKey, jiraConnection.BasicAuthEncoded)
+
+	decodedPassword, err := core.Decrypt(encKey, jiraConnection.Password)
```
klesh, line 53: Why not utilize `FindAuthConnectionByInput` ? and why it is called `FindAuthConnectionByInput`? What doesn't `Auth` stand for?

plugins/jira/models/connection.go
```diff
@@ -28,9 +28,9 @@ type EpicResponse struct {
 }
 
 type TestConnectionRequest struct {
-	Endpoint string `json:"endpoint"`
-	Auth     string `json:"auth"`
-	Proxy    string `json:"proxy"`
+	Endpoint         string `json:"endpoint"`
+	Proxy            string `json:"proxy"`
+	helper.BasicAuth `mapstructure:",squash" encrypt:"yes"`
```
klesh, line 33: Why do we need encryption for TestConnection when we don't need to store them into database?

plugins/jira/models/connection.go
```diff
@@ -40,15 +40,11 @@ type BoardResponse struct {
 }
 
 type JiraConnection struct {
-	common.Model
-	Name                       string `gorm:"type:varchar(100);uniqueIndex" json:"name" validate:"required"`
-	Endpoint                   string `json:"endpoint" validate:"required"`
-	BasicAuthEncoded           string `json:"basicAuthEncoded" validate:"required"`
+	helper.RestConnection      `mapstructure:",squash"`
+	helper.BasicAuth           `mapstructure:",squash" encrypt:"yes"`
```
klesh, line 44: Does `encrypt:"yes"` mean we gonna encrypt the whole struct?

plugins/jira/models/migrationscripts/updateSchemas20220524.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if db.Migrator().HasColumn(&JiraConnection20220505{}, "basic_auth_encoded") {
+		err = db.Migrator().DropColumn(&JiraConnection20220505{}, "basic_auth_encoded")
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
klesh, line 63: I think we have to decode the `basic_auth_encoed` and spread them into username/password accordingly before we drop them.

klesh:
The description doesn't seem enough to justify the PR, for those who didn't attend our discussion will not understand it a bit.
Please try to describe it better, so anyone can understand why this is needed.
i.e., you could link the epic issue we are working on, and point out for all plugins would need this logic, instead of repeating them many times, we should extract them into one place, etc...

klesh:
And we need unit tests for helper functions from `connection.go`
Remember to provide `godoc` for all Public functions/methods

plugins/helper/connection.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func saveToDb(connection interface{}, db *gorm.DB) error {
+	dataVal := reflect.ValueOf(connection)
+	if dataVal.Kind() != reflect.Ptr {
+		panic("entityPtr is not a pointer")
+	}
+
+	dataType := reflect.Indirect(dataVal).Type()
+	fieldName := getEncryptField(dataType, "encrypt")
```
klesh, line 70: Seems like `getEncryptField` is to find the name of the first matched field by tag.
I would suggest we name it `firstFieldNameWithTag`, or keep the naming but remove the second parameter.
Otherwise, one could use this function to query a field name in a completely different semantic.
i.e., I could use `getEncryptField(foo, "bar")` to find the field name with the tag `bar`, and it would not necessarily be a **encrypted field** so to speak.

plugins/helper/connection.go
```diff
@@ -0,0 +1,217 @@
...
+		panic("connection is not a pointer")
+	}
+	if len(fieldName) == 0 {
+		dataType := reflect.Indirect(dataVal).Type()
+		fieldName = getEncryptField(dataType, "encrypt")
+	}
+	return doDecrypt(dataVal, fieldName)
+}
+
+func doDecrypt(dataVal reflect.Value, fieldName string) error {
```
klesh, line 187: Can we name it `decryptField` / `encryptField`? Two verbs joined together doesn't seem right.

plugins/helper/connection.go
```diff
@@ -0,0 +1,217 @@
...
+			if field.Tag.Get(tag) == "yes" {
+				fieldName = field.Name
+			}
+		}
+	}
+	return fieldName
+}
+
+// DecryptConnection decrypts password/token field for connection
+func DecryptConnection(connection interface{}, fieldName string) error {
```
klesh, line 175: Unit test pls

klesh: We should at least check if `password` is changed after encryption

plugins/jira/tasks/api_client.go
```diff
@@ -29,10 +29,12 @@ import (
 func NewJiraApiClient(taskCtx core.TaskContext, connection *models.JiraConnection) (*helper.ApiAsyncClient, error) {
 	// load configuration
 	encKey := taskCtx.GetConfig(core.EncodeKeyEnvStr)
-	auth, err := core.Decrypt(encKey, connection.BasicAuthEncoded)
+	decodedPassword, err := core.Decrypt(encKey, connection.Password)
```
klesh, line 32: Maybe we should decrypt it before this function. It doesn't feel right to have api_client to aware this logic

plugins/helper/connection.go
```diff
@@ -0,0 +1,214 @@
...
+)
+
+type BaseConnection struct {
+	Name string `gorm:"type:varchar(100);uniqueIndex" json:"name" validate:"required"`
+	common.Model
+}
+
+type BasicAuth struct {
+	Username string `mapstructure:"username" validate:"required" json:"username"`
+	Password string `mapstructure:"password" validate:"required" json:"password" encryptField:"yes"`
```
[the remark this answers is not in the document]
warren830: I'll change it back

plugins/jira/api/connection.go
```diff
@@ -410,7 +238,11 @@ type GranularitiesResponse struct {
 }
 
 func GetGranularitiesByConnectionId(input *core.ApiResourceInput) (*core.ApiResourceOutput, error) {
-	jiraConnection, err := findConnectionByInputParam(input)
```
klesh, line 413: I don't think `GetGranularitiesByConnectionId` and `GetBoardsByConnectionId` are being used anywhere, and they shouldn't be used anyway, I suggest that we delete them

plugins/helper/connection_test.go
```diff
@@ -0,0 +1,151 @@
...
+package helper
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type TestConnection struct {
```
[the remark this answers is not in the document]
warren830: done

plugins/jira/api/connection.go
```diff
@@ -246,27 +129,11 @@ PATCH /plugins/jira/connections/:connectionId
...
-	// load from db
-	jiraConnection, err := findConnectionByInputParam(input)
-	if err != nil {
-		return nil, err
-	}
-
-	// update from request and save to database
-	err = refreshAndSaveJiraConnection(jiraConnection, input.Body)
+	jiraConnection := &models.JiraConnection{}
+	err := helper.Patch(input, jiraConnection, db)
```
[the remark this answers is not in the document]
warren830: done

klesh (approved):
LGTM

PR #2030, opened 2022-05-30
add migration for directrun; update feishu model

### LINKED ISSUE #2014: We need Migration when run in cli/single plugin
## Description
Now run in cli/single plugin without migration. So I must run `make dev` to compile all plugins and run. It's not a direct way and it's so slow.

add some help for run migrations in cli and plugin's main

### LINKED ISSUE #2029: feishu plugin model need replace `id` to other primaryKeys
## Describe the bug
feishu plugin model need replace `id` to other primaryKeys.
It's necessary to extract

### LINKED ISSUE #2031: When Time is primaryKey in model, extract can not save any record
## Describe the bug
When Time is primaryKey in model, extract can not save any record

## To Reproduce
1. add a model with time column as primary key, such as ```StartTime        time.Time `gorm:"primaryKey"```
2. collect and extract
3. look at tool layer, there is no data

likyh:
# Summary

copy migrate from `services/init.go` to `directrun.go` for migration in cli;

delete `id` in  feishu model and add new primaryKey

# Summary 2
I want to keep this PR small, but I got a error when fix feishu model because of time cannot use as primary key.

So Fix it in commit 2. 

image
StartTime is a struct so `isPrimaryKey` not run and StartTime cannot detect as primary key.
origin_img_v2_2bb6a928-3b48-48c8-87c3-1b015f49678g

I fix it like this: check isPrimaryKey before check struct; add need_recursion to avoid unuseful recursion

close #2029
close #2014
close #2031

### Key Points

### DIFF
--- a/plugins/feishu/feishu.go
+++ b/plugins/feishu/feishu.go
@@ -73,7 +73,9 @@ func (plugin Feishu) RootPkgPath() string {
 }
 
 func (plugin Feishu) MigrationScripts() []migration.Script {
-	return []migration.Script{new(migrationscripts.InitSchemas)}
+	return []migration.Script{
+		new(migrationscripts.InitSchemas), new(migrationscripts.UpdateSchemas20220524),
+	}
 }
 
 func (plugin Feishu) ApiResources() map[string]map[string]core.ApiResourceHandler {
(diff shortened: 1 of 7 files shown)


klesh (approved):
LGTM

PR #2034, opened 2022-05-30
fix: jira connections api should not depend on encKey

### LINKED ISSUE #2032: Delete `.env` causing jira connection management complete broken
## Describe the bug
In the current jira implementation, we encrypt `basicAuth` before we store it in the database, and the encryption key is created and saved to the `.env` file if it doesn't exist.
This is OK, well, in most cases. but occasionally, due to debugging or accidental reasons, we have to recreate `.env`, which would cause `basicAuth` undecryptable, and the sensible way to address the problem is to let the developer/user be able to update the `basicAuth` with newly generated `encKey`.

## To Reproduce
1. Create a new jira connection, make sure it is connected
2. Shutdown devlake, delete `.env`file, and recopy from `.env.example`
3. Fire up devlake, and reload config-ui

## Expected behavior
1. jira connections should be able to be displayed
2. I should be able to update the `basicAuth` to fix the connection

## Screenshots
image

klesh:
# Summary

This PR fixes #2032 by the following approach:
 for Jira connections endpoint, when `.env` got deleted and decryption failed, no more 400 bad requests, but with empty `basic auth`, so developer/user can still list out all Jira connections, and update their `basic auth` which will then be encrypted with the newly generated key. 

  Fixes #2032

### DIFF
--- a/plugins/jira/api/connection.go
+++ b/plugins/jira/api/connection.go
@@ -128,7 +128,7 @@ func getJiraConnectionById(id uint64) (*models.JiraConnection, error) {
 	encKey := v.GetString(core.EncodeKeyEnvStr)
 	jiraConnection.BasicAuthEncoded, err = core.Decrypt(encKey, jiraConnection.BasicAuthEncoded)
 	if err != nil {
-		return nil, err
+		log.Error("failed to decrypt basic auth: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


warren830 (approved):
LGTM

PR #2035, opened 2022-05-30
fix(jenkins): job collector pager

### LINKED ISSUE #2036: jenkins job collector got endless loop
## Describe the bug
jenkins job collector got endless loop when number of jobs is larger than page size 

## To Reproduce
set page size to be less than number of jobs
run jenkins
it will never stop

## Expected behavior
page number should get changed for every request

warren830:
# Summary

add tree value  like build collector in Query to define how many jobs we want every time

closes #2036

### Screenshots
image
### Other Information

### DIFF
--- a/plugins/jenkins/tasks/job_collector.go
+++ b/plugins/jenkins/tasks/job_collector.go
@@ -19,7 +19,9 @@ package tasks
 
 import (
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


mindlesscloud (approved):
LGTM

PR #2042, opened 2022-05-31
fix: unexpected EOF and busy buffer

### LINKED ISSUE #1992: data is missing for pingcap collectApiPullRequestCommits subtask
## Describe the bug
1. pull_request total sums: 23263
ibvKvePPCy
2. when the subtask run the pr.numer=13243(or other random one), the subtask finished
image

3. we can see the mysql has error:  **unexpected EOF and busy buffer**
so, the collectApiPullRequestCommits subtask finished.

### LINKED ISSUE #2043: Refactor: `ApiCollector.Input`  should only retrieve necessary columns
## What and why to refactor

For the sake of getting HTTP request params, the `ApiCollector` might need to iterate over a table. In most cases, we only need one or two columns, but some iterators return the whole row. 

1. check all plugin collectors for similar problems
2. select  useful columns

abeizn:
# Summary
fix #1992 : data is missing for pingcap collectApiPullRequestCommits subtask
fix #2043: Refactor: ApiCollector.Input should only retrieve necessary columns

### DIFF
--- a/plugins/github/tasks/pr_commit_collector.go
+++ b/plugins/github/tasks/pr_commit_collector.go
@@ -52,7 +52,7 @@ func CollectApiPullRequestCommits(taskCtx core.SubTaskContext) error {
 
 	incremental := false
 
-	cursor, err := db.Model(&models.GithubPullRequest{}).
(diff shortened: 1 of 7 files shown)


klesh (approved):
LGTM

PR #2046, opened 2022-05-31
fix: some database fields are not compliant. so some dashboards need to update.

abeizn:
# Summary
fix #2045 : due to some database fields update, dashboards need to update.
change key to pull_request_key

### Screenshots

image


mindlesscloud (approved):
LGTM

PR #2096, opened 2022-06-06
Add TagsPattern for use Pattern in refdiff

### LINKED ISSUE #1983: Refdiff to support another set of parameters tag regex and tag limit, and calculate diffs based on these 2 parameters

mappjzc:
# Summary
Add TagsPattern to set Pattern for refdiff
Add TagsLimit to limit the count of TagsPattern matching.
Add TagsOrder to set the order rule for  list of TagsPattern  matching.
Add surpport of DirctRun for new param. 
Optimize partial logs
The TagsPattern support CalculateIssuesDiff and CalculateCommitsDiff.

close #1983 

### Screenshots
image
image

image
image

### Other Information

### DIFF
--- a/plugins/refdiff/README.md
+++ b/plugins/refdiff/README.md
@@ -85,7 +85,7 @@ and if you want to perform certain subtasks.
 ]
 ```
 Or you can use tagsPattern to match the tags you want
-And you can use tagOrder (support `alphabetically` and `reverse alphabetically`) to set the order rule with tagLimit to limit the count of matching.
+And you can use tagOrder (support `alphabetically`,`reverse alphabetically`,`semver`,`reverse semver`) to set the order rule with tagLimit to limit the count of matching.
 This is support to calculateCommitsDiff and calculateIssuesDiff
 ```json
 [
@@ -100,7 +100,7 @@ This is support to calculateCommitsDiff and calculateIssuesDiff
         "repoId": "github:GithubRepo:384111310",
         "tagsPattern":".*\\.11\\..*",
         "tagLimit":3,
-        "tagOrder":"reverse alphabetically",
+        "tagOrder":"reverse semver",
       }
     }
   ]
--- a/plugins/refdiff/README.md
+++ b/plugins/refdiff/README.md
@@ -1,10 +1,4 @@
 # RefDiff
-
-
-| [English](README.md) | [中文](README-zh-CN.md) |
-| --- | --- |
-
-
 ## Summary
 
 For development workload analysis, we often need to know how many commits have been created between 2 releases. This plugin offers the ability to calculate the commits of difference between 2 Ref(branch/tag), and the result will be stored back into database for further analysis.
@@ -85,6 +79,28 @@ and if you want to perform certain subtasks.
             "oldRef": "refs/tags/0.4.0"
           }
         ]
+      }
+    }
+  ]
+]
+```
+Or you can use tagsPattern to match the tags you want
+And you can use tagOrder (support `alphabetically` and `reverse alphabetically`) to set the order rule with tagLimit to limit the count of matching.
+This is support to calculateCommitsDiff and calculateIssuesDiff
+```json
+[
+  [
+    {
+      "plugin": "refdiff",
+      "subtasks": [
+        "calculateCommitsDiff",
+        "calculateIssuesDiff",
+      ],
+      "options": {
+        "repoId": "github:GithubRepo:384111310",
+        "tagsPattern":".*\\.11\\..*",
+        "tagLimit":3,
+        "tagOrder":"reverse alphabetically",
       }
     }
   ]
--- a/plugins/refdiff/refdiff.go
+++ b/plugins/refdiff/refdiff.go
@@ -94,7 +94,10 @@ func main() {
 	//_ = refdiffCmd.MarkFlagRequired("old-ref")
 
 	refdiffCmd.Run = func(cmd *cobra.Command, args []string) {
-		tl, _ := strconv.Atoi(*tagsLimit)
+		tl, err := strconv.Atoi(*tagsLimit)
+		if err != nil {
+			panic(err)
+		}
 		pairs := make([]map[string]string, 0, 1)
 		if *newRef == "" && *oldRef == "" {
 			if *tagsPattern == "" {
--- a/plugins/refdiff/refdiff.go
+++ b/plugins/refdiff/refdiff.go
@@ -64,6 +64,16 @@ func (plugin RefDiff) PrepareTaskData(taskCtx core.TaskContext, options map[stri
 		return nil, err
 	}
 
+	db := taskCtx.GetDb()
+	tagsPattern := op.TagsPattern
+	tagsLimit := op.TagsLimit
+	tagsOrder := op.TagsOrder
+
+	op.TagsRefs, err = tasks.CaculateTagPattern(db, tagsPattern, tagsLimit, tagsOrder)
+	if err != nil {
+		return nil, err
+	}
+
 	return &tasks.RefdiffTaskData{
 		Options: &op,
 	}, nil
--- a/plugins/refdiff/refdiff.go
+++ b/plugins/refdiff/refdiff.go
@@ -18,8 +18,6 @@ limitations under the License.
 package main
 
 import (
-	"strconv"
-
 	"github.com/apache/incubator-devlake/plugins/refdiff/tasks"
 	"github.com/apache/incubator-devlake/runner"
 	"github.com/mitchellh/mapstructure"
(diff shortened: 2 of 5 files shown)


plugins/refdiff/refdiff.go
```diff
@@ -83,19 +85,34 @@ func main() {
 	newRef := refdiffCmd.Flags().StringP("new-ref", "n", "", "new ref")
 	oldRef := refdiffCmd.Flags().StringP("old-ref", "o", "", "old ref")
 
+	tagsPattern := refdiffCmd.Flags().StringP("tags-pattern", "p", "", "tags pattern")
+	tagsLimit := refdiffCmd.Flags().StringP("tags-limit", "l", "", "tags limit")
+	tagsOrder := refdiffCmd.Flags().StringP("tags-order", "d", "", "tags order")
+
 	_ = refdiffCmd.MarkFlagRequired("repo-id")
```
warren830, line 92: if we don't process the returned err, why do we call `MarkFlagRequired`

plugins/refdiff/refdiff.go
```diff
@@ -83,19 +85,34 @@ func main() {
 	newRef := refdiffCmd.Flags().StringP("new-ref", "n", "", "new ref")
 	oldRef := refdiffCmd.Flags().StringP("old-ref", "o", "", "old ref")
 
+	tagsPattern := refdiffCmd.Flags().StringP("tags-pattern", "p", "", "tags pattern")
+	tagsLimit := refdiffCmd.Flags().StringP("tags-limit", "l", "", "tags limit")
+	tagsOrder := refdiffCmd.Flags().StringP("tags-order", "d", "", "tags order")
```
warren830, line 90: can you write some comments for these three variables?

plugins/refdiff/refdiff.go
```diff
@@ -83,19 +85,34 @@ func main() {
...
+
 	_ = refdiffCmd.MarkFlagRequired("repo-id")
-	_ = refdiffCmd.MarkFlagRequired("new-ref")
-	_ = refdiffCmd.MarkFlagRequired("old-ref")
+	//_ = refdiffCmd.MarkFlagRequired("new-ref")
+	//_ = refdiffCmd.MarkFlagRequired("old-ref")
 
 	refdiffCmd.Run = func(cmd *cobra.Command, args []string) {
+		tl, _ := strconv.Atoi(*tagsLimit)
```
warren830, line 97: why swallow the err?

plugins/refdiff/tasks/ref_commit_diff_calculator.go
```diff
@@ -27,14 +27,27 @@ import (
 	"gorm.io/gorm/clause"
 )
 
-func CalculateCommitsDiff(taskCtx core.SubTaskContext) error {
+func CalculateCommitsPairs(taskCtx core.SubTaskContext) ([][4]string, error) {
```
warren830, line 30: please write comment to describe what does this func do

plugins/refdiff/tasks/ref_commit_diff_calculator.go
```diff
@@ -27,14 +27,27 @@ import (
...
 	data := taskCtx.GetData().(*RefdiffTaskData)
 	repoId := data.Options.RepoId
 	pairs := data.Options.Pairs
+	tagsLimit := data.Options.TagsLimit
 	db := taskCtx.GetDb()
-	ctx := taskCtx.GetContext()
-	logger := taskCtx.GetLogger()
-	insertCountLimitOfRefsCommitsDiff := int(65535 / reflect.ValueOf(code.RefsCommitsDiff{}).NumField())
+
+	rs, err := CaculateTagPattern(taskCtx)
```
warren830, line 37: same as above

plugins/refdiff/tasks/ref_issue_diff_calculator.go
```diff
@@ -26,15 +26,40 @@ import (
 	"github.com/apache/incubator-devlake/plugins/helper"
 )
 
-func CalculateIssuesDiff(taskCtx core.SubTaskContext) error {
+func CaculatePairList(taskCtx core.SubTaskContext) ([][2]string, error) {
```
warren830, line 29: please write comment

plugins/refdiff/tasks/refdiff_task_data.go
```diff
@@ -36,3 +46,77 @@ type RefPair struct {
...
+
+func (rs RefsReverseAlphabetically) Less(i, j int) bool {
+	return rs[i].Id > rs[j].Id
+}
+
+func (rs RefsReverseAlphabetically) Swap(i, j int) {
+	rs[i], rs[j] = rs[j], rs[i]
+}
+
+func CaculateTagPattern(taskCtx core.SubTaskContext) (Refs, error) {
```
warren830, line 82: please write comment

plugins/refdiff/tasks/refdiff_task_data.go
```diff
@@ -36,3 +46,77 @@ type RefPair struct {
...
+		if err != nil {
+			return rs, err
+		}
+
+		if ok := r.Match([]byte(ref.Id)); ok {
+			rs = append(rs, ref)
+		}
+	}
+	switch tagsOrder {
+	case "alphabetically":
```
warren830, line 114: I think `alphabetically` is too long for an arg

plugins/refdiff/README.md
```diff
@@ -90,6 +84,28 @@ and if you want to perform certain subtasks.
   ]
 ]
 ```
+Or you can use tagsPattern to match the tags you want
+And you can use tagOrder (support `alphabetically` and `reverse alphabetically`) to set the order rule with tagLimit to limit the count of matching.
+This is support to calculateCommitsDiff and calculateIssuesDiff
```
warren830, line 89: The following json supports both calculateCommitsDiff and calculateIssuesDiff

plugins/refdiff/README.md
```diff
@@ -90,6 +84,28 @@ and if you want to perform certain subtasks.
   ]
 ]
 ```
+Or you can use tagsPattern to match the tags you want
```
warren830, line 87: You can also use the following args:
tagsPattern: use regex pattern to match the tags you want
tagOrder: (either `alphabetically` or `reverse alphabetically`) set how you want to order the tags
tagLimit: limit the count of matching

plugins/refdiff/tasks/ref_issue_diff_calculator.go
warren830: and what is the [][2]string? Or define a Pairs to replace [2]string

warren830 (changes requested):

plugins/refdiff/refdiff.go
klesh: @warren830  there is a misunderstanding here,  errors occur when the name is invalid or missing. 
it is safe to ignore the error here.
and, we did this everywhere in plugin.main

plugins/refdiff/refdiff.go
```diff
@@ -83,19 +85,37 @@ func main() {
 	newRef := refdiffCmd.Flags().StringP("new-ref", "n", "", "new ref")
 	oldRef := refdiffCmd.Flags().StringP("old-ref", "o", "", "old ref")
 
+	tagsPattern := refdiffCmd.Flags().StringP("tags-pattern", "p", "", "tags pattern")
+	tagsLimit := refdiffCmd.Flags().StringP("tags-limit", "l", "", "tags limit")
```
klesh, line 89: why not make it `IntP` or `UnitP`? then you won't have to convert it youself.

klesh: should have assigned a default value for it, like 10

plugins/refdiff/README.md
klesh: I think alphabelcially order is no enough.
we should consider `semver` sorting, sth in that nature

plugins/refdiff/tasks/ref_commit_diff_calculator.go
```diff
@@ -27,14 +27,28 @@ import (
...
```
klesh, line 38: I think we should calculate `pairs` inside `PrepareTaskData` method once and for all instead of calculation inside every subtasks

klesh (approved):
LGTM

warren830 (approved):
LGTM

PR #2104, opened 2022-06-07
chore: create alpine-dbt-mysql

### LINKED ISSUE #1692: Build an image with python and dbt-mysql
## Description
I noticed that we need install python3 and dbt-mysql on an alpine:3.15 image everytime when we build lake image. If we provide an image with preinstalled these tools, it will be futher faster to build lake image.

Not find.

narrowizard:
# Summary
preinstall python and dbt-mysql in mericodev/alpine-dbt-mysql image.
to prevent install it every time we release lake

fix #1692

### Screenshots

### Other Information
Before merge this PR, we should push image `mericodev/alpine-dbt-mysql` to docker hub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,15 +27,7 @@ ENV GOBIN=/app/bin
 
 RUN make clean && make all
 
-FROM --platform=linux/amd64 alpine:3.15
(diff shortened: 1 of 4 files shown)


warren830 (approved):
LGTM

PR #2105, opened 2022-06-07
fix(ci): remove add tar in github action and modify lake-builder image

### LINKED ISSUE #2102: CI: golangci-lint post job fails
## Describe the bug
golangci-lint post job requires standard tar tools.
It just impacts the post job.

## To Reproduce
N/A

## Expected behavior
Hope ` golangci-lint post` success

## Screenshots
image

## Additional context
We start using `mericodev/lake-builder:latest` as golang lint container after #2084.
But no standalone `tar` in that images, we need to add tar tools in that devel container image.

warren830:
# Summary

Previously, we don't have tar in our base image lake-builder, so we have to run `add no cache tar` during github action workflow.
In this pr, we removed all `add no cache tar` in *.yml and add it to base image dockerfile

closes #2102

### Screenshots

### Other Information

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -24,10 +24,8 @@ jobs:
           MYSQL_USER: merico
           MYSQL_PASSWORD: merico
           MYSQL_ROOT_PASSWORD: root
-    container: mericodev/lake-builder:0.0.4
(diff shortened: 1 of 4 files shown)


mindlesscloud (approved):
LGTM

Issue #2130, opened 2022-06-08
[Feature][Helper/ApiCollector] Make query functions support getting token from collector.args.Params


SnowMoon-Dev:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Make query functions support getting token from collector.args.Params

### Use case

In the query function, support getting token from collector.args.Params, in order to request the following url : https://xxx.com/api/v5/repos/xxx/incubator-devlake/commits?access_token=02d1de27b9457cd&page=1&per_page=20

### Related issues

[#2129](https://github.com/apache/incubator-devlake/issues/2129)

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

klesh:
The use case doesn't seem legit.
Why would authentication get passed by `task.options`?
It should be with `connection` and handled by `ApiClient.SetBeforeFunction`, Check `plugins/github/tasks/api_client.go` for detail.
Do you have any actual cases to prove the necessity of this PR?

SnowMoon-Dev:
I have seen the implementation of ApiClient.SetBeforeFunction, but it can't meet the needs. I am writing a plug-in for gitee. The verification mode of gitee is different from that of github and gitblab. The token is placed on the query condition.
<img width="778" alt="image" src="https://user-images.githubusercontent.com/42627702/172977569-69d90110-9777-4f43-b868-12a101d16706.png">
Here is my local implementation

SnowMoon-Dev:
> The use case doesn't seem legit. Why would authentication get passed by `task.options`? It should be with `connection` and handled by `ApiClient.SetBeforeFunction`, Check `plugins/github/tasks/api_client.go` for detail. Do you have any actual cases to prove the necessity of this PR?

The previous idea was to get the token information in the configuration file env from the context of the task, and later @warren830  gave me a better suggestion

SnowMoon-Dev:
The problem can be solved by changing the getQuery function into an anonymous function, no need to modify the Query interface

PR #2131, opened 2022-06-08
[improvement][apiCollector]Make query functions support getting token from collector.args.Params

SnowMoon-Dev:
close [#2130](https://github.com/apache/incubator-devlake/issues/2130)

## Verify this pull request

Manually verified the change by testing locally.


SnowMoon-Dev:
The problem can be solved by changing the getQuery function into an anonymous function, no need to modify the Query interface

Issue #2153, opened 2022-06-10
[Bug] [helper] api_collector is missing the value of the input parameter


abeizn:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

Jenkins run subtask `collectApiBuilds`, you can find the lake._raw_jenkins_api_builds input parameter is NULL.
so I think the input parameter was forgotten to be passed.

### What you expected to happen

the lake._raw_jenkins_api_builds input parameter is not NULL.

### How to reproduce

run 
`go run plugins/jenkins/jenkins.go -c 1 -t=collectApiBuilds`

### Version

main

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Issue #2140, opened 2022-06-09
[Bug] [GitHub Plugin] ExtractApiPullRequestCommits throwing JSON error


keon94:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

Running the GitHub pipeline on a few repos throws "unexpected end of JSON input" on [this line](https://github.com/apache/incubator-devlake/blob/0d3aa41e5e3db60e7ddfd3c3835f40e8c1c8eeb4/plugins/github/tasks/pr_commit_extractor.go#L85). This is being caused by the "nil" row.Input which the code tries to deserialize. 

### What you expected to happen

row.Input should not be nil

### How to reproduce

here's one [repo](https://github.com/keon94/go-compose) that I used that threw this error, but this probably happens elsewhere too. Just run the GitHub pipeline on it and the JSON error mentioned above will show up on the UI.

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

keon94:
FYI @klesh

PR #2154, opened 2022-06-10
fix: api_collector is missing the value of the input parameter

abeizn:
# Summary
I think it is missing the InputJson. so `Input:  reqData.InputJSON` is NULL

image

image

fix: #2153 
fix: #2140

### Screenshots
image

### Other Information


klesh (approved):
LGTM

PR #2185, opened 2022-06-14
fix: update number to xxx_key

### LINKED ISSUE #1959: the ”number“ field in pull_requests table is deprecated.
## Describe the bug
the "number" field  is same to the "key" field, then we need to delete the number field.

## Screenshots
image

abeizn:
# Summary

some number filed is deprecated, so we need to delete it. 
and some number filed needs to update.

fix #1959


warren830 (approved):
LGTM

PR #2190, opened 2022-06-14
update e2e test helper

likyh:
# Summary

1.  add TableName for the domain layer model
2. add some helpers (csv writer / migrating table and flush) for e2e test
3. use updated e2e test helper to update gitlab e2e test. (Notice: detect connection id not fix bug)

related #1980


helpers/e2ehelper/data_flow_tester.go
```diff
@@ -71,45 +77,67 @@ func NewDataFlowTester(t *testing.T, pluginName string, pluginMeta core.PluginMe
...
 		Name:   pluginName,
 		Plugin: pluginMeta,
 		Log:    logger.Global,
 	}
 }
 
-// ImportCsv imports records from specified csv file into target table, note that existing data would be deleted first.
-func (t *DataFlowTester) ImportCsv(csvRelPath string, tableName string) {
+// ImportCsvIntoRawTable imports records from specified csv file into target table, note that existing data would be deleted first.
+func (t *DataFlowTester) ImportCsvIntoRawTable(csvRelPath string, tableName string) {
```
klesh, line 101: Why? I thought we agree that we should be able to import table other than `RawTable`

helpers/e2ehelper/data_flow_tester.go
```diff
@@ -71,45 +77,67 @@ func NewDataFlowTester(t *testing.T, pluginName string, pluginMeta core.PluginMe
...
-	t.FlushTable(tableName)
+	t.FlushRawTable(tableName)
 	// load rows and insert into target table
 	for csvIter.HasNext() {
-		// make sure
-		result := t.Db.Table(tableName).Create(csvIter.Fetch())
+		toInsertValues := csvIter.Fetch()
+		// FIXME Hack code
+		if t.Db.Dialector.Name() == `postgres` {
+			toInsertValues[`data`] = strings.Replace(toInsertValues[`data`].(string), `\`, `\\`, -1)
```
klesh, line 110: Maybe we can check if the `field` is `json.RawMessage`, and convert it to `[]byte` so it could be inserted for all kind of database properly?

helpers/e2ehelper/data_flow_tester.go
```diff
@@ -71,45 +77,67 @@ func NewDataFlowTester(t *testing.T, pluginName string, pluginMeta core.PluginMe
...
 			panic(result.Error)
 		}
 		assert.Equal(t.T, int64(1), result.RowsAffected)
 	}
 }
 
-// FlushTable deletes all records from specified table
-func (t *DataFlowTester) FlushTable(tableName string) {
+// MigrateRawTableAndFlush migrate table and deletes all records from specified table
+func (t *DataFlowTester) FlushRawTable(rawTableName string) {
```
klesh, line 121: Again, we should be able to flush other kinds of tables as well, shouldn't we?

helpers/e2ehelper/data_flow_tester.go
```diff
@@ -124,11 +152,81 @@ func (t *DataFlowTester) Subtask(subtaskMeta core.SubTaskMeta, taskData interfac
 	}
 }
 
+// CreateSnapshot reads rows from database and write them into .csv file.
+func (t *DataFlowTester) CreateSnapshot(dst schema.Tabler, csvRelPath string, pkfields []string, targetfields []string) {
+	location, _ := time.LoadLocation(`UTC`)
+	var allFields []string
```
klesh, line 158: Mind the linting error

helpers/e2ehelper/data_flow_tester.go
likyh: Raw table only needs name, and other tables need dst, which is the reason for the difference 😹 Unless reflection is added to the function. But it either cannot say that all string names are raw table.

helpers/e2ehelper/data_flow_tester.go
likyh: `json.RawMessage` ?? It's another question.

Here Question is that bytes at raw tables in pg need `\\n` to insert but only need `\n` in MySQL.

helpers/e2ehelper/data_flow_tester.go
likyh: fixed

helpers/e2ehelper/data_flow_tester.go
likyh: ImportCsvIntoTabler will added soon. (the coming PR which need to import tool data)

klesh (approved):
LGTM

PR #2219, opened 2022-06-16
fix bug report template label setting

klesh:
# Summary

Recently recreated "bug report" doesn't come with `type/bug` label
image

This PR aim to fix the problem


mindlesscloud (approved):
LGTM

PR #2238, opened 2022-06-17
Run config-ui nginx container as non-root

marcemv90:
config-ui nginx was running as root, which is not a best practice, as an attacker can potentially become root in a given kubernetes node using some security breach. This commit switches to use an unprivileged nginx container, which runs as nginx user. In addition, it fixes the tag of the image to inherit from, as using latest is also not a good practice, as it can lead to introduce breaking changes if the parent image moves the latest tag to a new image with said breaking changes.

# Summary

No

### Screenshots
We have it running with those changes in our Kubernetes cluster with no issues. Notice the custom securityContext configuration that we set states to not run the containers as root:

image

### Other Information
N/A

### DIFF
--- a/config-ui/Dockerfile
+++ b/config-ui/Dockerfile
@@ -31,7 +31,8 @@ RUN npm ci
 COPY . /home/node/code
 RUN npm run build-production
 
-FROM nginx:latest
+FROM nginxinc/nginx-unprivileged:1.21
+USER 0
 #ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.9.0/wait /wait
 #RUN chmod +x /wait
 RUN rm /etc/nginx/conf.d/default.conf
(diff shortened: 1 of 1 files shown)


klesh:
Hi, thanks for your contribution.
It is very helpful.

klesh (approved):
LGTM

warren830:
Thanks for your contribution! Would you like to join us on slack, then we can discuss some problems together

marcemv90:
> Thanks for your contribution! Would you like to join us on slack, then we can discuss some problems together

Sure, hopefully I can help 😂

warren830:
> 



> > Thanks for your contribution! Would you like to join us on slack, then we can discuss some problems together
> 
> Sure, hopefully I can help 😂

Hi, did you join slack? I did not find you, I need to issue you a certificate to honor your contribution. Or can you send an email to me(warren@apache.org)

Issue #2262, opened 2022-06-20
[Feature][tapd] multi-data connections support for tapd


abeizn:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Users often have multi instances of tapd, we should set up multi tapd connection and collect data independently. It's important to isolate data that from multi connection.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #2267, opened 2022-06-20
feat: multi-conn for tapd and update migrationscripts

abeizn:
# Summary

update api/connection to load and save through db not .env
and update migrationscripts

related to #2262 

### Screenshots
image


plugins/tapd/models/migrationscripts/init_schema.go
```diff
@@ -19,6 +19,7 @@ package migrationscripts
 
 import (
 	"context"
+
 	"github.com/apache/incubator-devlake/plugins/tapd/models/migrationscripts/archived"
 	"gorm.io/gorm"
 )
```
warren830, line 25: we need to drop tables first

warren830 (commented):
LGTM, only need a minor fix for init_schema

warren830 (approved):
LGTM

PR #2272, opened 2022-06-20
fix ci

likyh:
# Summary

Fix ci broken by Github user.

`AutoIncrement` is not necessary for `GithubUser`, so I delete it in MySQL.


klesh (approved):
LGTM

Issue #2282, opened 2022-06-21
[Bug][tapd] some field is error and fix bugs


abeizn:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

1. Required is not Unnecessary
![image](https://user-images.githubusercontent.com/101256042/174739073-40c8bc18-904c-4aee-a2a0-4484fdab9865.png)
2. TapdOptions: WorkspaceID and workspceId is error.
![image](https://user-images.githubusercontent.com/101256042/174739306-63cf762d-f97a-4ccf-825d-e322e41c0dbd.png)
3. db.First 
![image](https://user-images.githubusercontent.com/101256042/174764552-4915025c-4ad9-4ba4-a492-e1631b6c5262.png)
4. time.Time format
![image](https://user-images.githubusercontent.com/101256042/174780206-55bc43cc-4291-4149-85f6-209da98a5f57.png)

### What you expected to happen

some field is error

### How to reproduce

reviewed code

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #2289, opened 2022-06-21
fix(tapd): ignore err if it's record not found

warren830:
# Summary

ignore err if it's `record not found` when fetch latestUpdate from db in `Collector`

relate to #2282 

### Screenshots

### Other Information


abeizn (approved):
LGTM

PR #2325, opened 2022-06-22
Debug with delve / compile-script swallowing compilation error

klesh:
# Summary

1. run `make debug` to fire up api server with `delve` for breakpoint debugging
2. fix `scripts/compile-plugins.sh` error swallowing, it keeps going even if the compilation failed

image


warren830 (approved):
LGTM

Issue #1982, opened 2022-05-24
add `settings` to `blueprint` as `json` to represent user input for `NORMAL` mode 


klesh:
## Description
Design: https://www.figma.com/file/Eux4Dhorxp7M0TwzOnL6Yh/Config-UI-for-V1.0?node-id=293%3A10542

## Describe the solution you'd like
Each step is a independent PR
Step 1: create `BlueprintSetting` struct to represent user input from `config-ui`, provide proper api and save/load to/from db
Step 2: auto generate `tasks` when save to db

Blueprint setting json
```json
{
    "ID": 0,
    "name": "MY BLUEPRINT",
    "tasks": [
        []
    ],
    "settings": {
        "version": "1.0",
        "connections": [
            {
                "connectionId": 1,
                "plugin": "github",
                "scope": [
                    {
                        "options": {
                            "owner": "merico-dev",
                            "repo": "lake"
                        },
                        "entities": [
                            "code",
                            "ticket"
                        ],
                        "transformation": {
                            "prType": "type/(.*)$",
                            "prComponent": "component/(.*)$",
                            "issueSeverity": "severity/(.*)$",
                            "issueComponent": "component/(.*)$",
                            "issuePriority": "^(highest|high|medium|low)$",
                            "issueTypeRequirement": "^(feat|feature|proposal|requirement)$",
                            "issueTypeBug": "^(bug|failure|error)$",
                            "issueTypeIncident": ""
                        }
                    },
                    {
                        "options": {
                            "owner": "merico-dev",
                            "repo": "lake"
                        },
                        "entities": [
                            "code",
                            "ticket"
                        ],
                        "transformation": {
                            "prType": "type/(.*)$",
                            "prComponent": "component/(.*)$",
                            "issueSeverity": "severity/(.*)$",
                            "issueComponent": "component/(.*)$",
                            "issuePriority": "^(highest|high|medium|low)$",
                            "issueTypeRequirement": "^(feat|feature|proposal|requirement)$",
                            "issueTypeBug": "^(bug|failure|error)$",
                            "issueTypeIncident": ""
                        }
                    }
                ]
            },
            {
                "connectionId": 1,
                "plugin": "gitlab",
                "scope": [
                    {
                        "options": {
                            "projectId": "1000"
                        },
                        "entities": [
                            "code",
                            "ticket"
                        ],
                        "transformation": {}
                    },
                    {
                        "options": {
                            "projectId": "2000"
                        },
                        "entities": [
[...]

PR #2326, opened 2022-06-22
feat: blueprint normal mode framework-side

klesh:
# Summary

1. `blueprint.tasks` and `pipeline.tasks` were renamed to `plan`
2. `dal.First` would return `dal.ErrRecordNotFound` instead of `gorm.ErrRecordNotFound`
3. migration script is written in a novel convention for better readability, feel free to argue
4. define `Blueprint` protocol `1.0.0` including `interface` and `struct` they should adopt, so the framework and plugin can generate Pipeline.Plan based on it
5. define `DOMAIN_TYPES` (aka, `entities` from config-ui), the naming is open for discussion.
6. blueprint would call plugin to generate Plans, and then merge them into one unified Plan

**This is framework implementation, which would require plugin support to produce a meaningful result.**

Part of #1982


likyh (approved):
LGTM

warren830 (approved):
LGTM

PR #2331, opened 2022-06-23
Github blueprint normal mode support

klesh:
# Summary

Github blueprint normal mode support implementation

Part of #1982


utils/strings.go
```diff
@@ -0,0 +1,24 @@
+package utils
```
abeizn, line 1: Missed header

plugins/github/api/blueprint.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/apache/incubator-devlake/plugins/core"
+	"github.com/apache/incubator-devlake/plugins/core/dal"
+	"github.com/apache/incubator-devlake/plugins/github/models"
+	"github.com/apache/incubator-devlake/plugins/github/tasks"
+	"github.com/apache/incubator-devlake/plugins/helper"
+	"github.com/apache/incubator-devlake/utils"
+)
+
+func MakePipelinePlan(subtaskMetas []core.SubTaskMeta, connectionId uint64, scope []*core.BlueprintScopeV100) (core.PipelinePlan, error) {
```
likyh, line 35: MakePipelinePlan should have a default implementation to keep plugin development simple.

plugins/github/api/blueprint.go
```diff
@@ -0,0 +1,133 @@
...
+		err = json.Unmarshal(scopeElem.Options, &taskOptions)
+		if err != nil {
+			return nil, err
+		}
+		err = json.Unmarshal(scopeElem.Transformation, &taskOptions)
+		if err != nil {
+			return nil, err
+		}
+		taskOptions["connectionId"] = connectionId
+		op, err := tasks.DecodeAndValidateTaskOptions(taskOptions)
```
likyh, line 50: if `op` just use in `if utils.StringsContains(scopeElem.Entities, core.DOMAIN_TYPE_CODE) {`, it should defined in `if` statement.

plugins/github/impl/impl.go
```diff
@@ -174,3 +134,7 @@ func (plugin Github) ApiResources() map[string]map[string]core.ApiResourceHandle
 		},
 	}
 }
+
+func (plugin Github) MakePipelinePlan(connectionId uint64, scope []*core.BlueprintScopeV100) (core.PipelinePlan, error) {
```
likyh, line 138: maybe connectionId is not used in some plugins?

likyh: what the `entity` in `BlueprintScopeV100` means?

plugins/helper/pipeline_plan.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+
+	"github.com/apache/incubator-devlake/plugins/core"
+	"github.com/apache/incubator-devlake/utils"
+)
+
+// MakePipelinePlanSubtasks generates subtasks list based on sub-task meta information and entities wanted by user
+func MakePipelinePlanSubtasks(subtaskMetas []core.SubTaskMeta, entities []string) ([]string, error) {
```
likyh, line 28: It's confusing between running all tasks and running no task

plugins/github/tasks/comment_extractor.go
```diff
@@ -33,6 +34,7 @@ var ExtractApiCommentsMeta = core.SubTaskMeta{
 	EnabledByDefault: true,
 	Description: "Extract raw comment data  into tool layer table github_pull_request_comments" +
 		"and github_issue_comments",
+	DomainTypes: []string{core.DOMAIN_TYPE_CODE, core.DOMAIN_TYPE_TICKET},
```
likyh, line 37: how can some plugins use without DomainTypes?

plugins/github/api/blueprint.go
```diff
@@ -0,0 +1,133 @@
...
+				Plugin: "gitextractor",
+				Options: map[string]interface{}{
+					// TODO: url should be configuration
+					// TODO: to support private repo: username is needed for repo cloning, and we have to take
+					//       multi-token support into consideration, this is hairy
+					"url":    fmt.Sprintf("https://github.com/%s/%s.git", op.Owner, op.Repo),
+					"repoId": fmt.Sprintf("github:GithubRepo:%d", repoId),
+				},
+			})
+			// TODO, add refdiff in the future
```
likyh, line 128: why gitextractor should be added by the GitHub plugin instead of defined in the plan?

likyh: If I start the github & the ae plugin together and they both add `gitextractor`, will `gitextractor` be duplicated?

likyh (changes requested):

plugins/github/api/blueprint.go
warren830: > 

I think only github and gitlab need gitex to collect commits and refs

warren830: > why gitextractor should be added by the GitHub plugin instead of defined in the plan?

right now, we dropped collecting commits data by api because it's too slow. Using gitextractor will speed it up.

plugins/github/tasks/comment_extractor.go
klesh: For `data source` plugins, it is a must, one should implement `PluginBlueprint` so we know it supports Blueprint Normal Mode, which means it can take some `input`(different plugin has different input) from New Config-UI and generate a reasonable `PipelinePlan` .
For others, like `gitext` and `refdiff`, I don't think we need to declare DomainType for each subtasks, because it is generated by `data source` plugins, as a task inside the `PipelinePlan`

plugins/helper/pipeline_plan.go
klesh: `subtasks` is an optional parameter, running a `task` without any `subtasks` is meaningless in our system at this point.
But, it is another topic, feel free to open a discussion on dev@devlake.a.o.  
It is out of scope for this PR

abeizn (approved):
LGTM

plugins/github/api/blueprint.go
likyh: > > 
> 
> I think only github and gitlab need gitex to collect commits and refs

But Maybe the user has already collected it before. Or maybe the user use other custom `gitextractor` because of private repo?

likyh (approved):
Although there are some problems. But it's enough now.

plugins/github/api/blueprint.go
klesh: The private repo is already supported by `gitextractor`. we want users to improve `gitextractor` if they found anything missing.
if a Custom `gitextractor` is needed for any reason, he/she would have to use Advanced Mode anyway. And this is for Normal Mode.

klesh:
@e2corporation Github normal mode is ready, keep in mind `blueprint.tasks` and `pipeline.tasks` has been renamed to `plan`. Since it caused many naming problems, it was hard to write human-readable code with it :joy:

klesh:
@narrowizard

PR #2347, opened 2022-06-24
CI reported failed for non-semver tags which actually succeeded

klesh:
# Summary

CI reported failed for non-semver tags which actually succeeded


mindlesscloud (approved):
LGTM

PR #2356, opened 2022-06-24
add e2e raw table generator and change the e2e comparison way

### LINKED ISSUE #1822: Create plugin by template
## Description
When I was working on tapd plugin, I found that sub tasks shared a lot of common code. Most of time, we only need to fill some words. So I am thinking of creating plugin subtasks by cmd, then make some minor changes. (Maybe we can even do the entity mapping work on config-ui, and save the config and run plugins without coding? Just a knocking-head thought)

I read similar code from devstream, we can learn from them.
1. create template files
2. create cmd to create plugins subtasks by input `plugin_name`, `entity_name` and `subtask`

Please link any other

### LINKED ISSUE #2222: E2E test helper needs to resolve pk fields by models auto
## Description
E2E test helper needs to resolve pk fields by models auto

Add some reflection in `toVerify` to resolve which primary fields used in models

No

likyh:
# Summary

add e2e raw table generator and change the e2e comparison way

close #2222 
close #1822 

### Screenshots
Untitled


klesh (approved):
LGTM

PR #2392, opened 2022-06-30
fix(jira): change filter for jira_issues

### LINKED ISSUE #2386: [Bug][jira] `board_issues` contains relationships between different `connections`
### Search before asking

### What happened

image

This kind of relationship should not exist

### How to reproduce

Not sure

### Anything else

_No response_

### Version

main

### Code of Conduct

warren830:
# Summary

Previously, we select * from `_tool_jira_issues` and join `board_issues` with condition `_tool_jira_board_issues.connction_id = ?`
This will cause us to get all data from _tool_jira_issues with different connection_id but same issue_id.
So we change the condition to `_tool_jira_issues.connection_id = ?`

closes #2386

### Screenshots
image
image

### Other Information


plugins/jira/tasks/issue_convertor.go
```diff
@@ -50,7 +50,7 @@ func ConvertIssues(taskCtx core.SubTaskContext) error {
 		dal.From(jiraIssue),
 		dal.Join("left join _tool_jira_board_issues on _tool_jira_board_issues.issue_id = _tool_jira_issues.issue_id"),
 		dal.Where(
-			"_tool_jira_board_issues.connection_id = ? AND _tool_jira_board_issues.board_id = ?",
+			"_tool_jira_issues.connection_id = ? AND _tool_jira_board_issues.board_id = ?",
```
klesh, line 53: This doesn't seem legit.
We should modify the `Join` condition rather than `Where` condition

warren830: fixed

klesh (approved):
LGTM

Issue #2268, opened 2022-06-20
[GitLab] table.issues and pull_requests lack creator_id, author_id, etc.


Startrekzky:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

After the collection of GitLab data, the domain layer seemed not so 

- [x] table.issues lacks the values of `creator_id`, `creator_name`, and `assignee_id` resolved by #2396 
- [ ] table.pull_requests lacks the values of `author_id`
- [ ] mr_notes in the tool layer doesn't show in the domain layer table pull_request_comments

### What you expected to happen

- [ ] `creator_id`, `creator_name`, and `assignee_id` have values in table.issues. `creator_id` is the FK to table.accounts  (table.users before v0.12)
- [ ] `author_id` has value, and is the FK  to table.accounts (table.users before v0.12)
- [ ] mr_notes from GitLab appeared in table.pull_request_comments
- [ ] add a column `type` in domain layer's table.pull_request_comments, please see the value assignment logic in the section 'Definition of GitLab comments'.

### How to reproduce

1. setup DevLake v0.11
2. collect any GitLab repo

### Definition of GitLab comments
1. Gitlab consists of 2 types of comments:
   a. normal comments: type = null, system = false 
     ![image](https://user-images.githubusercontent.com/14050754/175992626-a21aaac5-7035-46c2-beda-bfc0badbf6f0.png)
   b. diff comments: type = diffNote, system = false
     ![image](https://user-images.githubusercontent.com/14050754/175992767-16d842f0-8507-402a-9573-0b3fb136ab50.png)

2. For values of the `type` field in table.pull_request_comments
   a. for normal comments: if it can't be set to null then empty string
   b. for diff comments: DIFF
   c. for review comments (exist in GitHub but not GitLab, see pic below): REVIEW
     ![image](https://user-images.githubusercontent.com/14050754/175993786-7e61c226-e288-4d82-b00a-bb29829a4022.png)

### Anything else

related to #2126

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

warren830:
This is a test @warren830

klesh:
@warren830 can u receive the notification email from githbub

thenicetgp:
[...]

Issue #2126, opened 2022-06-08
[Bug] [Gitlab Plugin] Gitlab users are not appeared in table.users


Startrekzky:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

I set up DevLake v0.11-test version and imported Gitlab and Jira data. 
When I tried to get Gitlab issue creator's detail by joining table.users. I didn't get the full user info, such as organization. Thus, I couldn't tell if an issue is created by a community member.

### What you expected to happen

GitLab users appeared in table.users.

### How to reproduce

1. Set up DevLake main
2. Collected Gitlab data
3. Connected to MySQL
4. Run query
`select distinct i.creator_id, i.creator_name, u.id, u.name FROM issues i
left join users u on i.creator_id = u.id`

### Anything else

It's related to #2122

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #2396, opened 2022-06-30
fix: gitlab users are not appeared in table.users

abeizn:
# Summary
add gitlab user collector, extractor and convertor.
update gitlab user model
update e2e test

related to #2126 
related to #2268 

### Screenshots
image

### Other Information


plugins/gitlab/tasks/user_collector.go
```diff
@@ -0,0 +1,72 @@
...
+
+func CollectUsers(taskCtx core.SubTaskContext) error {
+	rawDataSubTaskArgs, data := CreateRawDataSubTaskArgs(taskCtx, RAW_USER_TABLE)
+	logger := taskCtx.GetLogger()
+	logger.Info("collect gitlab users")
+
+	collector, err := helper.NewApiCollector(helper.ApiCollectorArgs{
+		RawDataSubTaskArgs: *rawDataSubTaskArgs,
+		ApiClient:          data.ApiClient,
+		UrlTemplate:        "/projects/{{ .Params.ProjectId }}/members",
```
likyh, line 46: Maybe users are not limited in members. Is this how it is planned?

abeizn: yes, I think everyone who is invited is an organizer, but the organizer determined by the company still needs to be determined by the company itself

plugins/gitlab/models/migrationscripts/archived/user.go
```diff
@@ -20,9 +20,16 @@ package archived
 import "github.com/apache/incubator-devlake/models/migrationscripts/archived"
 
 type GitlabUser struct {
-	ConnectionId uint64 `gorm:"primaryKey"`
-	Email        string `gorm:"primaryKey;type:varchar(255)"`
-	Name         string `gorm:"type:varchar(255)"`
+	ConnectionId    uint64 `gorm:"primaryKey"`
+	ProjectId       int    `gorm:"primaryKey"`
```
klesh, line 24: This should be a Many-to-Many relationship

plugins/gitlab/models/user.go
```diff
@@ -17,12 +17,21 @@ limitations under the License.
...
+import (
+	"github.com/apache/incubator-devlake/models/common"
+)
 
 type GitlabUser struct {
-	ConnectionId uint64 `gorm:"primaryKey"`
-	Email        string `gorm:"primaryKey;type:varchar(255)"`
-	Name         string `gorm:"type:varchar(255)"`
+	ConnectionId    uint64 `gorm:"primaryKey"`
+	ProjectId       int    `gorm:"primaryKey"`
```
klesh, line 26: Same as above

plugins/gitlab/models/migrationscripts/archived/user.go
abeizn: I don't understand very well, like jira, tapd and gitlab should all be unified, right? This definition structure should be able to achieve all cases

plugins/gitlab/tasks/user_collector.go
```diff
@@ -0,0 +1,72 @@
...
+			return query, nil
+		},
+
+		ResponseParser: func(res *http.Response) ([]json.RawMessage, error) {
+			var result json.RawMessage
+			err := helper.UnmarshalResponse(res, &result)
+			if err != nil {
+				return nil, err
+			}
+			return []json.RawMessage{result}, nil
```
klesh, line 62: This doesn't seem right, by convention, we store `items` of a `page`, not the raw response body.
Even though it works in this particular case, it breaks the convention.

plugins/gitlab/tasks/user_extractor.go
```diff
@@ -0,0 +1,70 @@
...
+		RawDataSubTaskArgs: *rawDataSubTaskArgs,
+		Extract: func(row *helper.RawData) ([]interface{}, error) {
+			var userRes []models.GitlabUser
+			err := json.Unmarshal(row.Data, &userRes)
+			if err != nil {
+				return nil, err
+			}
+
+			results := make([]interface{}, 0)
+			for _, v := range userRes {
```
klesh, line 48: We should be expecting a single User in this callback, not all Users

klesh (commented):
I think we extract the Author/Assignees as User of the Issue as well.
https://github.com/apache/incubator-devlake/blob/42f6c3237d65e36bf009cec987aacbbc4b233d71/plugins/gitlab/tasks/issue_extractor.go#L54
https://github.com/apache/incubator-devlake/blob/42f6c3237d65e36bf009cec987aacbbc4b233d71/plugins/gitlab/tasks/issue_extractor.go#L65

plugins/gitlab/tasks/user_extractor.go
abeizn: updated

klesh (approved):
LGTM

Issue #1681, opened 2022-04-18
Jira plugin: collect the epic issue in an issue's epic link


Startrekzky:
## Description
Many Jira users use epic to represent 'strategy'. For example, any issue created under an epic is considered "Strategic", otherwise "Non strategic". With the epic issue, we can calculate the percentage of issues in each category: "Strategic", "Non-strategic", and "Bug fixing", so that users know the task allocation.

## Background
The current Jira plugin only collects issues by `board`. However, the epics that issues belonging to may not be collected. 
For instance, a user collected `Jira board 76` and got `issue-1`, `issue-2` and `issue3`. `Issue-1`'s epic link is issue 4 (an epic issue). As a result, issue-4 is not collected and will not appear in the `issues` table in the domain layer.

## Describe the solution you'd like
Jira plugin should also collect the epics that issues belonging to.

klesh:
1. collect all epic issues referenced by existing issues
2. used to showing epic detail, no need to build board/epic relationship

leglars:
@klesh @mindlesscloud  
Epic is just one of standard issue type in Jira (core). Jira installs an official plugin to provide the special ability for Epic out of box. This plugin also create some custom key within JIra and lock them. Though those key are customized, Epic has been a key feature of JIra, and we should seem those keys as original attributes of issue object.
Ideally, in my opinion, we can follow Jira's pattern, and store epic as normal issue, but add epic related keys into issue object.

klesh:
@leglars No problem, since we had already collected `jira_issues.epic_key`, what we left to do is to collect those `epic issues` and store them into the `jira_issues` table as well,  Am I understand it correctly? @Startrekzky

Startrekzky:
> @leglars No problem, since we had already collected `jira_issues.epic_key`, what we left to do is to collect those `epic issues` and store them into the `jira_issues` table as well, Do I understand it correctly? @Startrekzky

Correct. I guess the flow is:
1. Collect the epic issue when a Jira issue's epic_key is not empty.
2. Store the 'epic-type' issues should be stored in tables in all 3 layers: _raw_jira_api_issues, _tool_jira_issues, issues
   a. Make sure the epic issues are de-duplicated.
   b. Make sure the epic issues are enriched and converted just like other issues.

Please let me know if the flow needs to be polished. @klesh @keon94

keon94:
@klesh is this still slotted for the v0.12.0 release?

klesh:
[...]

PR #2401, opened 2022-06-30
[feat-1681][jira]: Collect the epic issue in an issue's epic link

keon94:

# Summary

#1681 

### Screenshots

### Other Information


plugins/jira/tasks/external_epic_collector.go
```diff
@@ -0,0 +1,152 @@
...
+			return data.Issues, nil
+		},
+	})
+	if err != nil {
+		return err
+	}
+	return collector.Execute()
+}
+
+func GetExternalEpicKeys(db dal.Dal, data *JiraTaskData) ([]string, error) {
```
keon94, line 118: I refactored this function out of the Collect function so I could test it from my e2e test

plugins/jira/e2e/external_epic_test.go
```diff
@@ -0,0 +1,83 @@
+package e2e
```
mindlesscloud, line 1: please add the Apache license header to this file

plugins/jira/tasks/external_epic_collector.go
```diff
@@ -0,0 +1,152 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tasks
+
+import (
```
mindlesscloud, line 20: would you mind merging the two import blocks?

plugins/jira/tasks/external_epic_collector.go
```diff
@@ -0,0 +1,152 @@
...
+	return collector.Execute()
+}
+
+func GetExternalEpicKeys(db dal.Dal, data *JiraTaskData) ([]string, error) {
+	// union of
+	// 1. issues with epics not from this board and not in the issues table
+	// 2. issues with epics not from this board that ARE already in the issues table (from previous runs)
+	// the above two selections are mutually exclusive
+	cursor, err := db.RawCursor(fmt.Sprintf(`
+			SELECT tji.epic_key as epicKey FROM _tool_jira_issues tji
```
klesh, line 124: 1. the `connection_id` is missing
2. second, why union? shouldn't it be 
```sql
select
	distinct epic_key
from
	_tool_jira_issues i
left join `_tool_jira_board_issues` bi on
	(i.connection_id = bi.connection_id
		and i.issue_id = bi.issue_id)
where
	i.epic_key != ''
	and bi.board_id = 8
```

plugins/jira/tasks/external_epic_extractor.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			if data.Options.TransformationRules.StoryPointField != "" {
+				strStoryPoint, _ := apiIssue.Fields.AllFields[data.Options.TransformationRules.StoryPointField].(string)
+				if strStoryPoint != "" {
+					issue.StoryPoint, _ = strconv.ParseFloat(strStoryPoint, 32)
+				}
+			}
+			issue.Type = mappings.typeIdMappings[issue.Type]
+			issue.StdStoryPoint = int64(issue.StoryPoint)
+			issue.StdType = mappings.stdTypeMappings[issue.Type]
```
klesh, line 94: I think it should be hardcoded as "EPIC"

mindlesscloud (commented):
it looks like the two new subtasks were not registered into `Jira.SubTaskMetas` .

plugins/jira/tasks/external_epic_extractor.go
```diff
@@ -0,0 +1,136 @@
...
+			if err != nil {
+				return nil, err
+			}
+			err = apiIssue.SetAllFields(row.Data)
+			if err != nil {
+				return nil, err
+			}
+			var results []interface{}
+			sprints, issue, worklogs, changelogs, changelogItems, users := apiIssue.ExtractEntities(data.Options.ConnectionId)
+			for _, sprintId := range sprints {
```
klesh, line 73: Can we share these code with jira_issue_collector?

plugins/jira/tasks/external_epic_collector.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return collector.Execute()
+}
+
+func GetExternalEpicKeys(db dal.Dal, data *JiraTaskData) ([]string, error) {
+	// union of
+	// 1. issues with epics not from this board and not in the issues table
+	// 2. issues with epics not from this board that ARE already in the issues table (from previous runs)
+	// the above two selections are mutually exclusive
+	cursor, err := db.RawCursor(fmt.Sprintf(`
```
mindlesscloud, line 123: it's dangerous and not recommended to use `fmt.Sprintf` to generate SQL. consider prepared statements which are supported by `db.RawCursor`. It looks like:
```go
db.RawCursor("SELECT * FROM _tool_jira_issues WHERE board_id = ? ", data.Options.BoardId)
```

plugins/jira/tasks/external_epic_collector.go
keon94: The union is needed to also account for external epics we already collected. They are gonna be in the issues table but not in the board_issues table. Also, the first select statement ensures that we are only grabbing epic keys that are not issue keys in the issues table. For instance EE-1 could be an epic key in one row, and an issue key in another. We wouldn't want to be picking that up. But I know it's an ugly query and can probably be reduced.

plugins/jira/tasks/external_epic_extractor.go
keon94: you mean jira_issue_extractor? Already asked [here](https://github.com/apache/incubator-devlake/pull/2401/files/bcb7b10f0e7496227269ddf1529eb6d39128b2bc#r927206808). I can do that by moving most of the extract logic intoa shared function

plugins/jira/tasks/external_epic_collector.go
keon94: We discussed offline - we'll simplify the query to Klesh's suggestion at the expense of pulling in extra epics for the sake of simplicity

plugins/jira/tasks/external_epic_collector.go
keon94: done

plugins/jira/tasks/external_epic_extractor.go
keon94: done

plugins/jira/e2e/epic_test.go
```diff
@@ -0,0 +1,82 @@
+package e2e
```
klesh, line 1: Please add ASF header

plugins/jira/tasks/epic_collector.go
```diff
@@ -0,0 +1,145 @@
...
+	"encoding/json"
+	"github.com/apache/incubator-devlake/plugins/helper"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+)
+
+const RAW_EPIC_TABLE = "jira_api_epics"
+
+type JiraEpicParams struct {
```
klesh, line 35: Why not use `JiraApiParams`?

plugins/jira/tasks/epic_collector.go
```diff
@@ -0,0 +1,145 @@
...
+	EntryPoint:       CollectEpics,
+	EnabledByDefault: true,
+	Description:      "collect Jira epics from all boards",
+	DomainTypes:      []string{core.DOMAIN_TYPE_TICKET, core.DOMAIN_TYPE_CROSS},
+}
+
+func CollectEpics(taskCtx core.SubTaskContext) error {
+	db := taskCtx.GetDal()
+	data := taskCtx.GetData().(*JiraTaskData)
+	externalEpicKeys, err := GetEpicKeys(db, data)
```
klesh, line 53: Loading all epic_keys is fine in most cases, but it could be a problem when the total number of epics grows too big. say 1000? Not sure if Jira API would take that many keys.
Besides, it is pointless to send all the keys every time, since the max records we can get is 100.
I suggest that we create an `Iterator` as the `Input` that returns 100 keys max each time.

keon94: Are you saying have a Dal iterator that returns up to 100 rows at a time? I'll have to make a batched iterator version of it, because the existing one gives just 1 row at a time. Also, how will the urlTemplate look like? The jql is a query param, so I dont know what the template syntax is going to become like. I couldnt find existing examples of us doing something like this

klesh: Yes, here is what I think it should be
1. The `Iterator.Fetch` is returning a value of type `interface{}`, so, we may return `[]int` or `string` that contains up to 100 epic-keys at a time.
2. Within `ApiCollector.Query` callback, we can access those `epic-keys` via `reqData.Input`, and concatenate them to the jql parameter

plugins/helper/queue.go
```diff
@@ -23,8 +23,9 @@ import (
 )
```
keon94, line 23: I was using queue iterators before realizing the Dal iterator was what I needed. I did find some issues and incompleteness with the QueueNode code though that I've tried to improve (e.g. being able to access a node's data). @klesh let me know what you think or if you want it reverted.

klesh (approved):
LGTM

PR #2416, opened 2022-07-05
feat: [POC] add since in framework

likyh:
# Summary

That's the POC for simplifying plugin creation.

## What's New:

1. manage `since` in the framework.
2. merge collector and extractor.

## Basis of modification:
1. now collectors are matched with extractors, respectively.
3. now extractors are run right after corresponding collectors.
4. now, the var `since` is queried from the latest model, and it doesn't care about the missing model older than the latest one.
5. a collector and the corresponding collector have the same `RawDataSubTaskArgs`.
6. a collector and the corresponding collector have the similar `ExtractApiIssuesMeta`.


plugins/github/tasks/issue_extractor.go
```diff
@@ -213,7 +214,7 @@ func ExtractApiIssues(taskCtx core.SubTaskContext) error {
 			}
 			results = append(results, githubIssue)
 
-			return results, nil
+			return results, &githubIssue.GithubUpdatedAt, nil
```
warren830: What to do next after returning this `&githubIssue.GithubUpdatedAt`?
 Do we save this to DB or just use a struct to store it in the memory?

likyh: Maybe we can create a new meta table to save it.

keon94:
If we're going to merge collectors and extractors, is there a need for raw tables anymore?

plugins/helper/api_collector.go
```diff
@@ -75,6 +77,8 @@ type ApiCollectorArgs struct {
 	Concurrency    int
 	ResponseParser func(res *http.Response) ([]json.RawMessage, error)
 	AfterResponse  common.ApiClientAfterResponse
+
+	Extract func(row *RawData) (models []interface{}, latestUpdated *time.Time, err error)
```
keon94, line 81: it's a matter of preference to some degree, but I suggest we wrap the returned _models_ and _latestUpdated_ in a struct to logically group the returned arguments. It also enables us to easily extend the # of params returned if needed later with minimal code change

plugins/github/tasks/issue_extractor.go
```diff
@@ -59,170 +45,6 @@ type IssuesResponse struct {
...
-		},
-	})
-
-	if err != nil {
-		return err
-	}
-
-	return extractor.Execute()
-}
 func convertGithubIssue(issue *IssuesResponse, connectionId uint64, repositoryId int) (*models.GithubIssue, error) {
```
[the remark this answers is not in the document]
likyh: yeah~ you are right.

plugins/helper/api_collector.go
likyh: Good idea~

likyh:
> If we're going to merge collectors and extractors, is there a need for raw tables anymore?

I think it should be kept. It is very useful when developing and debugging.

PR #2451, opened 2022-07-08
feat: add ae swagger support

### LINKED ISSUE #2450: [Doc][ae] swagger support for AE plugin
## Documentation Scope
Sub task of #2449 

## Describe the Change
AE APIs should be display on swagger page

klesh:
# Summary
As title suggested

Closes #2450 

### Screenshots
image


api/docs/stub.go
```diff
@@ -0,0 +1 @@
+package docs
```
abeizn, line 1: Missed apache header

klesh:
> There's also a mistake in the swagger docs [here](https://github.com/apache/incubator-devlake/blob/e5830611b7b32e8b35faec32180c07a39b1291f7/api/pipelines/pipelines.go#L54). The parameter type should be models.NewPipeline, not string

Nice catch, but it is out of the scope, would you file a bug for it?

mindlesscloud (approved):

Issue #2433, opened 2022-07-06
[Feature][framework] record subtask execution time


klesh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Currently, we have `task.spendSeconds` to record the execution time of a `task`.
However, the execution time of `subtask` is missing. It would be beneficial to have this information recorded.

### Use case

This information can help identify the bottleneck of a pipeline
1. For developers: to improve the performance
2. For users: to skip some unneeded subtasks

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

keon94:
@klesh how do you want the duration recorded? 
1. A new table for subtasks - i.e. _devlake_subtasks (I think this is what you want)
2. Reuse the tasks table, and just write the time info into the subtasks Json
3. Just simply log the elapsed time after a subtask is done (write nothing to DB)

klesh:
Yes, you are correct.
I would prefer the First option because it would be easier for us to
analyze in the future. Say we like to know which subtasks ran slower
than others.

PR #2455, opened 2022-07-08
[feat-2433]: record execution time for plugin subtasks

keon94:

# Summary

#2433 

### Screenshots
included below.

### Other Information

### DIFF
--- a/runner/run_task.go
+++ b/runner/run_task.go
@@ -236,10 +236,6 @@ func RunPluginSubTasks(
 	// execute subtasks in order
 	taskCtx.SetProgress(0, steps)
 	subtaskNumber := 0
-	var subtasks []*models.Subtask
-	defer func() {
-		recordSubtasks(logger, db, subtasks)
-	}()
 	for _, subtaskMeta := range subtaskMetas {
 		subtaskCtx, err := taskCtx.SubTaskContext(subtaskMeta.Name)
 		if err != nil {
@@ -261,8 +257,7 @@ func RunPluginSubTasks(
 				SubTaskNumber: subtaskNumber,
 			}
 		}
-		subtask, err := runSubtask(taskID, subtaskNumber, subtaskCtx, subtaskMeta.EntryPoint)
-		subtasks = append(subtasks, subtask)
+		err = runSubtask(logger, db, taskID, subtaskNumber, subtaskCtx, subtaskMeta.EntryPoint)
 		if err != nil {
 			return &errors.SubTaskError{
 				SubTaskName: subtaskMeta.Name,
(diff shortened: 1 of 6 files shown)


models/migrationscripts/subtask_07112022.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	commonArchived "github.com/apache/incubator-devlake/models/migrationscripts/archived"
+	"gorm.io/gorm"
+	"time"
+)
+
+type Subtask_07112022 struct {
+	commonArchived.Model
+	ParentID     uint64     `json:"parent_id" gorm:"index"`
```
klesh, line 12: `TaskId` would be more appropriate, `ParentId` is for representing a Tree by convention.

keon94:
Screenshot of the new subtasks table after a manual test

image

models/migrationscripts/subtask_07112022.go
```diff
@@ -0,0 +1,35 @@
+package migrationscripts
+
+import (
+	"context"
+	commonArchived "github.com/apache/incubator-devlake/models/migrationscripts/archived"
+	"gorm.io/gorm"
+	"time"
+)
+
+// Subtask_07112022 DB snapshot model of models.Subtask
```
keon94, line 10: @klesh is this still the correct way of managing migrations? I see recently there was a big refactor and everything is using init_schema.go now as opposed to individual migrations

klesh: We added MultiConnection for all other plugins in v0.12. With this `connection` table added, every single table of the plugin was affected, as well as the Domain Layer tables, we have to flush the no-longer-valid data to maintain the data integrity, So, might as well drop existing migration scripts and rebuild the tables.
However, the Framework migration should be not affected, we should keep the migration scripts as it was, with that said, you are doing the right thing.

runner/run_task.go
```diff
@@ -292,3 +301,28 @@ func UpdateProgressDetail(db *gorm.DB, logger core.Logger, taskId uint64, progre
...
+	beginAt := time.Now()
+	subtask := &models.Subtask{
+		Name:    ctx.GetName(),
+		TaskID:  parentID,
+		BeganAt: &beginAt,
+	}
+	defer func() {
+		finishedAt := time.Now()
+		subtask.FinishedAt = &finishedAt
+		subtask.SpentSeconds = finishedAt.Unix() - beginAt.Unix()
```
klesh, line 319: I think we should save the `subtask` right here, it makes more sense than doing it in the caller function.
And, we should save the `subtaskNumber` as a reference as well.

keon94: I can do that but I was trying to save DB calls here since this is being called in a loop. With this code we just make a single call no matter how many subtasks we have.

klesh: It is ok because each iteration took quite some time to finish, we would like the information to be saved into the database as soon as the subtask is finished, or it would be gone if any of these subtasks went south.

keon94: Made the change

klesh (approved):
LGTM

PR #2486, opened 2022-07-13
fix: gitlab blueprint generator

klesh:
# Summary

1. fix typo in the plugin name
2. collect/extract/convert commits should be disabled by default.


warren830 (approved):
LGTM

PR #2496, opened 2022-07-14
force connect with flawed https cert

### LINKED ISSUE #2370: [Feature] force connection with Flawed HTTPS certificate
### Search before asking

### Description

With the HTTPS certificate expires.
The request will return an error as "x509: certificate has expired or is not yet valid:"
image

But in sometimes, it is hard to update the https certificate and beacuse of the Intranet or another reason.It is not necessary for user to ensure certificate is all right.

So it is better to have a setting or config to force connection without HTTPS  certificate check.

### Use case

Use a setting or config.
When we have to collect a data source with Flawed HTTPS certificate,there are no error sent out.

### Related issues

_No response_

### Code of Conduct

mappjzc:
# Summary
Add insecureSkipVerify in client Transport for api_client.
Add env IN_SECURE_SKIP_VERIFY to set if force to connect and skip the verify

close #2370 

### Screenshots
image
image

### Other Information


klesh (approved):
LGTM

PR #2500, opened 2022-07-15
Db Migration confirmation

### LINKED ISSUE #2440: [Feature][framework] DevLake should acquire user's consent before deleting data
### Search before asking

### Description

During v0.12 development, have added `connection` entity to many plugins which previously support only single connection, which effectively invalidates whatever data exists in the database. So, we all agreed that it is safely to purge those data to maintain data integrity.

But it poses another problem here: We are going to delete data silently, and it may be undesirable for some users. It is important for our users to know what is going to happen at what cost, and offer them a chance to stop the Migration.

I propose that we should do the following steps to make it happen:

1. Modify existing `GET /ping` API from BE, when BE detects Migration is happening, disable all other APIs, and all APIs  including `GET /ping` should return some kind of 428 error response like the following:
  ```
  {
        "success": false,
        "message": "System halt due to Pending Database Migration was detected, it might purge expired data and cripple your dashborads. Send a Post request to /status/migrate if that is ok for you"
  }
  ```
2. Add `POST /proceeed-db-migration` to BE, when it is called, migrate the database, and enable all APIs
3. Config-UI to adopt the API, and prompt users accordingly
4. Accept a setting `FORCE_MIGRATION` from configuration, when `FORCE_MIGRATION=true`, proceed without warning and confirmation.

### Use case

1. Users may know what is the stake
2. Users may have a chance to stop Migration if they have an extensive database and do not want to drop data for the time being. In this case, users should modify the `image tag` back to the previous version.

### Related issues

_No response_

### Code of Conduct

klesh:
# Summary

1. All APIs stop working and a 428 error would be returned if there were pending migrations
2. Users must request `GET /proceed-db-migration` endpoint to apply the migration, or revert the image tags to use the existing version.
3. No confirmation is needed if "FORCE_MIGRATION=true" is found in `.env` or Environment Variable

Close #2440

### Screenshots

### Other Information

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ swag:
 	swag init --parseDependency --parseInternal -o ./api/docs -g ./api/api.go -g plugins/*/api/*.go
 	echo "visit the swagger document on http://localhost:8080/swagger/index.html"
 
-dev: build-plugin swag run
+dev: build-plugin run
 
 debug: build-plugin-debug
 	dlv debug main.go
--- a/api/api.go
+++ b/api/api.go
@@ -18,12 +18,16 @@ limitations under the License.
 package api
 
 import (
+	"fmt"
+	"net/http"
 	"time"
 
 	_ "github.com/apache/incubator-devlake/api/docs"
+	"github.com/apache/incubator-devlake/api/shared"
 	"github.com/apache/incubator-devlake/logger"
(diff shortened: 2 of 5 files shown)


mindlesscloud (approved):
LGTM

PR #2504, opened 2022-07-15
feat: add relationship with commit files and its components

### LINKED ISSUE #953: Extract file-level data from git repository
## User story
As a tech lead, I want to see the bug distribution in low-level components (from code files), as a guide for the future framework evolution.

## Goal
The key is to match `bug` to `component`. 
- If we want to find the overall distribution of all bugs, we can do: `issues(bug)` join `prs` join `commits` join `commit_files(component)`;
- If we want to find the overall distribution of bugs fixed in a specific tag, we can do: `ref_diff_issues(bug)` join `prs` join `commits` join `commit_files(component)`;

## Proposed Solution
- [ ] Add a `commit_files` table in the domain layer
    | field                         | type       | length    | description | key  |                                                                                                                                              
    | -------------          | -------- | -------- | ----------- | -----|
    | `commit_sha`        | char       | 40          |                     | FK    |
    | `file_name`            |  varchar | 255        |                     |         |
    | `status`                  |  varchar | 255        |                     |         |
    | `additions`             |  int         |                |                     |         |
    | `deletions`             |  int         |                |                     |         |
    | `changes`              |  int         |                |                     |         |
    | `file_component`  |  varchar | 255        |                     |         |
    
    a. all the columns except `file_component` can be retrieved from [gitExtractor](https://github.com/apache/incubator-devlake/tree/main/plugins/refdiff). **Note: not from [Github API](https://docs.github.com/en/rest/commits/commits#get-a-commit) or Gitlab API.**
    b. how to assign `file_component` value to a commit file? It varies in different programming languages. 
          - In Go, a component can be found by folder names; 
          - while in other languages, it might need user input to define the `file_component`.

- [ ] Allow users to define the `file_component` in `.env` by REGEX

## An example of commit files
For instance, this [commit](https://github.com/apache/incubator-devlake/pull/2039/commits/aeec784efc4c21e9e92a414e3008987cb7f791fa) touches 4 files:
image

## Belongs to issue
This is a sub-issue of both #949 and #1771

## Additional context
N/A

xgdyp:
# Summary
This PR is related to issue953. 
First, I create a table `component` and user can input file component relationship by operating this table including component name ,repoid,component regexp
like
image
Then I changed the primarykey in commit_files table  from (commit_sha,filepath) to id(commit_sha+":"+filepath)
like:
image
Then, I create a new table to records which component every commit_file_id belongs to.
image

When user run a pipeline  using gitextractor plugin, lake gets the relationship in components table and match by regexp. Then update the commit_file_component during storing commit_files

closed #953

### Other Information
Really thanks to Klesh for the guidance TwT！


klesh:
@xgdyp Please fix the commit message by following https://www.conventionalcommits.org/en/v1.0.0/#summary

models/domainlayer/code/commit.go
```diff
@@ -46,12 +46,33 @@ func (Commit) TableName() string {
 
 type CommitFile struct {
 	common.NoPKModel
-	CommitSha string `gorm:"primaryKey;type:varchar(40)"`
-	FilePath  string `gorm:"primaryKey;type:varchar(255)"`
+	ID        string `gorm:"primaryKey;type:varchar(255)"`
```
klesh, line 49: Please use the `domainlayer.DomainEntity` instead, check other domain layer entities for reference.

models/domainlayer/code/commit.go
```diff
@@ -46,12 +46,33 @@ func (Commit) TableName() string {
...
 	Deletions int
 }
 
 func (CommitFile) TableName() string {
 	return "commit_files"
 }
+
+type Component struct {
+	RepoId    string `gorm:"primaryKey;type:varchar(255)"`
+	Name      string `gorm:"type:varchar(255)"`
```
klesh, line 62: I think we both agreed that the PrimaryKey should be the `Name`.
The `RepoId` alone is not a valid PK for the table

models/domainlayer/code/commit.go
```diff
@@ -46,12 +46,33 @@ func (Commit) TableName() string {
...
+}
+
+func (Component) TableName() string {
+	return "components"
+}
+
+type CommitFileComponent struct {
+	common.NoPKModel
+	CommitFileID string `gorm:"primaryKey;type:varchar(255)"`
+	Component    string `gorm:"type:varchar(255)"`
```
klesh, line 73: It should be `ComponentName` if we agreed that `Name` is the PK of `Components` table.

models/domainlayer/code/commit.go
```diff
@@ -46,12 +46,33 @@ func (Commit) TableName() string {
...
+	PathRegex string `gorm:"type:varchar(255)"`
+}
+
+func (Component) TableName() string {
+	return "components"
+}
+
+type CommitFileComponent struct {
+	common.NoPKModel
+	CommitFileID string `gorm:"primaryKey;type:varchar(255)"`
```
klesh, line 72: This should b `CommitFileId` if we nested `domainlayer.DomainEntity` instead of `common.NoPKModel` within `CommitFiles`

xgdyp:
> @xgdyp Please fix the commit message by following https://www.conventionalcommits.org/en/v1.0.0/#summary

OK, I think it will squash it.

models/migrationscripts/register.go
```diff
@@ -27,6 +27,7 @@ func All() []migration.Script {
 		new(updateSchemas20220601),
 		new(updateSchemas20220616),
 		new(blueprintNormalMode),
+		new(updateSchemas20220711),
```
klesh, line 30: I would suggest that we give it a meaningful name like `commitfile_component`

models/migrationscripts/updateSchemas20220711.go
```diff
@@ -0,0 +1,77 @@
...
+
+package migrationscripts
+
+import (
+	"context"
+	"github.com/apache/incubator-devlake/models/common"
+	"gorm.io/gorm"
+)
+
+type Component struct {
```
klesh, line 26: the migration script should be updated if we agreed to the above comments

plugins/gitextractor/models/interface.go
```diff
@@ -27,5 +27,6 @@ type Store interface {
 	Refs(ref *code.Ref) error
 	CommitFiles(file *code.CommitFile) error
 	CommitParents(pp []*code.CommitParent) error
+	FileComponent(component *code.CommitFileComponent) error
```
klesh, line 30: Lets name it `CommitFileComponents`

plugins/gitextractor/store/csv.go
```diff
@@ -132,6 +132,9 @@ func (c *CsvStore) Refs(ref *code.Ref) error {
 func (c *CsvStore) CommitFiles(file *code.CommitFile) error {
 	return c.commitFileWriter.Write(file)
 }
+func (c *CsvStore) FileComponent(component *code.CommitFileComponent) error {
```
klesh, line 135: the variable name and its type are uncorrelated... this is very confusing. is it a `component` or what? 😂.
and insert an empty line between functions pls.

plugins/gitextractor/store/database.go
```diff
@@ -91,6 +91,14 @@ func (d *Database) CommitFiles(file *code.CommitFile) error {
 	return batch.Add(file)
 }
 
+func (d *Database) FileComponent(commitfile *code.CommitFileComponent) error {
```
klesh, line 94: same as above

xgdyp:
> > > @xgdyp Please fix the commit message by following https://www.conventionalcommits.org/en/v1.0.0/#summary
> > 
> > 
> > OK, I think it will squash it.
> 
> Yes, we will. But the original commit messages will be kept inside that squashed commit.

I'll fix it later

xgdyp:
I fix it, please review it, thanks! @klesh

klesh:
@xgdyp Nice work. However, there is a conflict that needs to be dealt with.
image

models/domainlayer/code/commit.go
```diff
@@ -45,13 +46,33 @@ func (Commit) TableName() string {
...
+	FilePath  string `gorm:"type:varchar(255)"`
 	Additions int
 	Deletions int
 }
 
 func (CommitFile) TableName() string {
 	return "commit_files"
 }
+
+type Component struct {
```
klesh, line 60: It would be better to move `Component` definition to an independent file since it is not part of the `commit` structure, nor do they highly coupled.

xgdyp:
> @xgdyp Nice work. However, there is a conflict that needs to be dealt with. image

I'll resolve it now

models/domainlayer/code/commit.go
xgdyp: I think it can be replaced in gitextractor plugin's model file, what do you think?

klesh: No, just created a new file `models/domainlayer/code/component.go`, and move this struct into it and we are good.

xgdyp: OK!

xgdyp:
hi Klesh, please review again@Klesh

models/migrationscripts/20220721_commitfile_component.go
```diff
@@ -0,0 +1,77 @@
...
+	err := db.Migrator().AutoMigrate(Component{}, CommitFile{}, CommitFileComponent{})
+	if err != nil {
+		return err
+	}
+	return nil
+
+}
+
+func (*commitfileComponent) Version() uint64 {
+	return 202207211505
```
abeizn, line 72: Please keep 14 digits

models/migrationscripts/20220721_commitfile_component.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (CommitFileComponent) TableName() string {
+	return "commit_file_components"
+}
+
+type commitfileComponent struct{}
+
+func (*commitfileComponent) Up(ctx context.Context, db *gorm.DB) error {
+
+	err := db.Migrator().AutoMigrate(Component{}, CommitFile{}, CommitFileComponent{})
```
abeizn, line 63: please drop the `commit_files` table first, then add it,  because it is not a new table.

abeizn: otherwise, when it upgrades, the commit_files is not right.
image

xgdyp: I see

abeizn: It was really needed before, but this v0.12 is not compatible with the previous version, so you can be lazy, lol~~

xgdyp: thanks!

xgdyp: I fix it ,please review again, thanks!

abeizn: @xgdyp  I tested it, great finish.

abeizn (approved):
LGTM

PR #2512, opened 2022-07-16
fix: partially false statement removed

nikitakoselev:
Installing DevLake is not always "easy", so instead we shall let people know that community is there to support you.

image


warren830:
Thank you for your suggestions. Can you give use some advice which can help users deploy devlake easily?

klesh (approved):
LGTM

PR #2523, opened 2022-07-18
Sundry/some problems in config UI

likyh:
# Summary

1. use correct scope for plugins image
2. fix "enable refdiff in github" 
image
3. fix Clicking incident but focus on bug input. 
image
4. fix the bug that setting github config will clear other config


klesh (approved):
LGTM

PR #2525, opened 2022-07-18
Sundry/fix 3 small issues

likyh:
# Summary

1. not request jira issue when setting other plugins. 
image
2. hide a no-response button "View Schedule"  
image
3. fix cannot refresh auto after creating blueprint and pipeline 
image

### DIFF
--- a/config-ui/src/hooks/useJIRA.jsx
+++ b/config-ui/src/hooks/useJIRA.jsx
@@ -18,6 +18,7 @@
 import { useEffect, useState, useCallback } from 'react'
 import request from '@/utils/request'
 import { ToastNotification } from '@/components/Toast'
+import { Providers } from "@/data/Providers";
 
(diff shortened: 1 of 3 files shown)


warren830 (approved):
LGTM

config-ui/src/hooks/useJIRA.jsx
```diff
@@ -30,6 +31,9 @@ const useJIRA = ({ apiProxyPath, issuesEndpoint, fieldsEndpoint, boardsEndpoint
   const [error, setError] = useState()
 
   const fetchIssueTypes = useCallback(() => {
+    if (activeConnection.plugin !== Providers.JIRA) {
```
e2corporation, line 34: @warren830 @klesh This isn't the ideal way to do things... returning to escape async logic,  these calls should have been wrapped at a higher layer not at this layer....

PR #2526, opened 2022-07-18
fix: fix start_at/end_at display bug

likyh:
# Summary

1. fix start_at/end_at display bug
image
image
now start_at/end_at can display correctly when one of them or both of them are null

### DIFF
--- a/config-ui/src/pages/blueprints/blueprint-detail.jsx
+++ b/config-ui/src/pages/blueprints/blueprint-detail.jsx
@@ -664,15 +664,12 @@ const BlueprintDetail = (props) => {
         status: p.status,
         statusLabel: TaskStatusLabels[p.status],
         statusIcon: getTaskStatusIcon(p.status),
-        startedAt: dayjs(p.beganAt).format('L LTS'),
-        completedAt:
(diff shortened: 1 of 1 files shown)


config-ui/src/pages/blueprints/blueprint-detail.jsx
```diff
@@ -664,15 +664,12 @@ const BlueprintDetail = (props) => {
...
         statusIcon: getTaskStatusIcon(p.status),
-        startedAt: dayjs(p.beganAt).format('L LTS'),
-        completedAt:
-          p.status === 'TASK_RUNNING'
-            ? ' - '
-            : dayjs(p.finishedAt || p.updatedAt).format('L LTS'),
+        startedAt: p.beganAt ? dayjs(p.beganAt).format('L LTS') : '-',
+        completedAt: p.finishedAt ? dayjs(p.updatedAt).format('L LTS') : ' - ',
         duration:
-          p.status === 'TASK_RUNNING'
```
e2corporation, line 673: Duration calc is different from a Running task versus a completed/failed task. Why was the running check removed here?

PR #2527, opened 2022-07-18
feat: revert to check multi token in github plugin

likyh:
# Summary

Revert to check multi tokens in the GitHub plugin.

It's a temporary fix for "test connection" in the GitHub plugin.


warren830 (approved):
LGTM

Issue #2543, opened 2022-07-19
[Bug][jira] jira collectWorklogs lost records finished data in info


mappjzc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

![image](https://user-images.githubusercontent.com/2921251/179790332-1da06205-a222-4e0d-84ee-0ad9c92aa987.png)
at the collectAccounts and before
we can see the finished records info.

but when begin the collectWorklogs.
there was no finished records info anymore.

add test info for testing it
![image](https://user-images.githubusercontent.com/2921251/179800337-24191d34-f987-4292-ad5d-c3e1029360d7.png)
![image](https://user-images.githubusercontent.com/2921251/179800121-ea2a0d4d-a2ee-4b57-bb6e-4e5f25c717f0.png)

we can see it is because of the empty save not be write.

### What you expected to happen

show the detail info about progress to reduce user anxiety and help eliminate stuck conditions.

### How to reproduce

1.collect jira test board 73.
2.wait for collectWorklogs

### Version

main

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

mappjzc:
fixed by #2544

PR #2544, opened 2022-07-19
fix: add inc for lost

mappjzc:
# Summary

Add IncProgress to results range for ApiExtractor
Add IncProgress to ApiCollector when items was zero.

#2543 

### Screenshots
image


klesh (approved):
LGTM

PR #2552, opened 2022-07-20
fix: extract transtion manager to fix some bug

likyh:
# Summary

1. fix different connections that affect each other.
2. only save when clicking the button. (it's saved when inputting)
3. Fix the cursor will jump to the end of the text when inputting


e2corporation:
@likyh Thanks for the PR, as we discussed there are some very valuable changes proposed here, I especially like the addition of the Transformation Manager Hook, that will be useful in the future for better encapsulation and reduce the coupling concerns you raised earlier. Since my last service pack 1.0 PR for UI has resolved a few of the key issues that this ticket originally aimed to fix, we'll hold on merging for now.

We will collaborate on incorporating some of these ideas in a future ticket when the transformation logic is extracted.

PR #2569, opened 2022-07-21
feat: run pipeline in independent process

### LINKED ISSUE #2473: [Feature][framework] Throttling total number of pipelines running in parallel
### Search before asking

### Description

Currently, the pipelines are launched right after their creation, this leads to the following concerns:

1. Users might want to limit the number of running pipelines to avoid race conditions.
2. Users might want to limit the resource that DevLake is consuming.
3. Running too many pipelines simultaneously might be too heavy for the IO system, which increases the probability of timeout exception.

@CamilleTeruel @hezyin Sending out requests in a short burst increases the probability of timeout exception.

### Use case

1. Some users wanted to limit the number of pipelines running in parallel to exactly 1, so they could dump tables from our MySQL database to Starrocks without worrying about race-condition
2. Users might create pipelines as many as they need without crashing the system
3. Users could adjust the number to balance the speed and stability

### Related issues

_No response_

### Code of Conduct

likyh:
# Summary

run pipeline in an independent process.

WARNING:
It means NO LIMIT when DevLake run without env `PIPELINE_MAX_PARALLEL`, and it's not suggested.

close #2473


klesh:
@narrowizard

services/blueprint.go
```diff
@@ -312,5 +311,5 @@ func TriggerBlueprint(id uint64) (*models.Pipeline, error) {
 	}
 	pipeline, err := createAndRunPipelineByBlueprint(blueprint.ID, blueprint.Name, plan)
```
klesh, line 312: `createAndRunPipelineByBlueprint` should be renamed as well

services/pipeline.go
```diff
@@ -77,9 +78,16 @@ func pipelineServiceInit() {
 	if err != nil {
 		panic(err)
 	}
+
+	var pipelineMaxParallel = cfg.GetInt64("PIPELINE_MAX_PARALLEL")
+	if pipelineMaxParallel <= 0 {
```
klesh, line 83: We should allow `0`, and it should means "No Limit"

services/pipeline.go
```diff
@@ -183,8 +191,40 @@ func GetPipeline(pipelineId uint64) (*models.Pipeline, error) {
...
+			if pipeline.ID != 0 {
+				break
+			}
+			time.Sleep(time.Second)
+		}
+		startedPipelineIds = append(startedPipelineIds, pipeline.ID)
+		go func() {
+			pipelineLog.Info("run pipeline, %d", pipeline.ID)
+			_ = runPipeline(pipeline.ID)
+			defer sema.Release(1)
```
klesh, line 221: This should be moved up to the top of the go-routine

services/blueprint.go
likyh: createPipelineByBlueprint

services/pipeline.go
likyh: ok

services/pipeline.go
likyh: ok, now pipelineMaxParallel=0 means NO LIMIT by set pipelineMaxParallel = 10000. The reason of setting 10000 is avoid too many `if` in RunPipelineInQueue

klesh (approved):
LGTM

Issue #2550, opened 2022-07-20
[Bug][jira plugin] devlake collection jira server information


jinzhu2002:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

lake-v0110-devlake-1    | time="2022-07-20 01:19:53" level=error msg=" [task #24] [jira] retry #2 for http code error[404]:[{"errorMessages":["用户“aa@bbb.com”不存在"],"errors":{}}]"
lake-v0110-devlake-1    | time="2022-07-20 01:19:54" level=error msg=" [task #24] [jira] retry #2 for http code error[404]:[{"errorMessages":["用户“cc@bbb.com”不存在"],"errors":{}}]"
lake-v0110-devlake-1    | time="2022-07-20 01:19:56" level=error msg=" [task #24] [jira] retry #2 for http code error[404]:[{"errorMessages":["用户“ggg@bbb.com”不存在"],"errors":{}}]"

execute task fail

### What you expected to happen

collection jira server information success

### How to reproduce

curl http://127.0.0.1:8080/rest/agile/1.0/board/85/issue?expand=changelog&jql=ORDER+BY+created+ASC&maxResults=100&startAt=0

"changelog":{"startAt":0,"maxResults":13,"total":13,"histories":[{"id":"148308","author":{"self":"http://127.0.0.1:8080/rest/api/2/user?username=aaa","name":"aaa","key":"aaa","emailAddress":"aa@bbb.com","avatarUrls":{"48x48":"http://127.0.0.1:8080/secure/useravatar?ownerId=aaa&avatarId=11401","24x24":"http://127.0.0.1:8080/secure/useravatar?size=small&ownerId=aaa&avatarId=11401","16x16":"http://127.0.0.1::8080/secure/useravatar?size=xsmall&ownerId=aaa&avatarId=11401","32x32":"http://127.0.0.1:8080/secure/useravatar?size=medium&ownerId=aaa&avatarId=11401"},"displayName":"sss

no accountID tag

modify apiv2models/user.go

func (u *User) getAccountId() string {
	if u == nil {
		return ""
	}
	if u.AccountId != "" {
		return u.AccountId
	}
      //add this code,run success,no test for jira cloud
	if (u.AccountId == "") && (u.Name != "") {
		return u.Name
	}
	return u.EmailAddress
}

### Version

v0.11.0

### Are you willing to submit PR?

[...]

PR #2587, opened 2022-07-25
fix bug #2550

jinzhu2002:

- [ ] I have read through the [Contributing](https://devlake.apache.org/community/) Documentation & [PR Template](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- [ ] My code is has necessary documentation (if appropriate)
- [ ] I have added any relevant tests

# Summary

### Screenshots

### Other Information

### DIFF
--- a/plugins/jira/tasks/apiv2models/user.go
+++ b/plugins/jira/tasks/apiv2models/user.go
@@ -46,10 +46,6 @@ func (u *Account) getAccountId() string {
 	if u.AccountId != "" {
 		return u.AccountId
 	}
-	//TODO add this code,run success,no test for jira cloud
-	if u.Name != "" {
-		return u.Name
-	}
 	return u.EmailAddress
 }
 
--- a/plugins/jira/tasks/apiv2models/user.go
+++ b/plugins/jira/tasks/apiv2models/user.go
@@ -46,6 +46,10 @@ func (u *Account) getAccountId() string {
 	if u.AccountId != "" {
 		return u.AccountId
 	}
+	//TODO add this code,run success,no test for jira cloud
+	if u.Name != "" {
+		return u.Name
+	}
 	return u.EmailAddress
 }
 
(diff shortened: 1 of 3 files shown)


plugins/jira/tasks/apiv2models/user.go
```diff
@@ -46,6 +46,10 @@ func (u *Account) getAccountId() string {
 	if u.AccountId != "" {
 		return u.AccountId
 	}
+	//TODO add this code,run success,no test for jira cloud
```
keon94, line 49: why is this TODO?

jinzhu2002: Verify that the cloud version is OK

plugins/jira/tasks/apiv2models/user.go
```diff
@@ -46,6 +46,10 @@ func (u *Account) getAccountId() string {
...
+	if u.Name != "" {
+		return u.Name
```
mindlesscloud, line 51: The accountId is a primary key of table `_tool_jira_accounts`, it should be unique. How about we leave it alone and modify the Collector `CollectAccounts`. For the server version Jira, the endpoint `api/2/user/search` should be invoked instead of `api/2/user` which only accepts a username, not an email. Please be cautious about the response of the API `api/2/user/search`, which is an array, not an object. So the `ResponseParser` should also be adjusted accordingly.

jinzhu2002: OK , I'll try to modify CollectUsers

mindlesscloud: There is one difference between the response of the two endpoints. For `api/2/user`, it is an object, however, the response of `api/2/user/search` is an array. The collector's `ResponseParser` should be modified to handle both cases. The collector returns a slice of interface `[]interface{}`, and every item of the slice is a user object. There is no need to change the extractor for it always receives a single user object.

mindlesscloud (approved):
LGTM

PR #2600, opened 2022-07-25
[FIX]: Fixes for k8-deploy.

apujari-hippo:
# Summary
1. [FIX]: resolve network address with connection string. 
[error] failed to initialize database, got error Error 1045: Access denied for user 'mysql'@'x.x.x.x' (using password: YES)

1. [FIX]: Container port for config-ui

### DIFF
--- a/k8s-deploy.yaml
+++ b/k8s-deploy.yaml
@@ -152,7 +152,7 @@ spec:
         - name: config-ui
           image: mericodev/config-ui:latest
           ports:
-            - containerPort: 4000
+            - containerPort: 30004
           envFrom:
             - configMapRef:
                 name: devlake-config
--- a/k8s-deploy.yaml
+++ b/k8s-deploy.yaml
@@ -12,8 +12,8 @@
 #    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
-  
-  ---
+
+---
 apiVersion: v1
 kind: Namespace
 metadata:
@@ -106,11 +106,11 @@ spec:
                 name: devlake-config
           env:
             - name: GF_USERS_ALLOW_SIGN_UP
-              value: 'false'
+              value: "false"
             - name: GF_DASHBOARDS_JSON_ENABLED
-              value: 'true'
+              value: "true"
             - name: GF_LIVE_ALLOWED_ORIGINS
-              value: '*'
+              value: "*"
             - name: MYSQL_URL
               value: mysql:3306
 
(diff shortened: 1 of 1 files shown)


apujari-hippo:
@hezyin Will you please review this PR?

k8s-deploy.yaml
```diff
@@ -12,8 +12,8 @@
 #    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
-
```
hezyin, line 15: Indeed a bug, thanks for catching this.

k8s-deploy.yaml
```diff
@@ -152,7 +152,7 @@ spec:
         - name: config-ui
           image: mericodev/config-ui:latest
           ports:
-            - containerPort: 4000
+            - containerPort: 30004
```
hezyin, line 155: This container does listen on port 4000. And this `containerPort` value should match with `config-ui` service's `targetPort`, which is also 4000. So we don't want to change this.

k8s-deploy.yaml
```diff
@@ -206,7 +206,7 @@ spec:
             - containerPort: 8080
           env:
             - name: DB_URL
-              value: 'mysql://merico:merico@mysql:3306/lake?charset=utf8mb4&parseTime=True'
+              value: "merico:merico@tcp(mysql:3306)/lake?charset=utf8mb4&parseTime=True"
```
hezyin, line 209: The error you encountered earlier was due to using an old image. The new image that we switched to (v0.12.0-beta3) during our call would take the `mysql://` connection string without issue. Do you think you could try the original connection string with the new image again and let us know if the issue persists?

hezyin (changes requested):
@apujari-hippo Hi Aditya, thanks for submitting this PR! We really appreciate it. Please see my comments and let me know if you have any questions.

k8s-deploy.yaml
apujari-hippo: New image accepts the old DB string.

k8s-deploy.yaml
apujari-hippo: The service target port was not working. Hence we update it to use node port.
https://github.com/apache/incubator-devlake/blob/main/k8s-deploy.yaml#L175

k8s-deploy.yaml
hezyin: Great! In that case, we don't need to update line 209, could you please revert the change? Thanks.

k8s-deploy.yaml
hezyin: Hi Aditya, based on [Kubernetes API spec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#ports), the `containerPort` is primarily informational.

> Exposing a port here gives the system additional information about the network connections a container uses, but is primarily informational. Not specifying a port here DOES NOT prevent that port from being exposed. 

So even if we completely remove line 155, it probably won't affect our deployment. But its value should match with which port this container listens at, which is 4000. You can verify this in `config-ui`'s docker file: https://github.com/apache/incubator-devlake/blob/main/config-ui/Dockerfile#L43

abeizn:
All issues have been resolved and this pr will be closed

Issue #2609, opened 2022-07-26
[Doc][deployment] Add "how to replace mysql with RDS" for k8s-deploy


klesh:
## Documentation Scope
https://devlake.apache.org/docs/QuickStart/KubernetesSetup

## Describe the Change
Some users are concerned about using MySQL directly in production mode, which may lead to performance bottleneck and maintenance overhead.
Ideally, they may opt for RDS or other managed DBS, we should add document of how to do so for K8s setup

PR #2610, opened 2022-07-26
fix: grafana is reverse-proxied since v0.12

klesh:
# Summary

The `grafana` service is reverse-proxied by `config-ui` since v0.12

related to #2609


mindlesscloud (approved):
LGTM

Issue #2627, opened 2022-07-28
[Bug][mericodev dockerhub] "mericodev/devlake-dashboard:latest" throws error


ppborah:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

**Docker Pull Command used:** 
`docker pull mericodev/devlake-dashboard:latest`

**Response:** 
`Error response from daemon: manifest for mericodev/devlake-dashboard:latest not found: manifest unknown: manifest unknown`

### What you expected to happen

Docker should have pulled the `mericodev/devlake-dashboard:latest` image
- `latest` tag missing in the image
- Works fine if version is specified explicitly (ex: `v0.12.0-beta3`, which is the latest version)

### How to reproduce

`docker pull mericodev/devlake-dashboard` OR
`docker pull mericodev/devlake-dashboard:latest`

### Version:
NA

### Are you willing to submit PR?
NA

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

klesh:
Which docker-compose file were you using?

ppborah:
docker-compose.yml [from this repo(latest one), NOT v0.11.0]

klesh:
Hi, the `latest` tag is available now. Feel free to re-open if any problems

PR #2639, opened 2022-07-30
fix(k8s deploy): change mysql image version

warren830:
# Summary

mysql 8.0.26 can not be pulled anymore, so we should switch to 8.0.30

### Screenshots

### Other Information


nicoabie:
@warren830 changing the mysql image to 8:0:30 solves an issue where 8:0:26 cannot be installed to M1 with arm64 chip

8.0.26: Pulling from library/mysql
ERROR: no matching manifest for linux/arm64/v8 in the manifest list entries

could you split this PR? or I can create one with just that change.

nicoabie:
I created https://github.com/apache/incubator-devlake/pull/2946 which is a split of this PR

ppborah:
The `latest` tag was not working because it was missing, details [here](https://github.com/apache/incubator-devlake/issues/2627).
So now, we have only the mysql issue.

warren830:
> @warren830 changing the mysql image to 8:0:30 solves an issue where 8:0:26 cannot be installed to M1 with arm64 chip
> 
> 8.0.26: Pulling from library/mysql ERROR: no matching manifest for linux/arm64/v8 in the manifest list entries
> 
> could you split this PR? or I can create one with just that change.

Yes, you are right, thanks!

warren830:
> The `latest` tag was not working because it was missing, details here: #2627 So now, we have only the mysql issue.

@nicoabie created a pr #2946 , this fixes issues here, thanks to both of you @nicoabie @ppborah

ppborah:
Thanks @warren830 :)

PR #2654, opened 2022-08-01
fix: add NoPKModel to commit_parents

### LINKED ISSUE #2650: [Bug][GitExtractor] Some commits are missing during collectGitCommits
### Search before asking

### What happened

Hello everyone. I am using the latest main branch version and trying to collect gitlab commit data with Devlake. However, I found that sometimes some commits are not collected into the database after the pipeline run. Specifically, I have tested on 3 projects, each's outcome respectively:

1. all commits on 28 branches/tags are stored successfully
2. only the 7th latest commit (sha:7d7600..) on dev branch is missing. The commits on master branch are all stored
3. same as the second project, only the 7th latest commit on dev branch is missing

I'm not sure whether it is related to '7th latest commit' or the branches with name 'dev' or something else. But these are some things I tried and observed:

1. The first project also has a branch named 'dev' and all commits seems to be collected well
2. I submit a new commit on the second project and the 7d7600 commit (which is the 8th latest now) is still missing, and other commits are all collected.

This is how I query the commits on a specific branch:
`with res as (
SELECT
  @r AS _sha,
  (SELECT @r := parent_commit_sha FROM commit_parents WHERE commit_sha = _sha limit 1) AS parent_sha,
  @l := @l + 1 AS lvl 
FROM (SELECT @r := (SELECT commit_sha FROM refs WHERE id='gitlab:GitlabProject:1:175:origin/dev'), @l := 0) vars,
commit_parents h) 
SELECT _sha, parent_sha FROM res WHERE parent_sha IS NOT NULL;`

This is what I got:
image

It shows that there are only 7 commits on this branch, and the commit `7d7600..` are not found in `commit_parents` table so it is the end. However, the 7d7600 is also not found in `commits` table,
image

and the gitlab shows that there are more commits before 7d7600:
image

Moreover, I check the sha of the parent commit of 7d7600 from the gitlab (acc33d..), and I found that the commits after 7d7600 are all stored well in the tables:
image
...
image

So I think this commit is probably missing during the collection some how. This is the logs of the pipeline run for the second project: 
[[task #18] [gitextractor].log](https://github.com/apache/incubator-devlake/files/9231509/task.18.gitextractor.log)

I might did something wrong on this, thanks a lot for any ideas.

All commits on all branches of the project should be collected and stored.

### How to reproduce

Use the latest devlake, create a gitlab blueprint and run. 

### Anything else

_No response_

### Version

main

### Code of Conduct

mindlesscloud:
# Summary

fix #2650 ([Bug][GitExtractor] Some commits are missing during collectGitCommits)
When the first `commit` came in,
several lines of buggy code hid an error.
```go
if err != r.storeParentCommits(commitSha, commit) {
          return err
}
```
Which caused an interruption in the execution process. So the first `commit` can not be saved to the database.
It won't happen again from the second `commit`, due to the nature of `BatchSaveDivider.ForType`

close #2650 

### Screenshots

### Other Information

### DIFF
--- a/models/migrationscripts/20220801_add_NoPKModel_to_CommitParent.go
+++ b/models/migrationscripts/20220801_add_NoPKModel_to_CommitParent.go
@@ -15,16 +15,39 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package code
+package migrationscripts
 
-import "github.com/apache/incubator-devlake/models/common"
+import (
+	"context"
+	"github.com/apache/incubator-devlake/models/common"
+	"gorm.io/gorm"
+)
 
-type CommitParent struct {
+type commitParent struct {
 	common.NoPKModel
 	CommitSha       string `json:"commitSha" gorm:"primaryKey;type:varchar(40);comment:commit hash"`
 	ParentCommitSha string `json:"parentCommitSha" gorm:"primaryKey;type:varchar(40);comment:parent commit hash"`
 }
 
-func (CommitParent) TableName() string {
+func (commitParent) TableName() string {
 	return "commit_parents"
+}
+
+type addNoPKModelToCommitParent struct{}
+
+func (*addNoPKModelToCommitParent) Up(ctx context.Context, db *gorm.DB) error {
+	err := db.Migrator().AutoMigrate(&commitParent{})
+	if err != nil {
+		return err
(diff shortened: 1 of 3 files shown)


klesh (approved):
LGTM

PR #2630, opened 2022-07-28
`fix` config-ui: blueprints service pack 2.0

e2corporation:
### :package: Config-UI / Blueprints / Service Pack 2.0

- [x] **Feature** #2578
- [x] **Fix** Enhance Blueprint Details Stage Task
   - Add `Repository`, `Project ID`, & `Board ID` Context to Stage Task Name
- [x] **Feature** Add `Manual` Mode Filter to Blueprints Grid
- [x] **Bug** #2549
- [x] **Chore** Install `@file-saver` NPM Package
- [x] **Chore** Code Cleanups & Linting
- [x] Test Production Build

### Description

This PR applies additional hotfixes and enhancements to the Blueprints `v0.12.0` Release.

### Does this close any open issues?
#2621

### Screenshots
`[<PENDING>]`

Screen Shot 2022-07-28 at 9 34 17 AM

PR #2656, opened 2022-08-02
[feat-2578]: Download pipeline logging archives API

### LINKED ISSUE #2578: [Feature][framework] Users be able to download pipeline logging archives from config-ui
### Search before asking

## Objectives

1. Users should be able to set the path of a folder for logging file storage, by Environment Variable or `.env` file, namely `LOGGING_DIR`
2. All logging files of a specific `pipeline` should be placed in one folder, i.e. `PIPELINE_LOGGING_DIR=$LOGGING_DIR/pipeline-$PIPELINE_ID-$PIPELINE_CREATED_AT`
3. All logging files of `tasks` belongs to the `pipeline` should be in the `pipeline logging dir`, i.e. `$PIPELINE_LOGGING_DIR/task-$TASK_ID-$ROW-$COL-$PLUGIN.log`
4. Provide an API endpoint `/pipelines/:pipeline_id/logging.tar.gz` for user to download the archived logging folder of the pipeline

@keon94 @e2corporation 

### Use case

Users can download the logging archive and send it to us for debugging

### Related issues

_No response_

### Code of Conduct

keon94:

- [x] This PR is using a `label` (bug, feature etc.)

# Summary

Closes #2578 

### Screenshots
sample tar file for pipeline 58 using CURL
image

directory structure of logging directory in the container
image

Swagger doc:
image
FYI @e2corporation 

### Other Information
n/a


keon94:
UI-side work is here: #2630

klesh:
Hi, @keon94 , don't forget to prefix the issue number with "Closes" or "Fixes", so the issue could be closed automatically when the PR gets merged.

logger/logger.go
```diff
@@ -75,32 +95,84 @@ func (l *DefaultLogger) Error(format string, a ...interface{}) {
...
 
-// bind two writer to logger
-func (l *DefaultLogger) Nested(name string) core.Logger {
-	writerStd := os.Stdout
-	fileName := ""
-	loggerPrefixRegex := regexp.MustCompile(`(\[task #\d+]\s\[\w+])`)
-	groups := loggerPrefixRegex.FindStringSubmatch(fmt.Sprintf("%s [%s]", l.prefix, name))
-	if len(groups) > 1 {
-		fileName = groups[1]
+func (l *DefaultLogger) Nested(newPrefix string, config ...*core.LoggerConfig) core.Logger {
```
keon94, line 98: I am personally not a fan of this method signature, but wanted to do it without having to change any of the existing Nested(str) calls (for now anyway).
It makes more sense to just have Prefix also be a field inside LoggerConfig and make this function just take in a LoggerConfig instance. Maybe rename this from Nested to something else too

klesh: As I said, the `Nested` should be responsible for adding a prefix to the logging line only.
It should not be creating any file at all, to be accurate, the Logger itself should not be creating files at all.
Can we do it like the following:

```go

log := NewDefaultLogger(log *logrus.Logger)
taskLog := log.Nested(taskName)

// in Standalone mode
if LOGGING_DIR != "" {
  // an extra stream, the `log` would write it while writing to stdout/stderr
  log.SetStream(createPipelineLoggingStream(LOGGING_DIR, pipeline_id))
  // replace the stream
  taskLog.SetStream(createTaskLoggingStream(LOGGING_DIR, pipeline_id, task_id))
}

// in Temmporal mode in the FUTURE
log.SetStream(createRemotePipelineLoggingStream(LOGGING_DIR, pipeline_id))
taskLog.SetStream(createRemoteTaskLoggingStream(LOGGING_DIR, pipeline_id, task_id))


// do not do anything, let it write to the ParentLog
subtaskLog := log.Nested(subtaskName)
```

keon94: If you want Nested() to modify the existing logger rather than returning a new one, then we're irreversibly changing state on the logger. And as we pass this logger to other functions its state keeps getting changed and the behavior gets out of control. I think Nested() should return a new logger instance that inherits the properties of the original (like a fork()), but is detached from it. The change that I agree with you here is a SetStream API, I think that's a good idea.

keon94: If you want Nested() to modify the existing logger, then we need a copy constructor for the DefaultLogger to do the forking. Also, in that case, there's no point in Nested() returning Logger anymore.

klesh: No, I don't. Nested should return a new instance inherites the `Stream` from then original one.

keon94:
Both standalone and temporal workflows have been tested. Standalone behaves as expected and Temporal has the limitation that the log files from the workers are not downloaded, but the directory structure and the messages in them look correct. #2693 will tackle that limitation.

logger/logger.go
```diff
@@ -23,19 +23,39 @@ import (
...
+	filename  string
 }
 
-func NewDefaultLogger(log *logrus.Logger, prefix string, loggerPool map[string]*logrus.Logger) *DefaultLogger {
-	newDefaultLogger := &DefaultLogger{prefix: prefix, log: log}
-	newDefaultLogger.loggerPool = loggerPool
-	return newDefaultLogger
+func NewDefaultLogger(log *logrus.Logger, rootLogDir string) (core.Logger, error) {
+	if rootLogDir == "" {
+		rootLogDir = defaultBasePath
```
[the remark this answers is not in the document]
keon94: issue #2742 created

logger/init.go
```diff
@@ -48,10 +49,28 @@ func init() {
 		TimestampFormat: "2006-01-02 15:04:05",
 		FullTimestamp:   true,
 	})
+	basePath := cfg.GetString("LOGGING_DIR")
```
likyh, line 52: Maybe we should add some error log or panic here if `LOGGING_DIR` is empty?

models/logger.go
```diff
@@ -0,0 +1,56 @@
...
+package models
+
+import (
+	"fmt"
+	"github.com/apache/incubator-devlake/plugins/core"
+	"os"
+	"path/filepath"
+)
+
+func GetTaskLoggerPath(config *core.LoggerConfig, t *Task) string {
```
likyh, line 27: Sorry, why these logger files are put in `models/`?

logger/stream.go
```diff
@@ -0,0 +1,39 @@
...
+
+func GetFileStream(path string) (io.Writer, error) {
+	if path == "" {
+		return os.Stdout, nil
+	}
+	err := os.MkdirAll(filepath.Dir(path), os.ModePerm)
+	if err != nil {
+		return nil, err
+	}
+	file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0777)
```
likyh, line 34: It seems cannot send error message in HTTP request when file not exist.

likyh (changes requested):
Good job

likyh:
I find another bug here:
image
pipeline failed but Config-ui was not updated right.

I'm sure it's a backend bug and it does not exist in branch main.

logger/stream.go
keon94: This function call automatically creates the file if it doesn't exist. The os.O_CREATE flag is specifying that

likyh: if the path does not exist, it will be wrong.

models/logger.go
keon94: I moved this to logging/

keon94:
> I find another bug here: image pipeline failed but Config-ui was not updated right.
> 
> I'm sure it's a backend bug and it does not exist in branch main.

Did you try refreshing the UI page? I tried to recreating a similar problem and refreshing the UI fixed it

likyh:
> > I find another bug here: image pipeline failed but Config-ui was not updated right.
> > I'm sure it's a backend bug and it does not exist in branch main.
> 
> Did you try refreshing the UI page? I tried to recreating a similar problem and refreshing the UI fixed it

Yes, refreshed. I'm sure that task failed but pipelines are still underway.
image
image

It's right on main branch.

keon94:
@likyh yeah, it was a bug in my code - I was throwing the task error before updating the task status. I've fixed it now, please try again.

likyh (approved):
LGTM

Startrekzky (approved):
LGTM

PR #2662, opened 2022-08-02
fix the worklog collecting taking too long

### LINKED ISSUE #2653: [Bug][jira] The `worklog` Diff Sync didn't work as expected
### Search before asking

### What happened

According to @narrowizard reported:
He executed 2 consecutive pipelines in a row and the `worklog` collection took half an hour to complete

The logging of the First Pipeline:
image

The logging of the Second Pipeline:
image

By design, the Second collection should be in Diff Sync Mode, collects issues' worlogs only when its `updated` field had changed since the First collection.

Judged by the duration, it suggested that 5000+ issues had been updated in between 2 collections, which is highly unlikely.

### How to reproduce

Consult @narrowizard 

### Anything else

_No response_

### Version

main

### Code of Conduct

mindlesscloud:
# Summary
fix #2653 ([Bug][jira] The worklog Diff Sync didn't work as expected )

It is reasonable that collecting Jira worklogs takes a long time.
But from the second time, it should not take too long.
Before collecting, we first select issues which should be collecting worklogs.
It turns out, that issues without any worklogs were selected due to an unprecise SQL query.

close #2653 

### Screenshots

### Other Information


klesh (approved):
LGTM

PR #2770, opened 2022-08-18
fix: fix for ci

likyh:
# Summary

fix some connection ci and skip pipeline ci


config-ui/cypress/integration/0-api/api.spec.js
```diff
@@ -42,10 +42,8 @@ context('API Network Requests', () => {
         expect(response.body[0]).to.have.property('endpoint')
         expect(response.body[0]).to.have.property('username')
         expect(response.body[0]).to.have.property('password')
-        expect(response.body[0]).to.have.property('epicKeyField')
-        expect(response.body[0]).to.have.property('storyPointField')
-        expect(response.body[0]).to.have.property('remotelinkCommitShaPattern')
         expect(response.body[0]).to.have.property('proxy')
+        expect(response.body[0]).to.have.property('rateLimitPerHour')
```
e2corporation, line 46: The response key from the API is actually just `rateLimit`. All property assertions for rate limit should be updated to `rateLimit`.

e2corporation (approved):
@likyh Nice work, one small fix needed (`rateLimit` key) and rebase then LGTM.

config-ui/cypress/integration/0-api/api.spec.js
likyh: image
The response is rateLimitPerHour. Is it changed in any unmerged PR?

e2corporation: @likyh After rebuilding `main` it looks like it has been changed to `rateLimitPerHour` on response as well. No change necessary in that case.

Issue #2789, opened 2022-08-22
[Bug][config-ui] blueprint keep requesting jira connection when it was not selected


klesh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

I selected github connection only when creating blueprint
![image](https://user-images.githubusercontent.com/61080/185835220-0470d4b2-c86d-4ce1-be81-ac89d66aa13c.png)

However, it kept requesting jira connection even if none were selected.
![image](https://user-images.githubusercontent.com/61080/185835361-2fb5069d-2b10-418e-a386-9a6a7a73ebf3.png)

### What you expected to happen

Jira connection should not be requested when it was not selected.

### How to reproduce

1. create connection for jira and github
2. create blueprint, select github connection only
3. see log from `devlake`

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #2795, opened 2022-08-22
fix: fetch proxy only if jira is an enabled provider

e2corporation:
### 🔁  Config-UI / Blueprints / Create Blueprint

- [x] Fetch JIRA Proxy Resources only if JIRA is an Enabled (selected) Provider
- [x] Test Create Workflow

This PR adds a needed restriction to the **Create Blueprint Workflow** so that **JIRA Proxy** API resources are **only** fetched if JIRA is an _enabled_ Provider (ie. the user selects a JIRA connection for the current blueprint). In a scenario where JIRA is not detected, such as a blueprint with GitHub only -- API request calls for JIRA resources will not be made.

#2789


klesh (approved):
LGTM

PR #2783, opened 2022-08-19
feat: add connection rate limit setting

e2corporation:
### :package: Config-UI / Data Integrations / Connections - **Rate Limit** 

- [x] Add **Rate Limit** Connection Form Field
- [x] Update Connection Manager Hook (Define `rateLimit`)
- [x] Update Component Properties
- [x] Test Rate Limit Setting

### Description

This PR adds the **Rate Limit** (per hour) numeric setting for the **Connections** definition for All Data Providers (Plugins). The default value is `0`. NOTE: It's property is `NOT` included when _Testing_ a connection.

### Does this close any open issues?
#931

### Screenshots
Screen Shot 2022-08-19 at 7 19 03 PM

Screen Shot 2022-08-19 at 7 25 21 PM

PR #2796, opened 2022-08-22
refactor: active connections - use rateLimitPerHour property

e2corporation:
### Config-UI / Connection Manager / Rate Limit (#2783)

- [x] `Fix` use `rateLimitPerHour` property

This PR uses the updated API property key `rateLimitPerHour` instead of `rateLimit` that was previously used for the connection's response body.


klesh (approved):
LGTM

Issue #2882, opened 2022-08-30
[Feature][Config-UI] Create Plugin Registry


e2corporation:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Create a configurable Plugin Registry for Config-UI that will optimize the use and configuration of Data Providers.

### Use cases

As Developers build new Plugins for DevLake, we want to streamline the process of how those plugins are integrated with Configuration-UI.

- [x] Create a Defined Plugin Registration Schema (JSON Registration Files)
- [x] Create Local Plugin Registry Configuration
- [x] Minimize the amount of Configuration Scopes

### Advanced Use Cases
- [ ] Allow Plugins to Define Transformation Settings Fields
- [ ] Allow Dynamic "API" Plugins to be Loaded (Live Plugin Registry)

### Related issues

- #2847 
- #2826 

_No response_

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

likyh:
I don't think #2862 contradicts this issue. #2862 is the base of it.

e2corporation:
@likyh I had previously requested that you halt changes on refactoring key areas of the code when that ticket was opened, I don't agree with your approach on this, I would appreciate if you follow my lead instead of refactoring the whole ui codebase. Your changes are unnecessarily creating hours of code review and re-testing.

e2corporation:
Not so fast @github-actions

PR #2862, opened 2022-08-28
Refactor/refactor for dynamic connection

likyh:
### ⚠️ &nbsp;&nbsp;Pre Checklist

> Please complete _ALL_ items in this checklist, and remove before submitting

- [ ] I have read through the [Contributing](https://devlake.apache.org/community/) Documentation & [PR Template](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- [ ] This PR is using a `label` (bug, feature etc.)
- [ ] My code is has necessary documentation (if appropriate)
- [ ] I have added any relevant tests
- [ ] This section (**⚠️ &nbsp;&nbsp;Pre Checklist**) will be removed when submitting PR

# Summary

This PR just refactor some code and fix some bugs.
1. delete ProviderIcons and ProviderFormPlaceholders and replace them with one config `ProviderConfigMap`. (`integrationsData` will refactor in future)
    1. delete all usages of ProviderIcons and ProviderFormPlaceholders
    2. use ProviderConfigMap.columns to control UI in `config-ui/src/pages/configure/connections/ConnectionForm.jsx`
    3. replace some hard code array with generated array by ProviderConfigMap such as `config-ui/src/components/Sidebar/MenuConfiguration.jsx`
2. replace some `switch(provider.id)`/`[xxx,yyy].includes(provider.id)`/`[jira,github]` with more common code.
3. replace specific vars in connection with an object.
4. move some local vars to the local scope from the global manager. Such as `connectionsList` in `config-ui/src/pages/blueprints/create-blueprint.jsx`
5. fix some bugs when reading code:
    1. proxy cannot save in Jenkins.
    2. adding a connection shouldn't allow users to go back to the form page
    3. connection saving must return after 2 seconds even though the network is in localhost.
    4. get connectionList in wrong way when request all plugin's connections
6. refactor some unused code:
    1. delete empty `useEffect`
    2. replace `maxConnectionsExceeded` with existing `connectionLimitReached` in `config-ui/src/pages/configure/integration/manage.jsx`
    3. ......

### Does this close any open issues?
Please mention the issues here.

Issue #2846, opened 2022-08-26
[Feature][config-ui] generate plugin configuration page based on plugin metadata


tk103331:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Using the plugin mode is flexible. The same goes for the plugin's configuration parameters.
I found that the configuration page of the current plugin is defined by the front-end project. It cannot be automatically generated based on the plugins supported by the backend.

If I develop a new custom plugin myself, but the front-end does not have its configuration page. Front-end engineers must design new configuration pages.

I have an idea, I don't know if it will work or not.
**Customize the configuration page metadata in the plugin, and the front end generates page elements based on these metadata.**

The configuration parameters of the plugin need to be abstracted here. For example, parameters may have these properties: dataType, showType, validation, datasource, etc.
- **dataType**: Indicates the data type, such as number, string, array
- **showType**: Indicates the display type, such as input, select
- **validation**: used for form validation, such as required validation, regular validation, range validation
- **datasource**: the option data source for the 'select' component, which can be a simple dictionary, or the data returned by the interface provided by the plugin
- and other properties: **placeholder**, **helpText**, etc.

### Use case

generate plugin configuration page based on plugin metadata

![image](https://user-images.githubusercontent.com/4404609/186806234-bb250b52-0d6b-4efc-9db8-411cc3a1ea02.png)

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

tk103331:
Configuring metadata based on plugins can also solve this problem： https://github.com/apache/incubator-devlake/issues/2800。 

For example, add a 'security' attribute to the PAT parameter of the GitHub plugin.
The front-end configuration page generates a password input box.
The backend can store PAT parameters encrypted.

klesh:
Nice, this is exactly what we had in our mind for quite a while, we have been tweaking our system bit by bit toward this goal!
But, this is big, it may take a couple of iterations for things to work.
So, I created #2847 to track the overall progress regarding this matter.
I'm glad that we share the same vision, please keep a tab on #2847 if you are interested.

@likyh FYI

Startrekzky:
Bump

PR #2886, opened 2022-08-30
feat: setup configuration-ui plugin registry

e2corporation:
### 🆘 Config-UI / Plugin Registry (Phase I)

- [x] Create a Defined Plugin Registration Schema (JSON Registration Files)
- [x] Create **Local Plugin Registry** Configuration
   - [x] JIRA
   - [x] GitLab
   - [x] GitHub
   - [x] Jenkins
   - [x] TAPD
   - [x] AE
   - [x] DBT
   - [x] StarRocks
   - [x] RefDiff
   - [x] GitExtractor
   - [x] Feishu
   - [x] Azure 🆕 
   - [x] Bitbucket 🆕 
   - [x] Gitee 🆕 
- [x] Implement **Integrations Manager Hook** `@hooks/useIntegrations.jsx` 
   - [x] Create `Plugin` and other related **Data Models**
   - [x] Define Integrations Data Configuration Objects
   - [x] Validate JSON Plugin Configuration
- [x] Create **Integrations Context** & Provider (`@store/integrations-context.jsx`)
- [x] Create Basic Documentation for Plugin JSON Registration
- [x] `Refactor` Integrations Data Dependencies on `@data/Providers.js` & `@data/integrations.jsx`
- [x] `Refactor` **Transformations Manager** Hook (Setup _Default Transformations_ in Plugin Registry)
- [x] `Fix` Build Menu Configuration Dynamically
- [x] `Fix` Build All Connection Sources Dynamically
- [x] `Fix` Build Data Sources Connection List Dynamically
- [x] `Feat` Use Higher Order Component (HOC) Pattern for Transformation Settings Load
- [x] `Chore` Update Config-UI `README.md`
- [x] `Chore` Code Cleanups & Linting
- [x] `TEST` Data Connections Workflow
- [x] `TEST` Blueprints Workflow

### Description

This PR creates a **Plugin Registry** on **Configuration-UI** as outlined on #2882. The new **Integrations Manager** Hook now provides a **Dynamic API** Configuration for **Data Providers** that _replaces_ the static configuration formerly defined in the `@data/Provider.js` and `@data/integrations.jsx` global data constants. An **Integrations Context** has also been created to offer access to Dynamic Provider configuration to deeply nested child components and other Hook Services, when direct access to the IM Hook is unavailable or otherwise not desired.

As the Configuration API is preserved, Developers can continue to reference Provider Configuration objects in the same fashion as before without having to worry about maintaining multiple static configuration objects in multiple files.

### Developer Notes
For developers, access to the Dynamic **Plugin Registry** API Configuration can be done in the following ways, by using the Integrations Hook or the Integrations Context. With the Context, usage can be done in 2 ways via the `React.useContext` Hook or with using a Context Consumer & React Render Function.

**1. Integrations Hook** `@hooks/useIntegrations.jsx`
For Top-Level Page & Service Components, import and initialize the Integrations Manager Hook. Simply de-structure the specific configuration objects and properties needed.
```jsx
// Sample Hook Implementation (Abstract)
import React from 'react'
import useIntegrations from '@/hooks/useIntegrations'

class MyPageComponent {

 // Initialize Integrations Hook
 const {
    registry,
    plugins: Plugins,
    integrations: Integrations,
    activeProvider: IntegrationActiveProvider,
    Providers, 
    ProviderFormLabels,
    ProviderFormPlaceholders,
    ProviderConnectionLimits,
    setActiveProvider: setIntegrationActiveProvider
  } = useIntegrations()

 return (
   <>
       {Integrations.map((integration) => (<div>{integration.name}</div>) }
   </>
 )

}
```

**2. Integrations Context** + `React.useContext` Hook `@store/integrations-context.jsx`
Since Hooks can only be accessed by Top-Level Components, secondary _Child_ and deeper _nested_ components will need to use the Context for accessing provider configuration. Use object de-structure to reference the Configuration Objects needed.

```jsx
// Sample Context + useContext Implementation (Abstract)
import React, { useContext } from 'react'
import IntegrationsContext from '@/store/integrations-context'

const MyBlueprintComponent = (props) => {

 // Initialize Integrations Context
  const { 
    Integrations,
    Providers,
    ProviderIcons,  
    ProviderLabels 
  } = useContext(IntegrationsContext)

  return (
    <ul>
      <li>{ProviderLabels[Providers.JIRA]}</li>
      <li>{ProviderLabels[Providers.GITHUB]}</li>
      <li>{ProviderLabels[Providers.JENKINS]}</li>
    </ul>
  )
    
}
```

**3. Integrations Context + Consumer** Render Function `@store/integrations-context.jsx`
In cases where we don't need access to configuration at the Class or Function level, but only in specific areas of the Render, the Consumer Pattern with a React render function can also be used. Use object de-structure to reference the Configuration Objects needed.

```jsx
// Sample Context + Consumer Implementation (Abstract)
import React from 'react'
import IntegrationsContext from '@/store/integrations-context'

const MyNestedChildBlueprintComponent = (props) => {

  // Consumer + Render Function
  return (
        <IntegrationsContext.Consumer>
           {({ Providers, ProviderIcons }) => (
              <span style={{ marginLeft: '20px' }}>
                <span
                  className='iProvider-icon'
                >
                  {ProviderIcons[Providers.GITHUB] ? (
                    ProviderIcons[Providers.GITHUB](14, 14)
                  ) : (
                    <></>
                  )}
                </span>
              </span>
            )}
       </IntegrationsContext.Consumer> 
 )
}

```

**The Active Provider** (`activeProvider`)
The Active Data Provider is now managed by the Integrations Manager Hook (`@/hooks/useIntegrations.jsx`). After Plugin Registration, the `activeProvider` state variable will be an Instance of `Plugin` **Data Model** (`@models/Plugin.js`). The Provider Configuration is also accessible via the active provider when needed.

```jsx
// Using Active Provider & useIntegrations Hook (Abstract)
import React, { useEffect } from 'react'
import useIntegrations from '@/hooks/useIntegrations'

const MyDevLakeComponent = (props) => {

  // Initialize Integrations Hook
  const {
    activeProvider,
    setActiveProvider
  } = useIntegrations()

  useEffect(() => {
    // Active Provider (Plugin Instance)
    console.log('>>> MY ACTIVE DATA PROVIDER = ', activeProvider)
      
     // Data Entities?
    console.log('>>> Provider Data Entities = ', activeProvider?.entities)

    // Connection Form Labels?
    console.log('>>> Provider Connection Form Labels = ', activeProvider?.getConnectionFormLabels())
     
  }, [activeProvider])

} 
```

**Notice** ⚠️ 
In most cases old code has been commented out to lessen merge conflicts with other in-flight development tickets, more code cleanups will be performed in a future PR. Additionally, **legacy** component files **not** actively being used may still reference the old `@data/Providers.js` constant, hence they won't be updated.

### Milestone
`v0.15.0`

### Local Plugin Registry
The **local** registry defines plugins that are bundled with DevLake. Users creating "Private" local plugins may add and register plugins with the local registry.

### API Plugin Registry (**Future** Feature)
The **api** registry allows plugins to be defined by the backend and plugins can be fetched dynamically and merged with the local registry. `FEATURE` `TBD`
- [ ] Create **Dynamic / Live Plugin Registry** (Allow Plugins to load via Backend API)

### Plugin Registration Schema (JSON)

```
{
  "id": "gitlab",
  "type": "integration",
  "enabled": true,
  "multiConnection": true,
  "isBeta": false,
  "isProvider": true,
  "name": "GitLab",
  "icon": "src/images/integrations/gitlab.svg",
  "private": false,
  "connection": {
    "authentication": "token",
    "fields": {
      "name": { "enable": true, "required": true, "readonly": false },
      "endpoint": { },
      "proxy": { },
      "token": { },
      "rateLimitPerHour": { }
    },
    "labels": {
      "name": "Connection Name",
      "endpoint": "Endpoint URL",
      "proxy": "Proxy URL",
      "token": "Access Token",
      "rateLimitPerHour": "Rate Limit (per hour)"
    },
    "placeholders": {
      "name": "eg. GitLab",
      "endpoint": "eg. https://gitlab.com/api/v4/",
      "proxy": "eg. http://proxy.localhost:8080",
      "token": "eg. ff9d1ad0e5c04f1f98fa",
      "rateLimitPerHour": "1000"
    },
    "tooltips": {}
  },
  "entities": ["CODE", "TICKET", "CROSS"],
  "transformations": {
    "default": {
      "productionPattern": "",
      "deploymentPattern": ""
    }
  }
}
```

#2882

### Screenshots
`[<PENDING>]`
Screen Shot 2022-10-20 at 2 36 25 PM
Screen Shot 2022-10-20 at 2 36 34 PM

Screen Shot 2022-10-20 at 12 39 16 PM
Screen Shot 2022-10-20 at 12 37 24 PM


tk103331:
Is it better for the fields of the connection?

```
...
{
    "authentication": "token",
    "fields": [
        {
            "name": "name",
            "label": "Connection Name",
            "placeholder": "eg.GitLab",
            "enable": true,
            "required": true,
            "readonly": false
        },
        {
            "name": "endpoint",
            "label": "Endpoint URL",
            "placeholder": "eg. https://gitlab.com/api/v4/"
        },
        {
            "name": "token",
            "label": "Access Token",
            "placeholder": "eg. ff9d1ad0e5c04f1f98fa"
        }
    ]
}
...
```

This has at least the following benefits:

- Focus on all properties of a field when defining fields, rather than defining them in scattered places.
- The form of the fileds array can solve the sorting problem of the fields, and the way of using the JSON object is likely to cause the problem of the traversal order.

In addition, do you need to add some other attributes to the field, such as
- showType, which represents front-end display components, such as input, select, checkbox, etc.
- validation, which represents standard validation rules, including regular validation, length validation, etc.
- multiple, which means that it is a multiple format, such as the PAT of the GitHub plugin, you can fill in multiple.

There is some information in this issue https://github.com/apache/incubator-devlake/issues/2846

likyh:
> Wouldn't it be better to define the properties of the field like this:
> 
> ```
> ...
> {
>     "authentication": "token",
>     "fields": [
>         {
>             "name": "name",
>             "label": "Connection Name",
>             "placeholder": "eg.GitLab",
>             "enable": true,
>             "required": true,
>             "readonly": false
>         },
>         {
>             "name": "endpoint",
>             "label": "Endpoint URL",
>             "placeholder": "eg. https://gitlab.com/api/v4/"
>         },
>         {
>             "name": "token",
>             "label": "Access Token",
>             "placeholder": "eg. ff9d1ad0e5c04f1f98fa"
>         }
>     ]
> }
> ...
> ```
> 
> This has at least the following benefits:
> 
> * Focus on all properties of a field when defining fields, rather than defining them in scattered places.
> * The form of the fileds array can solve the sorting problem of the fields, and the way of using the JSON object is likely to cause the problem of the traversal order.
> 
> In addition, do you need to add some other attributes to the field, such as
> 
> * showType, which represents front-end display components, such as input, select, checkbox, etc.
> * validation, which represents standard validation rules, including regular validation, length validation, etc.
> * multiple, which means that it is a multiple format, such as the PAT of the GitHub plugin, you can fill in multiple.
> 
> There is some information in this issue #2846

I think so.

likyh:
No matter how to do it, it looks like we must use one config to control all the logic. I think #2862 "delete ProviderIcons and ProviderFormPlaceholders and replace them with one config ProviderConfigMap. (integrationsData will refactor in future)" and "replace some switch(provider.id)/[xxx,yyy].includes(provider.id)/[jira,github] with more common code." is necessary. Why you hate it so much?

e2corporation:
> No matter how to do it, it looks like we must use one config to control all the logic. I think #2862 「delete ProviderIcons and ProviderFormPlaceholders and replace them with one config ProviderConfigMap. (integrationsData will refactor in future)」 and 「replace some switch(provider.id)/[xxx,yyy].includes(provider.id)/[jira,github] with more common code.」 are necessary. Why you hate it so much?

You go through all the trouble of introducing another foreign `ProviderConfigMap` which is a hack workaround to stuff all the needed props into one place, add a note that you have to refactor it more in the future -- and you think that's a better solution that well defined JSON config files? Your solution was more out of desperation than a proper way to evolve the configuration system. I only have 1 commit on this PR -- there are more changes to be added. I already mentioned that the switch statements in connection manager will be consolidated and removed.

e2corporation:
A Plugin does not need to be responsible for determining it's own Connection Fields at this time with regards to them being dynamically created, it should be following a standard convention so Each Plugin has to satisfy the current Connection model defined by the backend. The plugin can have dynamic control (enabled state, required state etc) over the standard connection fields that exist currently, which is what this configuration is intended for. Based on a plugin's authentication type that also is a main factor as to what main fields are required for the connection.

In the future the plugin can perhaps have more control and custom rendering options for the field. This is something to be discussed in more detail.

likyh:
It's the first PR and ProviderConfigMap can be replaced with JSON or API requests at any time.

In addition, my two PRS may not be merged. But please don't create PR so big. It's an open-source project.

e2corporation:
> It's the first PR and ProviderConfigMap can be replaced with JSON or API requests at any time.
> 
> In addition, my two PRS may not be merged. But please don't create PR so big. It's an open-source project.

`ProviderConfigMap` is short term hack. Had you invested time in discussing with me what your intentions were you could have saved yourself a lot of time, Instead you choose to refactor on your own and make your own decisions, and create a PR with changes that go against my design strategy.

e2corporation:
@likyh @tk103331 I'll be making more commits as I continue ideating on the Plugin Registry concept. In the mean time, I'm fine with both of you contributing thoughts & ideas to this effort, that are hopefully inline with this strategy.

@tk103331 Feel free to comment on more suggestions/questions or issues for the Plugin Schema as it's being developed, some option configurations can be for a future need as well. The important thing is to get an essential schema for the short-term without overcomplicating it. You will also be able to add commits if you are wanting to contribute.

@likyh When I'm done with a few more commits, I'll tag you to adapt some of your valid refactor changes from your closed PR, such as the Dynamic Menu from new **$Integrations** var. There are also areas of the `useIntegrations` hook you can extend, such as the validation handler to verify a plugin's object properties meets Schema specifications, as well as working on live API registry.

tk103331:
Sorry, I didn't see what you edited in my reply two days ago.

> I do see your point, I wouldn't say it's scattered, sorting of the fields is not something needed or relevant, the order of the fields in the connection form follows a natural order starting with Name, ending with Rate Limit.

What I mean is to spread the multiple information (name, label, placeholder) of a field to be defined in multiple places instead of using only one object.
Regarding field sorting, in general, we will use the code writing order. However, if fileds is data requested from the backend. Possibly, the backend may use the map data structure (because the field is not fixed and cannot use struct), which will cause problems due to the uncertainty of the map traversal order. Of course, we can provide some instructions or helpers to avoid this problem, but it really depends on how the backend people write the code.

> I see the point, and we can certainly evolve the configuration to let validation rules be defined here but it's not needed right now. Instead of showType we would just use type (string|number|object) etc and add a component prop to specify a form control that's to be used for render.

Yes, the name `component` would be more appropriate, the name showType is a premature suggestion.

likyh:
Is this PR will continue now?

e2corporation:
> Is this PR will continue now?

I will be refreshing this soon, ideally other refactor PRs need to be completed or halted while I get this initial version of dynamic config ready.

e2corporation:
@klesh @likyh @mintsweet Phase 1 Of Dynamic **Plugin Registry** is ready for `main` branch **next** so more development & refactor can continue. Thanks.

config-ui/src/components/Sidebar.jsx
```diff
@@ -29,35 +35,46 @@ import { ReactComponent as Logo } from '@/images/devlake-logo.svg'
...
-    setMenu(MenuConfiguration(activeRoute))
-  }, [activeRoute])
+  // useEffect(() => {
+  //   setMenu(ActiveMenu)
+  // }, [ActiveMenu])
 
   useEffect(() => {
-    const fetchVersion = async () => {
-      try {
-        const versionUrl = `${DEVLAKE_ENDPOINT}/version`
```
likyh, line 48: this API has been deleted? @klesh

config-ui/src/registry/plugins/azure.json
```diff
@@ -0,0 +1,53 @@
...
+    "authentication": "plain",
+    "fields": {
+      "name": { "enable": true, "required": true, "readonly": false },
+      "endpoint": { },
+      "proxy": { },
+      "username": { },
+      "password": { },
+      "rateLimitPerHour": { }
+    },
+    "labels": {
```
likyh, line 22: why do we add `labels`, `tooltips` and `placeholders` into `fields`? It looks all these are dicribe `fields`.

config-ui/src/components/Sidebar.jsx
e2corporation: @likyh We no longer display the version tag on the sidebar, fetch is not needed at the moment.

config-ui/src/registry/plugins/azure.json
e2corporation: @likyh The fields group is for other meta configuration options for the connection field properties themselves, it is not fully implemented as yet and will be in another PR. These `labels` , `tooltips` and `placeholder` props maybe reorganized in the future according to suggestions by @tk103331 that would consolidate them.

config-ui/src/components/Sidebar.jsx
```diff
@@ -29,35 +35,46 @@ import { ReactComponent as Logo } from '@/images/devlake-logo.svg'
...
   const activeRoute = useRouteMatch()
   const uiContext = useContext(UIContext)
 
-  const [menu, setMenu] = useState(MenuConfiguration(activeRoute))
+  const getMenu = useCallback(
+    () => MenuConfiguration(activeRoute, integrations),
+    [activeRoute, integrations]
+  )
+
+  const ActiveMenu = useMemo(() => getMenu(), [getMenu])
```
mintsweet, line 50: `useMemo(() =>  MenuConfiguration(activeRoute, integrations), [...])`

Wouldn't this save the **useCallback** above?

config-ui/src/components/Sidebar.jsx
```diff
@@ -29,35 +35,46 @@ import { ReactComponent as Logo } from '@/images/devlake-logo.svg'
...
+    //     })
+    //     setVersionTag(res?.data ? res.data?.version : '')
+    //   } catch (e) {
+    //     setVersionTag('')
+    //   }
+    // }
+    // fetchVersion()
   }, [])
 
+  // useEffect(() => {}, [integrations])
```
mintsweet, line 76: I think since it is a refactoring, these commented-out codes can be deleted.

config-ui/src/hooks/data-scope/useTransformationsManager.jsx
```diff
@@ -15,103 +15,109 @@
...
+      let key = `not-distinguished`
+      switch (connectionProvider) {
+        case Providers.GITHUB:
+        case Providers.GITLAB:
+        case Providers.JENKINS:
+          key = projectNameOrBoard?.id
+          break
+        case Providers.JIRA:
+          key = projectNameOrBoard?.id
+          break
```
mintsweet, line 115: I found this place before, and I didn't see the difference between the two conditions.

config-ui/src/hooks/useIntegrations.jsx
```diff
@@ -15,17 +15,308 @@
...
+
+const ProviderTypes = {
+  PLUGIN: 'plugin',
+  INTEGRATION: 'integration',
+  PIPELINE: 'pipeline',
+  WEBHOOK: 'webhook'
+}
+
+function useIntegrations(
+  pluginRegistry = [
```
likyh, line 46: It looks like a const rather than a param.

config-ui/src/hooks/useDataScopesManager.jsx
```diff
@@ -363,27 +397,31 @@ function useDataScopesManager({
...
+        case Providers.JIRA:
+          entities = DEFAULT_DATA_ENTITIES.filter(
+            (d) => d.name === 'issue-tracking' || d.name === 'cross-domain'
+          )
+          break
+        case Providers.JENKINS:
+          entities = DEFAULT_DATA_ENTITIES.filter((d) => d.name === 'ci-cd')
+          break
+        case Providers.TAPD:
+          entities = DEFAULT_DATA_ENTITIES.filter((d) => d.name === 'ci-cd')
```
mintsweet, line 418: **entities** can also be defined in the plugin so that the conditional judgment here can be omitted.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
```diff
@@ -24,10 +24,12 @@ import {
   Radio,
   Tag
 } from '@blueprintjs/core'
-import { Providers, ProviderLabels } from '@/data/Providers'
+// import { Providers, ProviderLabels } from '@/data/Providers'
 
 const Deployment = (props) => {
   const {
+    Providers = {},
+    ProviderLabels = {},
```
likyh, line 32: What difference between props and useContext?

config-ui/src/hooks/usePipelineManager.jsx
```diff
@@ -47,18 +49,9 @@ function usePipelineManager(
...
-    Providers.JENKINS,
-    Providers.GITHUB,
-    Providers.REFDIFF,
-    Providers.GITEXTRACTOR,
-    Providers.FEISHU,
-    Providers.AE,
-    Providers.DBT,
-    Providers.TAPD
-  ])
+  const [allowedProviders, setAllowedProviders] = useState(
```
mintsweet, line 52: **Config UI** doesn't seem to support all plugins yet, is this judgment unnecessary?

config-ui/src/components/Sidebar.jsx
e2corporation: Commented code will be removed in a future cleanup.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
e2corporation: Context can be used by Child-components directly.

config-ui/src/hooks/useIntegrations.jsx
e2corporation: param with a default value

config-ui/src/hooks/data-scope/useTransformationsManager.jsx
e2corporation: Projects and Boards will be merged in a separate PR by @likyh, this switch block will be removed after.

config-ui/src/hooks/usePipelineManager.jsx
e2corporation: It's not being fully enforced in Advanced Mode, so this restriction won't apply.

config-ui/src/components/Sidebar.jsx
e2corporation: It's needed to prevent excessive re-render of the MenuConfiguration

config-ui/src/hooks/useDataScopesManager.jsx
e2corporation: Good point, I'll update this to get entities dynamically.

config-ui/src/hooks/data-scope/useTransformationsManager.jsx
e2corporation: Then again it would be good to make the update to the generateKey function on this PR.

e2corporation: Fixed.

config-ui/src/hooks/useDataScopesManager.jsx
e2corporation: Fixed.

config-ui/src/registry/plugins/azure.json
likyh: OK, fix in another PR

config-ui/src/hooks/useIntegrations.jsx
likyh: It looks like never used as a param. Just define it as a const.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
likyh: So here we can use `useContext` too?

config-ui/src/pages/configure/settings/tapd.jsx
```diff
@@ -24,6 +24,8 @@ import '@/styles/connections.scss'
 
 export default function TapdSettings(props) {
   const {
+    Providers,
```
likyh, line 27: Let's don't add unused parameters.

config-ui/src/pages/configure/integration/manage.jsx
```diff
@@ -461,7 +480,7 @@ export default function ManageIntegration() {
                         ))}
                       </tbody>
                     </table>
-                    {connectionLimitReached && (
+                    {connectionLimitReached ? (
```
likyh, line 483: They have the same effect.

config-ui/src/pages/configure/integration/manage.jsx
```diff
@@ -490,7 +509,7 @@ export default function ManageIntegration() {
                     }}
                   >
                     Fetched <strong>{connections.length}</strong> connection(s)
-                    from Lake API for <strong>{activeProvider.name}</strong>
+                    from Lake API for <strong>{activeProvider?.name}</strong>
```
likyh, line 512: If we know `activeProvider` won't be null, Let don't use `?.`. Sometimes it will uncover some bugs because it doesn't raise an error and just displays empty.

config-ui/src/pages/configure/settings/tapd.jsx
e2corporation: TAPD settings needs to be implemented, having unused imports here is not a problem for now.

config-ui/src/pages/configure/integration/manage.jsx
e2corporation: No, the first will cause an unnecessary render of a value leading to a stray "0" character in the table row.

config-ui/src/pages/configure/integration/manage.jsx
e2corporation: This is safer to use optional chaining, it may be removed in the future. If `activeProvider` was not available, other errors would be raised.

config-ui/src/hooks/useJIRA.jsx
```diff
@@ -15,17 +15,19 @@
...
+import IntegrationsContext from '@/store/integrations-context'
+// import { Providers } from '@/data/Providers'
 import DataScopeConnection from '@/models/DataScopeConnection'
 
 const useJIRA = (
   { apiProxyPath, issuesEndpoint, fieldsEndpoint, boardsEndpoint },
   activeConnection = null,
   setConnections = () => []
 ) => {
+  const { Providers } = useContext(IntegrationsContext)
```
likyh, line 30: `Providers` should not use in `useJIRA` hook. This hook is a util hook, and it should keep pure and basic of other UI components. 

Also it's never used.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
e2corporation: I will remove these added props in favor of the Integrations Context in another commit to this PR. These properties were defined before the context was added, this is why new props are being used in the first place.

config-ui/src/hooks/useJIRA.jsx
e2corporation: Correct, it was never used before, but it was defined, so it had to be updated as a result of these new changes. The Integrations Hook is mainly Configuration and can be used anywhere, including in the `useJIRA` Hook where it can be used if needed. I'll remove the reference for now.

e2corporation: Fixed.

config-ui/src/components/blueprints/ProviderTransformationSettings.jsx
```diff
@@ -50,102 +64,32 @@ const ProviderTransformationSettings = (props) => {
...
-          onSettingsChange={onSettingsChange}
-          entities={entities[connection?.id]}
-          isSaving={isSaving}
-          isSavingConnection={isSavingConnection}
-        />
-      )}
+  // Provider Transformation Components (LOCAL)
+  const TransformationComponents = useMemo(
+    () => ({
+      [Providers.GITHUB]: GithubSettings,
```
likyh, line 70: use ReactNode as the value of `useMemo`? It looks like not a common way, and I don't know if it can work. We can write like this:

```js
const TransformationComponents = {
  [Providers.GITHUB]: GithubSettings,
  [Providers.GITLAB]: GitlabSettings,
  [Providers.JIRA]: JiraSettings,
  [Providers.JENKINS]: JenkinsSettings,
  [Providers.TAPD]: TapdSettings,
  [Providers.AZURE]: AzureSettings,
  [Providers.BITBUCKET]: BitbucketSettings,
  [Providers.GITEE]: GiteeSettings
}
let TransformationWithProviderSettings = TransformationComponents[provider?.id]
return (
  <div className='transformation-settings' data-provider={provider?.id}>
    <TransformationWithProviderSettings
          provider={provider}
          connection={connection}
          transformation={transformation}
          onSettingsChange={onSettingsChange}
          entities={entities[connection?.id]}
          isSaving={isSaving}
          isSavingConnection={isSavingConnection}
        />
  </div>
)
```

https://reactjs.org/docs/jsx-in-depth.html#choosing-the-type-at-runtime

config-ui/src/data/Providers.js
```diff
@@ -381,13 +391,15 @@ const ProviderIcons = {
...
 
+// @note: migrated to @data/ConnectionStatus
 const ConnectionStatus = {
   OFFLINE: 0,
   ONLINE: 1,
   DISCONNECTED: 2,
   TESTING: 3
 }
 
+// @note: migrated to @data/ConnectionStatus
```
likyh, line 402: When will these move and delete?

config-ui/src/hooks/useConnectionManager.jsx
```diff
@@ -38,6 +37,18 @@ function useConnectionManager(
...
   const { handleOfflineMode } = useNetworkOfflineMode()
 
+  const {
+    registry,
+    plugins: Plugins,
+    integrations: Integrations,
+    activeProvider: IntegrationActiveProvider,
+    Providers,
+    ProviderFormLabels,
+    ProviderFormPlaceholders,
```
likyh, line 47: Just import the state used.

config-ui/src/hooks/useDataScopesManager.jsx
```diff
@@ -35,6 +36,19 @@ function useDataScopesManager({
...
+  const {
+    registry,
+    plugins: Plugins,
+    integrations: Integrations,
+    activeProvider: IntegrationActiveProvider,
+    Providers,
+    ProviderLabels,
+    ProviderIcons,
+    ProviderFormLabels,
+    ProviderFormPlaceholders,
```
likyh, line 48: Just import the necessaries

config-ui/src/hooks/useDataScopesManager.jsx
```diff
@@ -108,68 +122,88 @@ function useDataScopesManager({
...
-          newScope = {
-            ...newScope
-            // options: {
-            // },
-            // transformation: {},
-          }
-          break
-      }
+      // Generate scopes Dynamically for all Project/Board/Job/... Entities
+      // @todo: refactor again when boards & projects get merged...
```
likyh, line 126: ok

config-ui/src/hooks/useDataScopesManager.jsx
```diff
@@ -108,68 +122,88 @@ function useDataScopesManager({
...
+          ? projects[connection.id]
+          : [])
+      ].map((e) => ({
+        ...newScope,
+        options: {
+          ...getTransformationScopeOptions(connection?.providerId, e)
+        },
+        transformation: {
+          ...getTransformation(connection?.providerId, connection?.id, e)
+        }
```
likyh, line 139:
```js
        options: getTransformationScopeOptions(connection?.providerId, e),
        transformation: getTransformation(connection?.providerId, connection?.id, e),
```
Or they are return the same object every time?

config-ui/src/hooks/data-scope/useTransformationsManager.jsx
```diff
@@ -178,7 +188,27 @@ const useTransformationsManager = () => {
...
+      const key = generateKey(connectionProvider, projectNameOrBoard)
+      const plugin = Integrations.find((p) => p.id === connectionProvider)
+      console.debug(
+        '>> useTransformationsManager.getScopeOptions...',
+
+        connectionProvider,
+        projectNameOrBoard
+      )
+      return plugin &&
+        typeof plugin?.getDefaultTransformationScopeOptions === 'function'
```
likyh, line 207: Is this typo? Actual value is `getTransformationScopeOptions` defined in models?

config-ui/src/components/blueprints/ProviderTransformationSettings.jsx
e2corporation: This is well supported and a proper use case for applying a Higher-Order-Component (HOC), documented in the "JSX In Depth" section. Components can be created dynamically from a variable reference when needed, especially in our scenario where we need transformation settings to load dynamically without the need for having a switch statement.

https://reactjs.org/docs/jsx-in-depth.html

Screen Shot 2022-10-28 at 9 37 43 AM

config-ui/src/data/Providers.js
e2corporation: It's already done, this is just a comment to be removed when Providers.js is cleaned up and/or deprecated.

config-ui/src/components/blueprints/ProviderTransformationSettings.jsx
e2corporation: Additionally, Entire Components can be Memoized.

config-ui/src/hooks/useDataScopesManager.jsx
e2corporation: No not the same return value, `getTransformation` returns _transformation_ values. `getTransformationScopeOptions` is for the other _data scope options_  for the Entity that were previously declared when provider scopes were manually built with the switch statement. Configuration-wise in the registry json I've grouped it with transformations as it's related to transformations.

config-ui/src/hooks/data-scope/useTransformationsManager.jsx
e2corporation: No not a Typo, this is from the **Plugin** Model where it's defined as `getDefaultTransformationScopeOptions`, inside that function `getTransformationScopeOptions` is used on the Model correctly.

config-ui/src/pages/blueprints/blueprint-detail.jsx
```diff
@@ -59,13 +59,26 @@ import StageLane from '@/components/pipelines/StageLane'
...
   // eslint-disable-next-line no-unused-vars
   const history = useHistory()
+
+  const {
+    registry,
+    plugins: Plugins,
+    integrations: Integrations,
+    Providers,
+    ProviderIcons,
+    ProviderLabels,
```
likyh, line 77: just import states used.

config-ui/src/pages/blueprints/blueprint-settings.jsx
```diff
@@ -89,7 +89,18 @@ const BlueprintSettings = (props) => {
...
   const { bId } = useParams()
 
-  const [activeProvider, setActiveProvider] = useState(integrationsData[0])
+  const {
+    registry,
+    plugins: Plugins,
+    integrations: Integrations,
+    Providers,
+    ProviderLabels,
+    ProviderIcons,
```
likyh, line 98: just import states used.

config-ui/src/pages/blueprints/create-blueprint.jsx
```diff
@@ -91,6 +92,22 @@ const CreateBlueprint = (props) => {
...
+    registry,
+    plugins: Plugins,
+    integrations: Integrations,
+    activeProvider,
+    DataSources: DataSourcesList,
+    Providers,
+    ProviderLabels,
+    ProviderIcons,
+    ProviderFormLabels,
+    ProviderFormPlaceholders,
```
likyh, line 105: just import states used.

config-ui/src/pages/configure/connections/AddConnection.jsx
```diff
@@ -42,9 +43,23 @@ export default function AddConnection() {
...
-    integrationsData.find((p) => p.id === providerId)
-  )
+  const {
+    registry,
+    plugins: Plugins,
+    integrations: Integrations,
+    activeProvider,
+    Providers,
+    ProviderFormLabels,
+    ProviderFormPlaceholders,
```
likyh, line 53: just import states used. And some others like this.

config-ui/src/pages/configure/integration/manage.jsx
likyh: !!connectionLimitReached &&

config-ui/src/pages/configure/integration/manage.jsx
likyh: "other errors would be raised", then we can notice it. Displaying empty sometimes uncover some bugs because .

likyh: I think "Resolve Conversation" need to be clicked by the reviewer..

config-ui/src/pages/configure/connections/AddConnection.jsx
e2corporation: This is can be done in a future cleanup PR.

config-ui/src/pages/blueprints/create-blueprint.jsx
e2corporation: This is can be done in a future cleanup PR.

config-ui/src/pages/blueprints/blueprint-settings.jsx
e2corporation: This is can be done in a future cleanup PR.

config-ui/src/pages/blueprints/blueprint-detail.jsx
e2corporation: This is can be done in a future cleanup PR.

config-ui/src/hooks/useConnectionManager.jsx
e2corporation: Fixed.

config-ui/src/data/Providers.js
likyh: ok

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
e2corporation: Fixed.

config-ui/src/components/blueprints/ProviderTransformationSettings.jsx
e2corporation: The main reason for the `Memo` as well is because `Providers` is no longer imported from a Global Data Constant, Originally I had `TransformationComponents` Map defined _outside_ the class, but with the new design Providers is not accessible in a global state outside the class, therefore it has to be moved inside the the Component and wrapped with a Memo.

config-ui/src/pages/configure/integration/manage.jsx
e2corporation: This is a non-issue, optional chaining was not designed to "cover" bugs but to cleanly add non-exist state checking. I can resolve a conversation If I feel the question has been answered and there is no problem identified.

config-ui/src/components/Sidebar.jsx
e2corporation: @likyh Version Tag is not a key focus of this PR, anything related to version tag can be tracked under #3607

config-ui/src/pages/configure/integration/manage.jsx
e2corporation: Suggestion applied.

e2corporation:
@likyh @mintsweet All notable code review feedback has been resolved, any additional cleanups will be deferred to a subsequent PR.

likyh (approved):

PR #2890, opened 2022-08-31
Issues/2888-2 move local var connectionsList to the local scope

### LINKED ISSUE #2888: [Refactor][config] get `connectionsList` in the wrong way when requesting all plugin connections.
## What and why to refactor
get `connectionsList` in the wrong way when requesting all plugin connections.

https://github.com/apache/incubator-devlake/blob/main/config-ui/src/hooks/useConnectionManager.jsx#L402-L422

c is not set when requesting allSources.

move it to the right statement.

likyh:
# Summary

base on #2889 

move local var connectionsList to the local scope config-ui/src/pages/blueprints/create-blueprint.jsx from the global manager.

We should not use `setXXX` to change the response result in the manager. It breaks the cohesion rule of useManager.

close #2888

### DIFF
--- a/config-ui/src/hooks/useConnectionManager.jsx
+++ b/config-ui/src/hooks/useConnectionManager.jsx
@@ -78,8 +78,6 @@ function useConnectionManager (
   [/*provider?.id, */sourceLimits, connectionCount],
   )
 
-  const [connectionsList, setConnectionsList] = useState([])
-
   const [saveComplete, setSaveComplete] = useState(false)
   const [deleteComplete, setDeleteComplete] = useState(false)
   const connectionTestPayload = useMemo(() => ({ endpoint: endpointUrl, username, password, token, proxy }), [endpointUrl, password, proxy, token, username])
@@ -636,28 +634,6 @@ function useConnectionManager (
     setProvider(activeProvider)
   }, [activeProvider])
 
-  useEffect(() => {
-    console.log('>>> ALL DATA PROVIDER CONNECTIONS...', allProviderConnections)
-    setConnectionsList(
-      allProviderConnections?.map((c, cIdx) => ({
-        ...c,
-        id: cIdx,
-        key: cIdx,
-        connectionId: c.id,
-        name: c.name,
-        title: c.name,
-        value: c.id,
-        status:
-          ConnectionStatusLabels[c.status] ||
-          ConnectionStatusLabels[ConnectionStatus.OFFLINE],
-        statusResponse: null,
-        provider: c.provider,
-        providerId: c.provider,
-        plugin: c.provider,
-      }))
-    )
-  }, [allProviderConnections])
-
   return {
     activeConnection,
     fetchConnection,
@@ -702,11 +678,9 @@ function useConnectionManager (
     setTestResponse,
     setAllTestResponses,
     setConnectionLimits,
-    setConnectionsList,
     setSaveComplete,
     allConnections,
     allProviderConnections,
-    connectionsList,
     domainRepositories,
     testedConnections,
     sourceLimits,
--- a/config-ui/src/pages/blueprints/blueprint-settings.jsx
+++ b/config-ui/src/pages/blueprints/blueprint-settings.jsx
@@ -43,7 +43,7 @@ import {
 import { integrationsData } from '@/data/integrations'
 import { NullBlueprint, BlueprintMode } from '@/data/NullBlueprint'
 import { NullPipelineRun } from '@/data/NullPipelineRun'
-import { Providers, ProviderLabels, ProviderIcons } from '@/data/Providers'
+import { Providers, ProviderLabels, ProviderIcons, ConnectionStatusLabels, ConnectionStatus } from '@/data/Providers'
 import {
   TaskStatus,
 } from '@/data/Task'
@@ -186,10 +186,7 @@ const BlueprintSettings = (props) => {
   })
 
   const {
-    fetchConnection,
     allProviderConnections,
-    connectionsList,
-    isFetching: isFetchingConnection,
     fetchAllConnections,
   } = useConnectionManager(
     {
@@ -198,6 +195,26 @@ const BlueprintSettings = (props) => {
     },
     configuredConnection && configuredConnection?.id !== null
   )
+
+  const connectionsList = useMemo(() => {
+    console.log('>>> ALL DATA PROVIDER CONNECTIONS...', allProviderConnections)
+    return allProviderConnections?.map((c, cIdx) => ({
+      ...c,
+      id: cIdx,
+      key: cIdx,
+      connectionId: c.id,
+      name: c.name,
+      title: c.name,
+      value: c.id,
+      status:
+        ConnectionStatusLabels[c.status] ||
+        ConnectionStatusLabels[ConnectionStatus.OFFLINE],
+      statusResponse: null,
+      provider: c.provider,
+      providerId: c.provider,
+      plugin: c.provider,
+    }))
+  }, [allProviderConnections])
 
   const {
     // eslint-disable-next-line no-unused-vars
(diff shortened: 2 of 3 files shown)


config-ui/src/hooks/useConnectionManager.jsx
```diff
@@ -393,38 +393,32 @@ function useConnectionManager (
           )
           console.log('>> ALL SOURCE CONNECTIONS: ', aC)
         } else {
-          c = await request.get(
+          const c = await request.get(
```
e2corporation, line 396: A variable declaration should not be done inside an IF block, I think this will get hoisted anyway. The definition I had on `Ln 362` is correct.

e2corporation (changes requested):
Resolve comment regarding `Ln 362` on how `c` is defined, then LGTM.

config-ui/src/hooks/useConnectionManager.jsx
likyh: No, it won't be hoisted.
image
```
function f() {
    if (true) {
        const b = 1
        console.log(b)
    }
    console.log(b)
}
```

e2corporation: I still say var declarations inside if blocks aren't ideal, keep it outside as it was before. I want to rework this logic in the future.

e2corporation: Only reason it isn't hoisted is because of ES6 block scoping and using `const` or `let` won't get hoisted.

likyh: Yes, var will but const/let won't. `c` is just used in the else block, why do we need to move it out of `if`? forget `var` and es5.

likyh: ... ok, although I don't agree with it but this time according to your preference.

e2corporation: variable delcarations inside an `If` statement is not ideal practice in either case. Only in workaround situations do vars get declared inside conditionals.

config-ui/src/pages/blueprints/blueprint-settings.jsx
```diff
@@ -199,6 +196,26 @@ const BlueprintSettings = (props) => {
     configuredConnection && configuredConnection?.id !== null
   )
 
+  const connectionsList = useMemo(() => {
```
e2corporation, line 199: Why would we want to move this to local scope when it should be, and already was managed by the Connection Manager to begin with? Now the same const `connectionsList` has duplicated definition in 2 places. `connectionsList` belongs in the Connection Manager Hook like previously defined.

likyh: We should not use setXXX to change the response result in the manager. It breaks the cohesion rule of useManager.
Or we can define a new memo on the page component if we must change it.

likyh: I think useXxxxManager should be a black box. We just know what its input and output are, and then we can use it.

https://github.com/apache/incubator-devlake/blob/main/config-ui/src/pages/blueprints/create-blueprint.jsx#L940-L947
But here, it changes output out of this black box. We should not change the states which are the output vars or control flags out of useXxxxManager, such as isFetching, testResults or others. 
Here is the same. If we defined `connectionsList` in the manager, that means we regard it as the output of the response. So we should change it out of the manager.

Or we can define a new memo like `withStatusConnectionsList` on the page component if we must change `connectionsList` which is defined by the manager. I think connectionsList is just a tmp var and it will be refactor soon, so just move it out of useXxxxManager to avoid confusion when debuging.

e2corporation: That's not how this hook, or any of the other hooks should function -- Too restricive. `isFetching` like many of the other processing flags I added are to be controlled by whichever action is doing a processing operation. Sometimes this maybe internal to the hook, in which case it's state is toggled within the hook. For outside operations the caller implementing the hook will also alter state. It's not a temp var, it's a Map list built specifically for BlueprintJS selector components. The whole point of a hook is to produce shareable reusable state logic and associated callbacks, not a blackbox.

Please define the `connectionsList` in connection manager as originally configured, test that there were no breakages after having converting to a memo like you've done, and this one can be closed.

likyh: > isFetching like many of the other processing flags I added are to be controlled by whichever action is doing a processing operation.
That is one of the most confusing if we use output/flag out of manager, I think. Too many useEffect between different levels of components and it confuse me once and once again. Maybe we can launch a conversation with other frontend developers to decide if it's a good practice.

e2corporation: Use of Effects will be subjective, I'm not breaking any React Rules on how effects are to be used. Effect Chains are important to creating a controlled data flow that allows for data to be processed/manipulated and then kicked off to the next effect in the chain to continue the process until no more effects re triggered. I try to organize and group all effects in the same block, in order to track the behavior you look at the dependency arrays and follow what's changing. Let's not make this more complex than it has to be, the amount of effects constructed depends on the features being provided.

e2corporation: Effects can also return a Callback/closure to "cleanup" state which is sometimes needed and recommended if you are worried about what happens after the effect runs, for example if some interval polling is turned on or state needs to be reset during unmount, the cleanup function returned by the effect will clear the interval. In many cases however a cleanup function is not needed.

e2corporation: I think if you start building features and you see the correlation on how effects can make the UI interactions more meaningful, you will start adopting the use of more effects. In most cases Effects are only problem if it suffers from a memory leak.

likyh: Let's listen to other frontend developers about the code existing next week. Another frontend developer will join this project next week.

e2corporation: I don't think that new developers who have not invested any time in this stack, are not familar with it's design and development, should be responsible for determining how effects are used in this context. You have moved centralized logic from one consolidated place, and duplicated that logic in 2 component files and are now trying to justify the change. I can't see anyone approving a change like this in any language...

Please revert the change to the Effect like it was originally, or if replaced with useMemo it should be defined in the Connection Manager Hook.

likyh: It's a question if new developers have the same confusion as me. 

This pr closed.

e2corporation: The only way I see having a Memo for `connectionsList` duplicated at the component layer is to migrate the map closure into the Connection Manager Hook, that way this logic isn't duplicated at the component level.

```
const createConnectionList = (c, cIdx) => ({
       ...c,
       id: cIdx,
       key: cIdx,
       connectionId: c.id,
       name: c.name,
       title: c.name,
       value: c.id,
       status:
         ConnectionStatusLabels[c.status] ||
         ConnectionStatusLabels[ConnectionStatus.OFFLINE],
       statusResponse: null,
       provider: c.provider,
       providerId: c.provider,
       plugin: c.provider,
     }))
   }
```

Then at least the useMemo() calls in 2 pages would be using a centralized callback, and the object creation logic won't be duplicated.

e2corporation: > It's a question if new developers have the same confusion as me.
> 
> This pr closed.

I understand, I think a certain degree of confusion for any developer when entering a new stack is expected and normal, regardless of experience there is still a slight learning curve when adapting to a codebase.

PR #2895, opened 2022-08-31
refactor: replace connectionLimit with exist state

### LINKED ISSUE #2894: [Refactor][config-ui] replace maxConnectionsExceeded with existing connectionLimitReached
## What and why to refactor
An `maxConnectionsExceeded()` defined in page component:
https://github.com/apache/incubator-devlake/blob/main/config-ui/src/pages/configure/integration/manage.jsx#L114
But actually, it also defined in manager:
https://github.com/apache/incubator-devlake/blob/main/config-ui/src/hooks/useConnectionManager.jsx#L424-L427

replace maxConnectionsExceeded with existing connectionLimitReached in config-ui/src/pages/configure/integration/manage.jsx

likyh:
# Summary

replace `connectionLimit` with exist state.

After the fix, it works right (if we would limit these plugins).
image
image

close #2894

### DIFF
--- a/config-ui/src/pages/configure/integration/manage.jsx
+++ b/config-ui/src/pages/configure/integration/manage.jsx
@@ -55,7 +55,7 @@ export default function ManageIntegration () {
   const [deleteId, setDeleteId] = useState()
 
   const {
-    sourceLimits,
+    connectionLimitReached,
     allConnections: connections,
     testedConnections,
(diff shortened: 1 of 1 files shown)


e2corporation (approved):
Even though multi-connection is enabled for all providers and no source limits are required at this time, this is a useful fix. LGTM.

PR #2898, opened 2022-08-31
feat: replace specific vars in connection with an object.

### LINKED ISSUE #2897: [Bug][config-ui] proxy in jenkins cannot be saved
### Search before asking

### What happened

image
image

save proxy correctly.

### How to reproduce

Modify a Jenkins connection and open it again, and you'll see the proxy is empty and not saved.

### Anything else

_No response_

### Version

main

### Code of Conduct

likyh:
# Summary

replace specific vars in connection with an object.

If we would add a new input on connection pages, we must pass it on several levels of components. That's not a good way to do it. So I use an object to replace it.

And I also fix #2897 in this PR.

close #2897

### DIFF
--- a/config-ui/src/components/blueprints/ConnectionDialog.jsx
+++ b/config-ui/src/components/blueprints/ConnectionDialog.jsx
@@ -92,14 +92,9 @@ const ConnectionDialog = (props) => {
     setTestStatus = () => {},
     setTestResponse = () => {},
     connection = NullBlueprintConnection,
-    name,
-    endpointUrl,
-    proxy,
-    rateLimit = 0,
-    token,
+    editingConnection,
+    setConnectionColumn,
     initialTokenStore = {},
-    username,
-    password,
     isLocked = false,
     isLoading = false,
     isTesting = false,
@@ -114,13 +109,6 @@ const ConnectionDialog = (props) => {
     onClose = () => {},
     onCancel = () => {},
     onValidate = () => {},
-    onNameChange = () => {},
-    onEndpointChange = () => {},
-    onProxyChange = () => {},
-    onRateLimitChange = () => {},
-    onTokenChange = () => {},
-    onUsernameChange = () => {},
-    onPasswordChange = () => {},
     showConnectionError = false,
     testStatus,
     testResponse,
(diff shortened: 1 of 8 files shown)


e2corporation:
@likyh This current Connection Form is a controlled form design of specific Connection Properties. You are destroying nicely designed state variables so you can stuff state in to a Configuration Object meant for reading the Plugin Config? This is not the way, I am going to decline this PR. There is nothing wrong with the state props and setters designed for managing the connection form. If there is a need to consolidate these props into one object at a later time I will do so.

e2corporation:
@likyh Moreover, if you feel the need to create a separate form component which is built on dynamic inputs, you should make a separate component (`DynamicConnectionForm`) instead of erasing already working code and changing it's design strategy.

likyh:
> @likyh This current Connection Form is a controlled form design of specific Connection Properties. You are destroying nicely designed state variables so you can stuff state in to a Configuration Object meant for reading the Plugin Config? This is not the way, I am going to decline this PR. There is nothing wrong with the state props and setters designed for managing the connection form. If there is a need to consolidate these props into one object at a later time I will do so.

So If we would add a new input on connection pages, we must find all of the usages and pass them on several levels of components. I don't think it a good way.

likyh:
> @likyh Moreover, if you feel the need to create a separate form component which is built on dynamic inputs, you should make a separate component (`DynamicConnectionForm`) instead of erasing already working code and changing it's design strategy.

That's another PR.

e2corporation:
> > @likyh This current Connection Form is a controlled form design of specific Connection Properties. You are destroying nicely designed state variables so you can stuff state in to a Configuration Object meant for reading the Plugin Config? This is not the way, I am going to decline this PR. There is nothing wrong with the state props and setters designed for managing the connection form. If there is a need to consolidate these props into one object at a later time I will do so.
> 
> So If we would add a new input on connection pages, we must find all of the usages and pass them on several levels of components. I don't think it a good way.

Even if we had to add 5 more state props to a connection form's signature, it's still manageable. You are worried about the wrong things.

e2corporation (commented):
@likyh PR Declined. You have repeatedly proposed deleting controlled state setters because it's "easier" for you to manage. Having granular control over a form's controlled input props allows for more flexibility. If I find an immediate need to consolidate these properties into one structure I will do so, but not in the fashion you have proposed here.

PR #2926, opened 2022-09-01
feat: add projects selector fetching by api for gitlab when adding blueprint

### LINKED ISSUE #2861: [Feature][Config-UI] Implement Gitlab API Proxy Access on UI
### Search before asking

### Description

The BE Team has previously added **GitLab API Proxy support** (See PR #2705 for BE Implementation Details). Config-UI can now adopt this endpoint by creating a `useGitlab.jsx` Hook (Modeled after JIRA's API Proxy -- see (`hooks/useJIRA.jsx`) to access Gitlab Project Information. This will allow enhanced Project Selection for Gitlab by either **a)** using auto-completed searchable results so projects can be identified/selected quickly or **b)** Create a Dropdown Project Selector just like JIRA's implementation where a **MultiSelect** component is used. 

The caveat is that we are potentially lots of projects received by the API, so this multi-selector may not be equipped to handle all of that data being rendered in one cycle out-of-the-box, in which case we'll have to enhance that widget with inline pagination. I think option a) which is to upgrade the input tag component with auto-complete capability maybe the first route to try. 

**Important Note**: Another thing to consider is that GitHub may not have API proxy endpoint as yet (?), so for GitHub the project objects will still be user-driven, not api driven but internally those references will still be treated like an Object form so the project input component should be behave normally like it does today for GitHub (let's call that "pure mode"), but for GitLab we'll want to use "api-mode" since we have live data backing it.

### Core Development Tasks

- [ ] Create and Implement new Gitlab API Hook  (`hooks/useGitlab.jsx`)
- [ ] Update Project Selector Form Component + Auto-complete Tag Input and/or Searchable Select
- [ ] Update Related Hooks
- [ ] Alter Validation Rules
- [ ] Refactor `configuredProject` to be a formal **Object** since these selections will be data-driven  
   - [ ] Upgrade/Enhance Workflow Components
   - [ ] Update Gitlab Provider Input Widget to use _API_ Mode
   - [ ] Keep Gitlab Provider Input Widget "as-is" in _Pure_ Mode
- [ ] Restrict API Resource Fetching to Gitlab Provider **ONLY**
- [ ] Test Blueprint Create Workflow + GITLAB

### Milestone
`TBD` Potentially `v0.14.0`

### BlueprintJS Documentation / Resources (V3)
Here are doc references to the core underlying components that will be used to implement this feature.

**MultiSelect**
https://blueprintjs.com/docs/versions/3/#select/multi-select

**TagInput**
https://blueprintjs.com/docs/versions/3/#core/components/tag-input

### Use case

Users creating Blueprints with a Gitlab connection now have a formal means of selecting projects available on the Gitlab Platform. Currently users are required to manually enter project-ids after having looking them up themselves manually by leaving Config-UI if they don't already have the information on hand. The idea is to enrich the Gitlab user experience by making it easier to select their projects.

### Related issues

_No response_

### Code of Conduct

likyh:
# Summary

Add more user-friend project selector for project choosing when add blueprint for gitlab.

Closes #2861

### Screenshots
image
image

image
Gitlab has so many public projects. So add a checkbox to control if display these projects which user is a member of.
image
image
image

### DIFF
--- a/config-ui/src/components/blueprints/DataScopesGrid.jsx
+++ b/config-ui/src/components/blueprints/DataScopesGrid.jsx
@@ -156,7 +156,7 @@ const DataScopesGrid = (props) => {
                 >
                   {c.projects.map((project, pIdx) => (
                     <li key={`list-item-key-${pIdx}`}>
-                      {project}
+                      {project.title}
                     </li>
                   ))}
                 </ul>
@@ -169,8 +169,8 @@ const DataScopesGrid = (props) => {
                     padding: 0,
                   }}
                 >
-                  {c.boards.map((board, bIdx) => (
-                    <li key={`list-item-key-${bIdx}`}>{board}</li>
+                  {c.boardsList.map((board, bIdx) => (
+                    <li key={`list-item-key-${bIdx}`}>{board.title}</li>
                   ))}
                 </ul>
               )}
--- a/config-ui/src/components/blueprints/DataScopesGrid.jsx
+++ b/config-ui/src/components/blueprints/DataScopesGrid.jsx
@@ -169,8 +169,8 @@ const DataScopesGrid = (props) => {
                     padding: 0,
                   }}
                 >
-                  {c.boardsList.map((board, bIdx) => (
-                    <li key={`list-item-key-${bIdx}`}>{board.title}</li>
+                  {c.boards.map((board, bIdx) => (
+                    <li key={`list-item-key-${bIdx}`}>{board}</li>
                   ))}
                 </ul>
               )}
--- a/config-ui/src/components/blueprints/DataScopesGrid.jsx
+++ b/config-ui/src/components/blueprints/DataScopesGrid.jsx
@@ -155,7 +155,7 @@ const DataScopesGrid = (props) => {
                   }}
                 >
                   {c.projects.map((project, pIdx) => (
-                    <li key={`list-item-key-${pIdx}`}>
+                    <li key={`list-item-key-${pIdx}`} style={{ whiteSpace: 'break-spaces' }}>
                       {project.title}
                     </li>
                   ))}
--- a/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
+++ b/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
@@ -21,8 +21,8 @@ import { MultiSelect } from '@blueprintjs/select'
 
 const GitlabProjectsSelector = (props) => {
   const {
-    fetchGitlabProjects = () => [],
-    isFetchingGitlab = false,
+    fetchGitlabProjects: onFetch = () => [],
+    isFetchingGitlab: isFetching = false,
     configuredConnection,
     placeholder = 'Select Projects',
     items = [],
@@ -68,24 +68,12 @@ const GitlabProjectsSelector = (props) => {
   const [onlyQueryMemberRepo, setOnlyQueryMemberRepo] = useState(true)
 
   useEffect(() => {
-    if (query.length <= 2) {
-      // only search when type more than 2 char or empty
-      return
-    }
     // prevent request too frequently
     const timer = setTimeout(() => {
-      fetchGitlabProjects(query, onlyQueryMemberRepo)
+      onFetch(query, onlyQueryMemberRepo)
     }, 200)
     return () => clearTimeout(timer)
-  }, [fetchGitlabProjects, query])
-
-  useEffect(() => {
-    if (query.length <= 2) {
-      // only search when type more than 2 char or empty
-      return
-    }
-    fetchGitlabProjects(query, onlyQueryMemberRepo)
-  }, [fetchGitlabProjects, onlyQueryMemberRepo])
+  }, [onFetch, query, onlyQueryMemberRepo])
 
   return (
     <div
@@ -118,8 +106,8 @@ const GitlabProjectsSelector = (props) => {
             },
           }}
           noResults={
-            (query.length <= 2 && <MenuItem disabled={true} text='Please type more than 2 char to search.' />) ||
-            (isFetchingGitlab && <MenuItem disabled={true} text='Fetching...' />) ||
+            (query.length <= 2 && <MenuItem disabled={true} text='Please type more than 2 characters to search.' />) ||
+            (isFetching && <MenuItem disabled={true} text='Fetching...' />) ||
             <MenuItem disabled={true} text='No Projects Available.' />
           }
           onRemove={(item) => {
@@ -149,7 +137,7 @@ const GitlabProjectsSelector = (props) => {
         />
 
         <Checkbox
-          label='only search repos joined' checked={onlyQueryMemberRepo}
+          label='Only search my repositories' checked={onlyQueryMemberRepo}
           onChange={e => setOnlyQueryMemberRepo(!onlyQueryMemberRepo)}
           style={{ margin: '10px 0 0 6px' }}
         />
--- a/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
+++ b/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
@@ -61,7 +61,7 @@ const GitlabProjectsSelector = (props) => {
         }}
       />
     ),
-    tagRenderer = (item) => item?.title,
+    tagRenderer = (item) => item.shortTitle || item.title
   } = props
 
   const [query, setQuery] = useState('')
--- a/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
+++ b/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
@@ -15,12 +15,10 @@
  * limitations under the License.
  *
  */
-import React, { useEffect, useMemo, useState } from 'react'
-import {
-  Intent,
-  MenuItem,
-} from '@blueprintjs/core'
-import { MultiSelect, Select } from '@blueprintjs/select'
+import React, { useEffect, useState } from 'react'
+import { Checkbox, Intent, MenuItem, } from '@blueprintjs/core'
+import { MultiSelect } from '@blueprintjs/select'
+
 const GitlabProjectsSelector = (props) => {
   const {
     fetchGitlabProjects = () => [],
@@ -67,14 +65,27 @@ const GitlabProjectsSelector = (props) => {
   } = props
 
   const [query, setQuery] = useState('')
+  const [onlyQueryMemberRepo, setOnlyQueryMemberRepo] = useState(true)
 
   useEffect(() => {
+    if (query.length <= 2) {
+      // only search when type more than 2 char or empty
+      return
+    }
     // prevent request too frequently
     const timer = setTimeout(() => {
-      fetchGitlabProjects(query)
+      fetchGitlabProjects(query, onlyQueryMemberRepo)
     }, 200)
     return () => clearTimeout(timer)
   }, [fetchGitlabProjects, query])
+
+  useEffect(() => {
+    if (query.length <= 2) {
+      // only search when type more than 2 char or empty
+      return
+    }
+    fetchGitlabProjects(query, onlyQueryMemberRepo)
+  }, [fetchGitlabProjects, onlyQueryMemberRepo])
 
   return (
     <div
@@ -107,9 +118,9 @@ const GitlabProjectsSelector = (props) => {
             },
           }}
           noResults={
-            isFetchingGitlab
-              ? <MenuItem disabled={true} text='Fetching' />
-              : <MenuItem disabled={true} text='No Projects Available.' />
+            (query.length <= 2 && <MenuItem disabled={true} text='Please type more than 2 char to search.' />) ||
+            (isFetchingGitlab && <MenuItem disabled={true} text='Fetching...' />) ||
+            <MenuItem disabled={true} text='No Projects Available.' />
           }
           onRemove={(item) => {
             onRemove((rT) => {
@@ -135,6 +146,11 @@ const GitlabProjectsSelector = (props) => {
             })
           }}
           style={{ borderRight: 0 }}
+        />
+
+        <Checkbox
+          label='only search repos joined' checked={onlyQueryMemberRepo}
+          onChange={e => setOnlyQueryMemberRepo(!onlyQueryMemberRepo)}
         />
       </div>
     </div>
--- a/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
+++ b/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
@@ -151,6 +151,7 @@ const GitlabProjectsSelector = (props) => {
         <Checkbox
           label='only search repos joined' checked={onlyQueryMemberRepo}
           onChange={e => setOnlyQueryMemberRepo(!onlyQueryMemberRepo)}
+          style={{ margin: '10px 0 0 6px' }}
         />
       </div>
     </div>
--- a/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
+++ b/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
@@ -21,8 +21,8 @@ import { MultiSelect } from '@blueprintjs/select'
 
 const GitlabProjectsSelector = (props) => {
   const {
-    fetchGitlabProjects: onFetch = () => [],
-    isFetchingGitlab: isFetching = false,
+    onFetch = () => [],
+    isFetching = false,
     configuredConnection,
     placeholder = 'Select Projects',
     items = [],
--- a/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
+++ b/config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
@@ -1,0 +1,144 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import React, { useEffect, useMemo, useState } from 'react'
+import {
+  Intent,
+  MenuItem,
+} from '@blueprintjs/core'
+import { MultiSelect, Select } from '@blueprintjs/select'
+const GitlabProjectsSelector = (props) => {
+  const {
+    fetchGitlabProjects = () => [],
+    isFetchingGitlab = false,
+    configuredConnection,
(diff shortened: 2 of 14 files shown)


config-ui/src/components/blueprints/DataScopesGrid.jsx
```diff
@@ -169,8 +169,8 @@ const DataScopesGrid = (props) => {
                     padding: 0,
                   }}
                 >
-                  {c.boards.map((board, bIdx) => (
-                    <li key={`list-item-key-${bIdx}`}>{board}</li>
+                  {c.boardsList.map((board, bIdx) => (
```
e2corporation, line 172: This was correctly set to `c.boards`. `c.boards`, represents the _stored/selected_ boards attached to the blueprint. `c.boardsList` represents all _available_ boards from the JIRA API.

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
```diff
@@ -0,0 +1,161 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import React, { useEffect, useState } from 'react'
+import { Checkbox, Intent, MenuItem, } from '@blueprintjs/core'
+import { MultiSelect } from '@blueprintjs/select'
+
+const GitlabProjectsSelector = (props) => {
```
e2corporation, line 22: I think what you've built can be re-used in the future for GitHub API Proxy as well, with the only difference being the API model data driving the "project" entity. I think we should call this `LiveProjectsSelector` or `ApiProjectsSelector` and not make it a Gitlab specific component.

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
```diff
@@ -0,0 +1,161 @@
...
+          marginBottom: '2px',
+          fontWeight: items.includes(item) ? 700 : 'normal',
+        }}
+      />
+    ),
+    tagRenderer = (item) => item?.title,
+  } = props
+
+  const [query, setQuery] = useState('')
+  const [onlyQueryMemberRepo, setOnlyQueryMemberRepo] = useState(true)
```
e2corporation, line 68: This prop being named `isPrivate` would be more fitting.

```suggestion
  const [isPrivate, setIsPrivate] = useState(true)
```

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
```diff
@@ -0,0 +1,161 @@
...
+ * limitations under the License.
+ *
+ */
+import React, { useEffect, useState } from 'react'
+import { Checkbox, Intent, MenuItem, } from '@blueprintjs/core'
+import { MultiSelect } from '@blueprintjs/select'
+
+const GitlabProjectsSelector = (props) => {
+  const {
+    fetchGitlabProjects = () => [],
```
e2corporation, line 24: This should be refactored to something more generic to the component, `onFetch` or `onProjectsFetch`

```suggestion
    onFetch = () => [],
```

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
```diff
@@ -0,0 +1,161 @@
...
+    isFetchingGitlab = false,
```
e2corporation, line 25: Refactor to `isFetching`, it can be aliased as isFetchingGitlab at a higher level when this component is used.

```suggestion
    isFetching = false,
```

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
```diff
@@ -0,0 +1,161 @@
...
+      />
+    ),
+    tagRenderer = (item) => item?.title,
+  } = props
+
+  const [query, setQuery] = useState('')
+  const [onlyQueryMemberRepo, setOnlyQueryMemberRepo] = useState(true)
+
+  useEffect(() => {
+    if (query.length <= 2) {
```
e2corporation, line 71: The min query check should be handled in the `useGitlab` Hook for better control, add it as an `if-else` condition to the existing chain and `Throw` a new error that query length is too short (like we do when connection id is null). Then it can be removed from being duplicated here.

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
```diff
@@ -0,0 +1,161 @@
...
+    }
+    // prevent request too frequently
+    const timer = setTimeout(() => {
+      fetchGitlabProjects(query, onlyQueryMemberRepo)
+    }, 200)
+    return () => clearTimeout(timer)
+  }, [fetchGitlabProjects, query])
+
+  useEffect(() => {
+    if (query.length <= 2) {
```
e2corporation, line 83: See Comment on `Ln 71` about migrating this check to the `useGitlab` hook.

config-ui/src/hooks/useDataScopesManager.jsx
```diff
@@ -137,7 +138,8 @@ function useDataScopesManager ({ provider, blueprint, /* connection, */ settings
           newScope = projects[connection.id]?.map((p) => ({
             ...newScope,
             options: {
-              projectId: Number(p),
+              projectId: Number(p.value),
```
e2corporation, line 141: Use optional chaining check here (`?`)
```suggestion
              projectId: Number(p?.value),
```

config-ui/src/hooks/useDataScopesManager.jsx
```diff
@@ -154,8 +156,8 @@ function useDataScopesManager ({ provider, blueprint, /* connection, */ settings
           newScope = projects[connection.id]?.map((p) => ({
             ...newScope,
             options: {
-              owner: p.split('/')[0],
-              repo: p.split('/')[1],
+              owner: p.value.split('/')[0],
```
e2corporation, line 159: Use optional chaining check here for both owner and repo (`?`)
```suggestion
              owner: p.value?.split('/')[0],
```

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
```diff
@@ -0,0 +1,161 @@
...
+                    ],
+                  }
+                : { ...rT }
+            })
+          }}
+          style={{ borderRight: 0 }}
+        />
+
+        <Checkbox
+          label='only search repos joined' checked={onlyQueryMemberRepo}
```
e2corporation, line 152: If refactored, this would change to `isPrivate`. Label alternatives might be `Only search my repositories` or `Only private repositories`

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
```diff
@@ -0,0 +1,161 @@
...
+          itemRenderer={itemRenderer}
+          tagRenderer={tagRenderer}
+          tagInputProps={{
+            tagProps: {
+              intent: Intent.PRIMARY,
+              minimal: true
+            },
+          }}
+          noResults={
+            (query.length <= 2 && <MenuItem disabled={true} text='Please type more than 2 char to search.' />) ||
```
e2corporation, line 121:
```suggestion
            (query.length <= 2 && <MenuItem disabled={true} text='Please type more than 2 characters to search.' />) ||
```

e2corporation:
@likyh Thanks for getting this feature lifted, overall looking good, I may add some additional feedback after some functional/visual tests.

config-ui/src/pages/blueprints/create-blueprint.jsx
```diff
@@ -241,14 +243,27 @@ const CreateBlueprint = (props) => {
...
   )
 
+  const {
+    fetchProjects: fetchGitlabProjects,
+    projects: gitlabProjects,
+    isFetching: isFetchingGitlab,
+    error: gitlabProxyError,
+  } = useGitlab(
+    {
+      apiProxyPath: JIRA_API_PROXY_ENDPOINT,
```
e2corporation, line 261: I think you intended to use Gitlab's API Proxy path instead?
```suggestion
      apiProxyPath: GITLAB_API_PROXY_ENDPOINT,
```

config-ui/src/pages/blueprints/blueprint-settings.jsx
```diff
@@ -288,14 +290,27 @@ const BlueprintSettings = (props) => {
...
   )
 
+  const {
+    fetchProjects: fetchGitlabProjects,
+    projects: gitlabProjects,
+    isFetching: isFetchingGitlab,
+    error: gitlabProxyError,
+  } = useGitlab(
+    {
+      apiProxyPath: JIRA_API_PROXY_ENDPOINT,
```
e2corporation, line 308:
```suggestion
      apiProxyPath: GITLAB_API_PROXY_ENDPOINT,
```

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: We can rename it when github added.

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: No, if you join in a public repo. it also will be queried.

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: ok

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: ok

config-ui/src/hooks/useDataScopesManager.jsx
likyh: We shouldn't use `?.` too much if we could know p is not null. Because it will cover up some bugs such as p is not a object.

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: ok

config-ui/src/components/blueprints/DataScopesGrid.jsx
likyh: Maybe we can delete `c.boards` because it is just a title array. But projects and `boards in create-blueprint are object arrays. Maybe we can rename it to boards and use it as object array.

config-ui/src/hooks/useDataScopesManager.jsx
e2corporation: It's not about covering up bugs, for vars that are dynamic when components are unmounted/mounted react will throw a warning unnecessarily. Optional chaining will allow code-flow to continue even though we don't expect or prefer p to be null. Any valid bug will present itself in the UI.

config-ui/src/components/blueprints/DataScopesGrid.jsx
e2corporation: No, for Advanced mode it's just a array of board titles, it gets mapped and converted to Object boards for normal Mode. These props will get revised in another ticket for now, there are different board props for easier use, boardIds it's own prop array etc.

config-ui/src/hooks/useDataScopesManager.jsx
likyh: Here p is from projects. projects maybe is an empty array. But it won't be a null array.

config-ui/src/components/blueprints/DataScopesGrid.jsx
likyh: > No, for Advanced mode it's just a array of board titles, it gets mapped and converted to Object boards for normal Mode. These props will get revised in another ticket for now, there are different board props for easier use, boardIds it's own prop array etc.

Sometimes boards is string array, sometimes is object array????? No, I don't think it a good practice.

e2corporation: Let's not go down this road where you overreact about mixed Types again, on my Refactor PR these are all wrapped with Data Models (DataScopeConnection). If the backend provided the same data structure and capabilities for normal and advanced blueprints to begin with there wouldn't be a need for this deviance of mixed types. Moreover, the backend wasn't storing any details regarding the JIRA Api.

e2corporation: Whether you interact with an array of Objects, Strings or Integers, it shouldn't be a problem for you. The differences and needs for rendering in Normal Mode vs Advanced mode created the scenario for mixed types, not so much the lack of good practices. Making them Objects in all scenarios wouldn't be a problem either and things will most likely gravitate to that anyway.

config-ui/src/hooks/useDataScopesManager.jsx
e2corporation: Leave chaining off for now and we'll see if React throws an error in dev or built mode.

config-ui/src/components/blueprints/DataScopesGrid.jsx
e2corporation: Had I known backend is storing everything submitted in transformations as raw json we could have potentially stored board metadata much sooner and not have to refetch it from Proxy all the time. However if the board data is updated by the user it would be out of sync and we'd still have to re-associate the stored board IDs against the live list to ensure the metadata was up to date.

config-ui/src/pages/blueprints/blueprint-settings.jsx
```diff
@@ -579,20 +590,45 @@ const BlueprintSettings = (props) => {
...
+        key: s.options?.projectId,
+        value: s.options?.projectId,
+        title: s.options?.title || `Project ${s.options?.projectId}`,
+      }))
       : []
     // @todo: handle multi-stage
     const getAdvancedGithubProjects = (t, providerId) => [Providers.GITHUB].includes(providerId)
-      ? [`${t.options?.owner}/${t.options?.repo}`]
+      ? [{
+          id: `${t.options.owner}/${t.options?.repo}`,
```
e2corporation, line 611: We should add optional chaining (`?`) for `owner` here too like all the other props.
```suggestion
          id: `${t.options?.owner}/${t.options?.repo}`,
```

config-ui/src/pages/blueprints/blueprint-settings.jsx
likyh: fixed

config-ui/src/pages/blueprints/blueprint-settings.jsx
likyh: fixed

config-ui/src/hooks/useDataScopesManager.jsx
likyh: No error in console after add blueprint and edit blueprint except this. Because of #2994, I cannot go to other page.

image

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: fixed

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: fixed

config-ui/src/components/blueprints/DataScopesGrid.jsx
likyh: Now connection.boards haven't been used. So we delete boards here ?

e2corporation: DataScopes grid should not be rendering `boardsList` here, it should be rendering `boards`.

config-ui/src/hooks/useDataScopesManager.jsx
e2corporation: #2994 looks to be a new issue, I still need to local test this PR.

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
e2corporation: I don't the argument props definitions shouldn't be aliased, 

```
    fetchGitlabProjects: onFetch = () => [],
     isFetchingGitlab: isFetching = false,
```

should just be:

```
    onFetch = () => [],
    isFetching: isFetching = false,
```


the alias can be done when GitlabProjectsSelector is implemented (isFetching and fetchGitlabProjects are retrieved from the Gitlab Hook when hook is initialized, and aliased during de-structuring)

```
<GitlabProjectsSelector isFetching={isFetchingGitlab) ...
```

config-ui/src/components/blueprints/DataScopesGrid.jsx
e2corporation: More refinements and property consolidation on the data scope connection model will be done later, especially with regard to duplicated props. Ideally we want to build the same data scope connection for both advanced and normal, due to the differences in how advanced data is referenced currently these "connection" objects constructed in different ways even though they are sharing the same model.

Regardless of this, `boards` has always referenced _selected_ boards and `boardsList` anywhere in the code base has always reflected the _available_ list of boards.

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: ok

config-ui/src/components/blueprints/GitlabProjectsSelector.jsx
likyh: fixed. Now judge on `useGitlab`. if length<=2, projects will be set to `[]`.

config-ui/src/components/blueprints/DataScopesGrid.jsx
likyh: reverted

image

image

e2corporation: @likyh I think the ability to display the Named board in Advanced mode is still possible, the JIRA resources would need to be fetched and then boards can be mapped like we do in normal mode. However, after your recent changes Board $title is being stored on the transformation record. So In advanced mode we could just reference the stored Board Title instead of having to fetch.

likyh: @e2corporation On main branch, it like this too. Jira only fetch when open Data Scope Modal. But advance mode doesn't have it.

likyh:
All of review if fixed. Can it merge? @e2corporation

e2corporation:
@likyh Functional tests looking good, the selector operates nicely so far. I think there are some visual tweaks we may want to consider. I do like that Full Repository Path is shown, but maybe we want to consider shortening the nodes displayed to just the repo name and it's immediate parent for multi-level scenarios? **-or-** we need some CSS tweaks on the Stacked Project List selector to handle lengthy project paths, mainly by css truncation+ellipsis to prevent a rendering bug.

The other useful thing that can be done, is for the TagItem renderer callback we can customize the selector tags that get inserted into the widget. We could wrap them with a tooltip or popover that contains the full repository path and other project details. This way we can just show the Project Name alone for the tag.

Screen Shot 2022-09-07 at 9 57 56 AM

Screen Shot 2022-09-07 at 9 57 48 AM

e2corporation:
> All of review if fixed. Can it merge? @e2corporation

@likyh It's looking good and will be merge ready soon, see my last comment about Long Project Name/Path.

likyh:
@e2corporation Ok, I prepare to hide left chars in title add add a popover to show full text.

likyh:
image
image
image

likyh:
image
image
image
image
image
image
That's fix.

e2corporation (approved):
@likyh Nice work, thanks for making the updated revisions. LGTM (after rebase).

PR #2950, opened 2022-09-05
Fix#2935

### LINKED ISSUE #2935: [Bug][jira] list jira connections occur error when users use a new postgres database
### Search before asking

### What happened

ERROR: cached plan must not change result type (SQLSTATE 0A000)

list connection suceess

### How to reproduce

caJaJ9kvf2

### Anything else

_No response_

### Version

main

### Code of Conduct

abeizn:
# Summary
old: rename a->a
new: rename a->b->a

Closes #2935 

### Screenshots

### Other Information


plugins/jira/models/migrationscripts/20220716_add_init_tables.go
```diff
@@ -99,8 +117,15 @@ func (*addInitTables) Up(ctx context.Context, db *gorm.DB) error {
 	}
 	defer db.Migrator().DropTable(&JiraConnectionTemp{})
 
+	// cached plan must not change result type in postgres, so need to update db name
+	err = db.Migrator().RenameTable(JiraConnectionV011{}, JiraConnectionV011Bak{})
```
warren830, line 121: if you just want to change table name, I think it'd better delete all fields in `JiraConnectionV011`

plugins/jira/models/migrationscripts/20220716_add_init_tables.go
```diff
@@ -99,8 +117,15 @@ func (*addInitTables) Up(ctx context.Context, db *gorm.DB) error {
...
+		return err
+	}
+	defer db.Migrator().DropTable(&JiraConnectionV011Bak{})
+
 	var result *gorm.DB
-	var jiraConns []JiraConnectionV011
+	var jiraConns []JiraConnectionV011Bak
 	result = db.Find(&jiraConns)
 
 	if result.Error == nil {
```
warren830, line 131: Also, please do not use if like this, you can simply use `if result.Error != nil` and return err. Code like this make it difficult to understand

plugins/jira/models/migrationscripts/20220716_add_init_tables.go
abeizn: > Seems like we could pass the table name as `string` directly, will it be simpler? image

updated

warren830 (commented):
LGTM

plugins/jira/models/migrationscripts/20220716_add_init_tables.go
```diff
@@ -92,61 +94,79 @@ func (*addInitTables) Up(ctx context.Context, db *gorm.DB) error {
 		return err
 	}
 
-	// get connection history data
-	err = db.Migrator().CreateTable(&JiraConnectionTemp{})
+	// due to cached plan must not change result type in postgres,
+	// so need to use JiraConnectionOld, JiraConnectionNew and JiraConnection to operate.
```
klesh, line 98: In order to avoid postgres reporting "cached plan must not change result type", we have to avoid loading the `_tool_jira_connections` table before our migration. The trick is to rename it before loading.

plugins/jira/models/migrationscripts/20220716_add_init_tables.go
```diff
@@ -92,61 +94,79 @@ func (*addInitTables) Up(ctx context.Context, db *gorm.DB) error {
...
 	}
 
-	err = db.Migrator().DropTable(&JiraConnectionV011{})
-	if err != nil {
-		return err
-	}
-
-	err = db.Migrator().RenameTable(JiraConnectionTemp{}, archived.JiraConnection{})
+	// step4
+	err = db.Migrator().RenameTable(JiraConnectionNew{}, archived.JiraConnection{})
```
warren830, line 169: I think we can directly use table name instead of archived.JiraConnection{}
Two reasons:
1. more intuitive 
2. gorm can do one less parse

warren830 (approved):
LGTM

PR #3038, opened 2022-09-09
fix: task failure caused by 404

### LINKED ISSUE #2960: [Bug][gitihub] collect account failed by not found user
### Search before asking

### What happened

we can found the user `congweihai` at.
`https://api.github.com/repos/panjf2000/ants/issues/events?direction=asc&page=7&per_page=100&state=all`
But failed when visit user path.
`https://api.github.com/users/congweihai`

It make the collect error.

work without error.

### How to reproduce

collect panjf2000/ants
or see the api return 
`https://api.github.com/repos/panjf2000/ants/issues/events?direction=asc&page=7&per_page=100&state=all`
`https://api.github.com/users/congweihai`

### Anything else

_No response_

### Version

main

### Code of Conduct

mindlesscloud:
# Summary
fix #2960 [Bug][gitihub] collect account failed by not found user
In the case of `err == ErrIgnoreAndContinue`, the `err` should not be wrapped, because on the caller side, the expression `err == ErrIgnoreAndContinue`  would evaluate as false.
```go
if apiClient.afterResponse != nil {
		err = apiClient.afterResponse(res)
		if err != nil && err != ErrIgnoreAndContinue {
			res.Body.Close()
			return nil, errors.Default.Wrap(err, fmt.Sprintf("error running afterRequest for %s", req.URL.String()), errors.UserMessage("error after making API call"))
		}
	}
```
Closes #2960 

### Screenshots

### Other Information

### DIFF
--- a/plugins/helper/api_client.go
+++ b/plugins/helper/api_client.go
@@ -265,7 +265,11 @@ func (apiClient *ApiClient) Do(
 	// after receive
 	if apiClient.afterResponse != nil {
 		err = apiClient.afterResponse(res)
-		if err != nil && err != ErrIgnoreAndContinue {
+		if err == ErrIgnoreAndContinue {
+			res.Body.Close()
+			return res, err
+		}
+		if err != nil {
 			res.Body.Close()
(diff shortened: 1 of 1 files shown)


abeizn (approved):
LGTM

PR #3032, opened 2022-09-09
fix: golangci-lint error

klesh:
# Summary

golint-ci was down for a while, this PR try to fix all missed linting errors during the period.

### Screenshots
image

PR #3040, opened 2022-09-09
fix: let workflow `check-Apache-license-header` work

thenicetgp:
### 2. By the way, added the apache license header to the files that need to be added.
Screen Shot 2022-09-09 at 22 30 09

### DIFF
--- a/.github/workflows/asf-header-check.yml
+++ b/.github/workflows/asf-header-check.yml
@@ -27,4 +27,4 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Check License Header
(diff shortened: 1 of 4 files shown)


thenicetgp:
> #3032 resolved all linter problems including the license headers, so I think we can close this one.

Tanks for your remind, I undid the commit that added headers to the required files. But I found that the workflow `check Apache license header` in the PR #3032 still doesn't work, so I think this PR is still useful. At the same time,I found there are still two files to add `apache-license-header`.
Screen Shot 2022-09-10 at 23 19 59

Screen Shot 2022-09-10 at 23 30 44

mindlesscloud (approved):
LGTM

PR #3047, opened 2022-09-11
fix: remove invalid test case TestReadConfig

klesh:
# Summary

The `TestReadConfig` is highly coupled with an external setting, which is out of the scope of the test case.
This makes it an ineffectual test case.


likyh (approved):
LGTM

mappjzc (approved):

PR #3099, opened 2022-09-16
fix: non-empty body of get request

### LINKED ISSUE #3098: [Bug][ApiAsyncClient] The GET request sent by `ApiAsyncClient.DoAsync` contains a body
### Search before asking

### What happened

All the GET requests sent by `ApiAsyncClient.DoAsync` contained a body consisting of 4 characters `null`

The GET request should not contain any body

### How to reproduce

set breakpoint and check the value of the variable
image
image

### Anything else

_No response_

### Version

main

### Code of Conduct

mindlesscloud:
# Summary
The argument `body` was shadowed by another variable with an identical name. The method `apiClient.Do` would always receive an empty `[]byte` which turns into a string request body `null`. To fix the problem, the variable `body` was renamed to `respBody` which was supposed to hold the response body.
```go
func (apiClient *ApiAsyncClient) DoAsync(..., body interface{},...) {
	...
	request = func() errors.Error {
		...
		var body []byte
		res, err = apiClient.Do(method, path, query, body, header)
		...
	}
	...
}
```	

Closes #3098 

### Screenshots

### Other Information


warren830 (approved):
LGTM

Issue #2078, opened 2022-06-03
Relocate files/dirs


klesh:
## What and why to refactor
Many folders were randomly located in the wrong folder, and sometimes, they caused `cycle import` error,
and the root folder is bloated

## Describe the solution you'd like
- #2156
- remove `scripts` folder
- move `CONTRIBUTING.md` to `lake-website`
- move `plugins/*.md` to `lake-website`
- move `k8s-deploy.yml` to `releases` and update documentation
- rename `plugin_db_migration.go` to `migratable`
- remove `e2e` folder
- remove `plugins/helper/common/callbacks.go`

To reorganize the packages/folders to the following structure.
![dependency-layers drawio](https://user-images.githubusercontent.com/61080/172016216-c6adaf31-4200-476a-b12a-b6e27dbd813c.svg)

## Document

- Add the new structure image and description to Developer Document

keon94:
bump

keon94:
@klesh 
1. Do you have any suggestions for where any framework-level python code should go (from #3701)? Should we create a top level directory called "python" and throw all *.py files in there? Or should we place them next to the *.go files at the appropriate places? (I personally prefer the first option because it keeps things well-isolated)
2. I've used the scripts folder for the model generator of Singer taps (#3123). Do we still want to remove it?

klesh:
> @klesh
> 
> 1. Do you have any suggestions for where any framework-level python code should go (from [[Feature][Python plugin SDK] Support plugin development in Python #3701](https://github.com/apache/incubator-devlake/issues/3701))? Should we create a top level directory called "python" and throw all *.py files in there? Or should we place them next to the *.go files at the appropriate places? (I personally prefer the first option because it keeps things well-isolated)
> 2. I've used the scripts folder for the model generator of Singer taps ([[Feature][Framework][POC] Streamed data collection using Singer Spec #3123](https://github.com/apache/incubator-devlake/issues/3123)). Do we still want to remove it?
> 
> cc @CamilleTeruel

@keon94
1. I think the top-level directory makes sense. Everything related to python-support should be put into it including *.go files if they are meant for python plugin support
2. I don't have an opinion on this, it is up to you to judge.

keon94:
[...]

PR #3151, opened 2022-09-20
[issue-3123][backend] Singer-spec framework support

### LINKED ISSUE #3123: [Feature][Framework][POC] Streamed data collection using Singer Spec
### Search before asking

### Description

Do a POC on adding support for streamed, incremental data collections using the [Singer spec](https://www.singer.io/). The existing incremental update support does not protect against a half-way failure or error in the collection process, forcing the user to have to start all over again. Specifically, this POC will focus on creating a new plugin based on an existing one that supports the Singer spec.

### Use case

The goal would be to add resilience to the collection phase of devlake plugins, and to do so using a standard. Currently if there's an interruption in the middle of collection we don't have a means of recovery/resuming from that interrupted point. This becomes a problem when pulling data from large data sources. We can alleviate this problem by introducing incremental states as the collection takes place. If, say, we have to make 10 API calls to perform 10 DB writes for a given collector, and a failure happens on the 5th invocation, our next attempt should try to pick the collection back up from that step.

### Related issues

_No response_

PR: #3151 

### Code of Conduct

keon94:
This PR adds support for abstract Taps, using a Golang interface, wrapped by a "Tap Extractor". The default implementation is for Singer-Taps. Using this scheme, the collector layer is completely eliminated (done by the taps instead). The design has focused on keeping things as consistent as possible with the current architecture.

Closes #3123 

### Screenshots
Tap state table
image

Sample Tap-Extractor skeleton code
image

Plugin entry-code:
image

### Other Information


keon94:
POC plugin implementations moved to https://github.com/merico-dev/lake/tree/issues/3123-poc

plugins/core/tap/singer_impl.go
```diff
@@ -0,0 +1,196 @@
...
+		stateFile      *fileData[[]byte]
+		configFile     *fileData[[]byte]
+	}
+	fileData[Content any] struct {
+		path    string
+		content *Content
+	}
+)
+
+func NewSingerTap(cfg *SingerTapConfig) (*SingerTapImpl, errors.Error) {
```
klesh, line 47: Please do NOT put implementation under `core`, move it to `impl/tap`

plugins/core/tap/singer_models.go
```diff
@@ -0,0 +1,45 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tap
+
+// singer-tap specific types
+type (
+	// SingerTapSchema the structure of this is determined by the catalog/properties JSON of a singer tap
+	SingerTapSchema map[string]interface{}
```
klesh, line 23: Please move it to `models` if they are

models/migrationscripts/20221015_create_singer_state.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"github.com/apache/incubator-devlake/errors"
+	"gorm.io/datatypes"
+	"gorm.io/gorm"
+	"time"
+)
+
+type createTapState struct{}
+
+type SingerTapState20221015 struct {
```
klesh, line 30: Where is the final `model struct` located?

klesh (commented):
I have some questions:
1. Do we have the progress update for this kind of subtasks?
2. Can we cancel the operation?
3. Can we collect the error message reported by the tap
4. Should we update the `Dockerfile` to include the `tap` binary?

keon94:
> I have some questions:
> 
> 1. Do we have the progress update for this kind of subtasks?
> 2. Can we cancel the operation?
> 3. Can we collect the error message reported by the tap
> 4. Should we update the `Dockerfile` to include the `tap` binary?

1. Yes See [this](https://github.com/apache/incubator-devlake/pull/3151/files#diff-77ec73334c6c15e351538bd12467875fa40665d75beb11627899363e071b7668R154)
2. Yes. See [this](https://github.com/apache/incubator-devlake/pull/3151/files#diff-77ec73334c6c15e351538bd12467875fa40665d75beb11627899363e071b7668R140)
3. Yes. Stderr of the tap is captured in the error
4. Once we support concrete taps in future tickets. At this moment, no.

models/migrationscripts/20221015_create_singer_state.go
keon94: [here](https://github.com/apache/incubator-devlake/pull/3151/files#diff-c1a46698f48456226e901578ec4d6bce4049b1c4c923b574300f3a5d60399abeR50)

plugins/core/tap/singer_impl.go
klesh: The one located in the project root folder, there are 2 implementations already, `dalgorm` and `migration`.
They both implemented the `interfaces` defined in `core` package.
In this way, we can separate the contract from the concrete implementation, which helps us avoid importing 3rd libraries directly into our system and writing test cases among other benefits.
Generic or not doesn't matter, we don't plan to implement another specific `migration` or `dal` as well.

Come to think about it, maybe `helpers/pluginhelper/tap` is a more appropriate place to be.
I think the whole `singer` feature is not the `core` of our system, it is sth we offer to developers, they could choose to use RESTful API or use `singer`, so, no, they should not be in the `core` package.

Take a look at the `plugins/helper` which contains all RESTFul helpers we offer for plugin developers, it was put into a separate package apart from the `core` package as an optional feature. I think the `singer` should do the same, except the existing `plugins/helper` will be moved to `helpers/pluginhelper/restful`

keon94: Ok, good idea. Wouldn't `helpers/restful` and `helpers/singer` not sufficient? Because they're already in the plugins parent directory.

plugins/core/tap/singer_models.go
keon94: I've kept them under helper/tap because these are mostly helper models, not high level models like the ones in that package

plugins/core/tap/singer_impl.go
keon94: I only made helpers/tap to keep this PR focused on just the new stuff. We can have another refactor PR to move the existing src files to helpers/restful (and update the generator)

klesh: I'm planning to move `plugins/helpers` to `helpers/pluginhelper` according to #2078 .
It would be helpful if you place the `tap` there.
`helpers` and `core` were put into `plugins/` folder without much consideration, they shouldn't be there, because, obviously, they are not plugin 😂

keon94: Done.

models/migrationscripts/20221015_create_singer_state.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (SingerTapState20221015) TableName() string {
+	return "tap_state"
+}
+
+func (*createTapState) Up(basicRes core.BasicRes) errors.Error {
+	err := basicRes.GetDal().AutoMigrate(SingerTapState20221015{})
+	if err != nil {
+		return errors.Convert(err)
```
klesh, line 45: I don't think we need the `errors.Convert` anymore

models/migrationscripts/20221015_create_singer_state.go
```diff
@@ -0,0 +1,56 @@
...
+	Id           string `gorm:"primaryKey;type:varchar(255)"`
+	ConnectionId uint64
+	Type         string
+	Value        datatypes.JSON
+	CreatedAt    time.Time
+	UpdatedAt    time.Time
+}
+
+func (SingerTapState20221015) TableName() string {
+	return "tap_state"
```
klesh, line 39: For framework table, it should be prefixed by `_devlake_`

plugins/helper/batch_save_divider.go
```diff
@@ -35,11 +35,12 @@ type BatchSaveDivider struct {
 	db        dal.Dal
 	batches   map[reflect.Type]*BatchSave
 	batchSize int
-	table     string
-	params    string
+	rawTable  string
```
klesh, line 38: I think we should revert this file if we all agreed raw layer is neccessary

keon94:
Sample collection raw later:
image

keon94:
New state table after revisions (cc @CamilleTeruel)
image

models/migrationscripts/20221015_create_singer_state.go
keon94: Done

models/migrationscripts/20221015_create_singer_state.go
keon94: Done

plugins/helper/batch_save_divider.go
keon94: Done

keon94:
Sample collector usage:
image
Sample plugin impl usage:
image

plugins/helper/api_rawdata.go
```diff
@@ -52,11 +52,12 @@ type RawDataSubTaskArgs struct {
 // RawDataSubTask is Common features for raw data sub-tasks
 type RawDataSubTask struct {
 	args   *RawDataSubTaskArgs
-	table  string
-	params string
+	Table  string
```
klesh, line 55: The reason these Fields are private is they should not be modified after initialization.
Any accidental modified can cause bugs.

Maybe we should move this file to `models` module since it is part of the infrastructure of the framework and shared by different helpers and change these Fields to Methods to avoid modification.

klesh (commented):
It looks good overall.
One last little thing and we can merge it

plugins/helper/api_rawdata.go
keon94: I think moving this file to a new package is a good idea, but that'll end up touching a lot of different files . I think for now I'll just add Getter methods to this struct and make the fields private again to keep this PR simple.

klesh (approved):
LGTM

PR #3165, opened 2022-09-22
fix(config-ui): adjust some content about dora

mintsweet:
### ⚠️ Pre Checklist

# Summary

1. Added environment mapping field.
2. Adjust the default value of the dora deployment field.

none.

### Screenshots
screenshot-20220922-144912

### Other Information
none.

### DIFF
--- a/config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
+++ b/config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
@@ -15,76 +15,45 @@
  * limitations under the License.
  *
  */
-import React, { useState, useEffect, useMemo, useCallback } from 'react'
+import React, { useState, useEffect } from 'react'
 import {
   Intent,
   FormGroup,
   RadioGroup,
   InputGroup,
   Radio,
-  Tag,
-  Icon,
-  Colors,
-  Tooltip
+  Tag
 } from '@blueprintjs/core'
 import { Providers, ProviderLabels } from '@/data/Providers'
+
 const Deployment = (props) => {
   const {
-    connection,
     provider,
     transformation,
-    configuredProject,
-    configuredBoard,
     entityIdKey,
-    entities = [],
     isSaving = false,
     onSettingsChange = () => {}
   } = props
 
-  const [deployTag, setDeployTag] = useState(
-    transformation?.productionPattern || ''
-  )
-  const [enableDeployTag, setEnableDeployTag] = useState(
-    [
-      transformation?.productionPattern
-      // transformation?.stagingPattern,
-      // transformation?.testingPattern
-    ].some((t) => t && t !== '')
-      ? 1
-      : 0
-  )
-  const isDeployTagEmpty = useMemo(
-    () =>
-      enableDeployTag &&
-      (transformation?.productionPattern === '' ||
-        !transformation?.productionPattern),
-    [enableDeployTag, transformation?.productionPattern]
-  )
+  const [selectValue, setSelectValue] = useState(1)
 
-  const clearDeploymentTags = useCallback(
-    (deploymentOption) => {
-      if (entityIdKey && deploymentOption === 0) {
-        onSettingsChange({ productionPattern: '' }, entityIdKey)
-        // onSettingsChange({ stagingPattern: '' }, entityIdKey)
-        // onSettingsChange({ testingPattern: '' }, entityIdKey)
-      }
-    },
-    [entityIdKey, onSettingsChange]
-  )
+  useEffect(() => {
+    setSelectValue(
+      transformation?.deploymentPattern ||
+        transformation?.deploymentPattern === ''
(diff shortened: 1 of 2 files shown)


abeizn (approved):
LGTM

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
```diff
@@ -202,95 +190,19 @@ const Deployment = (props) => {
...
-                          style={{ margin: '8px' }}
-                        />
-                      </Tooltip>
-                    ) : null
-                  }
-                  required
-                />
-              </FormGroup>
-            </div>
-            {/* <div className='formContainer'>
```
e2corporation, line 234: @mintsweet These tags are supposed to be re-activated in the future that's why this code was commented and left in tact. Why was it removed?

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
```diff
@@ -15,76 +15,45 @@
...
+  useEffect(() => {
+    setSelectValue(
+      transformation?.deploymentPattern ||
+        transformation?.deploymentPattern === ''
+        ? 1
+        : 0
+    )
+  }, [transformation?.deploymentPattern])
 
-  const handleDeploymentPreference = useCallback(
```
e2corporation, line 75: @mintsweet This handler should not have been refactored, with these changes you are now introducing `undefined` and `[object]` entries into the main Transformations stack.

Screen Shot 2022-09-22 at 9 40 03 AM

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
```diff
@@ -15,76 +15,45 @@
...
-          break
-        case 1:
-          break
-      }
-    },
-    [clearDeploymentTags]
-  )
+  const handleChangeSelectValue = (sv) => {
+    if (entityIdKey && sv === 0) {
+      onSettingsChange({ deploymentPattern: undefined }, entityIdKey)
```
e2corporation, line 51: `deploymentPattern` is a string field that is bound to the input, it should never be set or initialized to undefined.

Screen Shot 2022-09-22 at 9 38 57 AM

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
mintsweet: Sorry, I am a little bit of code cleanliness, I want to delete the commented content, I don't know if it will be used again, I will restore it.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
mintsweet: This is the result of asking the backend, deployPattern is empty by default, which means option 1, and option 2 does not need to pass deployPattern, so I set it to undefined.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
mintsweet: delopyPattern does not need to pass any value to the backend when option 2, so I set it to undefined.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
e2corporation: I would normally delete it as well, but this may save time in the future from having to re-create the inputs again.

mintsweet:
@e2corporation Today's teammates are anxious to test, so this PR has not been reviewed by you, I am very sorry.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
e2corporation: backend will treat `deploymenPattern === ""` differently than it being missing from the request payload? It should be safe to pass deploymentPattern as empty string if its not being used. General rule is, if a property is bound to a text input like this it should not have an undefined value.

e2corporation:
> @e2corporation Today's teammates are anxious to test, so this PR has not been reviewed by you, I am very sorry.

I understand, I am just being extra cautious as we are about to freeze so I don't want to introduce last minute bugs. We should have time to resolve any issues during our testing window.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
e2corporation: It's fine to leave it stripped for now.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
mintsweet: I will confirm this with my back-end colleagues today to ensure that no problems will be introduced.

PR #3176, opened 2022-09-23
Migration adopts `BasicRes`

### LINKED ISSUE #3097: [Feature][Framework] Create BasicRes in framwork
### Search before asking

### Description

Right now, we use `log, config, dal` by passing them to plugin.init() and wrap them in BasicRes

### Use case

We should create BasicRes in framework and use it anywhere. In this way, the structure of code will be more clean.
Also, we should use `dal` in migration scripts.

### Related issues

_No response_

### Code of Conduct

klesh:
# Summary

In order to isolate our code from dependencies for better modularity and testability, this refactor aims to remove the direct dependency to `gorm`.

By adopting `BasicRes` interface, we also gain the ability to read `config` from the system (like reading the encryption key for data migration), and a logger, which would make migration script debugging easier

Closes #3097


mindlesscloud (approved):
LGTM

PR #3191, opened 2022-09-26
fix(config-ui): adjust the parameters so the devlake can not execute DORA tasks

### LINKED ISSUE #3189: [Bug][Config-UI] Cannot configure to not execute DORA enrichTaskEnv subtask
### Search before asking

### What happened

DORA executes regardless of whether I have DORA configured or not.

1fc80657-9acf-46f8-b2d5-28d15f6990ef

Don't perform DORA related tasks when I don't have DORA set.

### How to reproduce

Do not configure DORA-related content and observe whether DORA-related tasks are executed.

### Anything else

_No response_

### Version

v0.13.0

### Code of Conduct

mintsweet:
# Summary

**deploymentPattern** and **productionPattern** are not passed when DORA is not required.

Closes #3189 

### Screenshots
**DORA is required**
screenshot-20220926-144804

**DORA is not required**
screenshot-20220926-144812

### Other Information

### DIFF
--- a/config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
+++ b/config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
@@ -22,6 +22,7 @@ import React, {
   useCallback,
   useMemo
 } from 'react'
+import { isEqual } from 'lodash'
 import {
   Button,
   Icon,
@@ -96,6 +97,13 @@ const DataTransformations = (props) => {
     cardStyle = {}
   } = props
 
+  // Used to determine whether to display edit transformation or add transformation
+  const [initializeTransformations, setInitializeTransformations] = useState({})
+
+  useEffect(() => {
+    setInitializeTransformations(transformations)
+  }, [])
+
   const isTransformationSupported = useMemo(
     () =>
       configuredProject ||
@@ -138,15 +146,20 @@ const DataTransformations = (props) => {
   )
   const [activeEntity, setActiveEntity] = useState()
 
-  const transformationHasProperties = useCallback(
+  const transformationHasChanged = useCallback(
     (item) => {
+      const initializeTransform = initializeTransformations[item?.id]
       const storedTransform = transformations[item?.id]
+
+      console.log(initializeTransform)
+      console.log(storedTransform)
       return (
+        initializeTransform &&
         storedTransform &&
-        Object.values(storedTransform).some((v) => v && v.length > 0)
+        !isEqual(initializeTransform, storedTransform)
       )
     },
-    [transformations]
+    [initializeTransformations, transformations]
   )
 
   useEffect(() => {
@@ -340,7 +353,7 @@ const DataTransformations = (props) => {
                           activeItem={configuredProject}
                           onAdd={addProjectTransformation}
                           onChange={addProjectTransformation}
-                          isEditing={transformationHasProperties}
+                          isEditing={transformationHasChanged}
                         />
                         {projects[configuredConnection.id].length === 0 && (
                           <NoData
@@ -365,7 +378,7 @@ const DataTransformations = (props) => {
                           activeItem={configuredBoard}
                           onAdd={addBoardTransformation}
                           onChange={addBoardTransformation}
-                          isEditing={transformationHasProperties}
+                          isEditing={transformationHasChanged}
                         />
                         {boards[configuredConnection.id].length === 0 && (
                           <NoData
--- a/config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
+++ b/config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
@@ -40,17 +40,25 @@ const Deployment = (props) => {
   useEffect(() => {
     setSelectValue(
       transformation?.deploymentPattern ||
-        transformation?.deploymentPattern === ''
+        transformation?.deploymentPattern === '' ||
+        transformation?.productionPattern ||
+        transformation?.productionPattern === ''
         ? 1
         : 0
     )
-  }, [transformation?.deploymentPattern])
+  }, [transformation?.deploymentPattern, transformation?.productionPattern])
 
   const handleChangeSelectValue = (sv) => {
     if (entityIdKey && sv === 0) {
-      onSettingsChange({ deploymentPattern: undefined }, entityIdKey)
+      onSettingsChange(
+        { deploymentPattern: undefined, productionPattern: undefined },
+        entityIdKey
+      )
     } else if (entityIdKey && sv === 1) {
-      onSettingsChange({ deploymentPattern: '' }, entityIdKey)
+      onSettingsChange(
+        { deploymentPattern: '', productionPattern: '' },
+        entityIdKey
+      )
     }
     setSelectValue(sv)
   }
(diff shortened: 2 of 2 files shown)


e2corporation:
@mintsweet Please remove the ":warning: Pre Checklist" Section and replace with an appropriate description if this PR is ready, otherwise tag it as a **Draft**

config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
```diff
@@ -96,6 +97,13 @@ const DataTransformations = (props) => {
     cardStyle = {}
   } = props
 
+  // Used to determine whether to display edit transformation or add transformation
+  const [initializeTransformations, setInitializeTransformations] = useState({})
```
e2corporation, line 101: This is not an ideal naming choice as we `InitializeTransformations` is already in use at the DSM Hook level for representing the global Transformations map. Additionally, this extra check is not necessary.

config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
```diff
@@ -138,15 +146,20 @@ const DataTransformations = (props) => {
   )
   const [activeEntity, setActiveEntity] = useState()
 
-  const transformationHasProperties = useCallback(
+  const transformationHasChanged = useCallback(
```
e2corporation, line 149: This transformation check handler was already working properly, no refactoring is needed here please revert this entire file.

config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
```diff
@@ -96,6 +97,13 @@ const DataTransformations = (props) => {
     cardStyle = {}
   } = props
 
+  // Used to determine whether to display edit transformation or add transformation
+  const [initializeTransformations, setInitializeTransformations] = useState({})
+
+  useEffect(() => {
```
e2corporation, line 103: This effect should be removed.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
```diff
@@ -40,17 +40,25 @@ const Deployment = (props) => {
...
         : 0
     )
-  }, [transformation?.deploymentPattern])
+  }, [transformation?.deploymentPattern, transformation?.productionPattern])
 
   const handleChangeSelectValue = (sv) => {
     if (entityIdKey && sv === 0) {
-      onSettingsChange({ deploymentPattern: undefined }, entityIdKey)
+      onSettingsChange(
+        { deploymentPattern: undefined, productionPattern: undefined },
```
e2corporation, line 54: Using `undefined` for String inputs is not desired practice. When you set this to `undefined` it will propagate to `$activeTransformation` which is the main `$transformation` object at this layer, and React does not expect or want an undefined value for its input components.

It appears you are wanting to use `undefined` as a means to determine the "Off" state for radio selectors, when we should be using `empty` string instead (as originally designed).

If empty string means something else to the backend, we may need to consider using another string value to "disable" a deploy tag, perhaps "[disabled]" or "/disabled/"

config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
mintsweet: This is to solve the button text display "add transformation" or "edit transformation" exists.

config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
mintsweet: Because "deploymentPattern" and "productionPattern" are empty, DORA-related tasks will be started, and DORA tasks will not be started when they do not exist. The original judgment here cannot support the situation that these two parameters are empty strings.

config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
mintsweet: This is the same as the above.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
mintsweet: **deploymentPattern** and **productionPattern** are empty strings by default, DORA-related tasks will be executed at this time, and default by the backend will convert their values. when the request does not have these two parameters, DORA-related tasks Tasks will not be executed, set to undefined in the request parameters, and they will not be brought into the request parameters.

If using an extra parameter to manage them, I think this would incur additional maintenance costs.

e2corporation: This is not the ideal solution using `undefined` values, I don't see the extra prop as being more maintenance. When using React controlled components, `undefined` cannot be assigned to input component values. This is introducing errors into the React console that will affect the production build. If empty string is not supported we'll need to add some post-processing to the board or project's transformation to exclude those properties from payload if empty. It would be easier for backend to treat empty strings as "disabled" and not execute DORA tasks.

config-ui/src/components/blueprints/create-workflow/DataTransformations.jsx
e2corporation: That behavior is working already working as desired, Edit transformation only shows if one of the props has a non-empty value. Your method of how deployment tags are being utilized is causing you to have to modify this logic.

e2corporation: While I'm a fan of `lodash` and it has many useful helpers for working with arrays and objects etc., It's not really needed to get the existing logic working, and introduces another 3rd party dependency.

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
mintsweet: I think there is some truth to what you said. Whether a controlled component can receive undefined depends on the definition of the component. The reason I avoid these is that the current component hierarchy is too deep. If you can reduce the calling level, use what you said. The way I think it would indeed be better.

abeizn (approved):

config-ui/src/components/blueprints/transformations/CICD/Deployment.jsx
e2corporation: @mintsweet Nesting level is not related to the fact that Controlled Input Components must not be defined. I am going to refactor this all later to meet those original principles. This PR has caused 2 problems:

- Creating `[object]` and `[undefined]` Entries in main **Transformations** Map due to use of entityIdKey within an Effect instead of the original handler I had implemented.
- Improper use of controlled input switching from `undefined` to a controlled value during its lifecycle.

e2corporation: Linting was also not run prior to finalizing this PR, and a missing Effect dependency has caused new warning.

Screen Shot 2022-09-28 at 9 35 17 AM

e2corporation: @mintsweet @likyh This effect will definitely need to be removed, adding the dependency here creates a circular loop resulting in a maximum nesting level error. This is why it's always extremely important to run lint locally and fix the react-hooks/exhaustive-deps warnings.

Screen Shot 2022-09-29 at 7 26 02 PM

PR #3214, opened 2022-09-27
fix: search jira account by key instead of username

### LINKED ISSUE #3184: [Bug][Jira] inconsistent account id within account_extractor and issue_changelog_extractor (jira server)
### Search before asking

### What happened

Could not find accounts whose id is in issue_changelogs in account entity, when I collecting data from jira server (7.13.0). After some debug, I find that jira api return account's key as changelogs' value id(entity). but devlake using accounts' email as account's  id(entity).

## account api
image

## changelog api (issue's extend)
image

## reference
- https://github.com/apache/incubator-devlake/blob/1c88f5cd087f7079d4f4603b8e7ef0c18895133f/plugins/jira/tasks/apiv2models/user.go#L42-L50

Supposed to  use same key so that we can associate changeloges with accounts.

### How to reproduce

Run jira plugin to collect data from jira server 7.13

### Anything else

Please cherry-pick the fix to release-v0.12 branch as we are still in it. 

### Version

v0.12.0

### Code of Conduct

mindlesscloud:
# Summary

fix #3184 
Using `key` instead of `username` as a query keyword while collecting Jira accounts.

Closes #3184 

### Screenshots

### Other Information

### DIFF
--- a/plugins/jira/tasks/account_collector.go
+++ b/plugins/jira/tasks/account_collector.go
@@ -60,8 +60,7 @@ func CollectAccounts(taskCtx core.SubTaskContext) errors.Error {
 	queryKey := "accountId"
 	urlTemplate := "api/2/user"
 	if data.JiraServerInfo.DeploymentType == models.DeploymentServer {
-		queryKey = "username"
-		urlTemplate = "api/2/user/search"
+		queryKey = "key"
 	}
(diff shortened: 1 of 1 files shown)


abeizn (approved):
LGTM

PR #3230, opened 2022-09-28
Feat plugin coding init

warren830:
# Summary

add coding plugin

Closes xx

### Screenshots
image

### Other Information


plugins/helper/api_collector.go
```diff
@@ -335,6 +337,13 @@ func (collector *ApiCollector) fetchAsync(reqData *RequestData, handler func(int
 			panic(err)
 		}
 	}
+	var reqBody interface{}
+	if collector.args.RequestBody != nil {
+		reqBody = collector.args.RequestBody(reqData)
+		if err != nil {
```
keon94, line 343: should we panic here? Why not just return the error?

PR #3247, opened 2022-09-28
fix: config-ui: prevent autofill of passwords on new connection form

### LINKED ISSUE #3241: [Bug][config-ui] Config-ui using values configured in a jira connection as default values when creating a new github connection
### Search before asking

### What happened

Config-ui using values configured in a jira connection as default values when creating a new GitHub connection. As a result, the token was invalid and I had to delete it and re-type again.

TxfMr0PAKi

Not using values configured in other data source connections as the default values when creating any new connection.

### How to reproduce

1. go to config-ui
2. create a jira connection and fill in the form
3. create a github connection, see default values

### Anything else

_No response_

### Version

main

### Code of Conduct

e2corporation:
### Config-UI / Data Connections / **New Connection**

- [x] `Fix` set `autoComplete` to `new-password` mode for  **Username** & **Password** Type fields

### Description

This PR applies an update to **Username** & **Password** Type fields on the **Connection Form** so that the `autoComplete` attribute is properly respected. This will prevent certain browsers like Firefox from forcibly entering default values for login and password fields when creating a new connection.

For additional reference on how autoComplete attribute is used, please see https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

Closes #3241

### Screenshots
Screen Shot 2022-09-28 at 2 41 17 PM

### DIFF
--- a/config-ui/src/pages/configure/connections/ConnectionForm.jsx
+++ b/config-ui/src/pages/configure/connections/ConnectionForm.jsx
@@ -516,7 +516,7 @@ export default function ConnectionForm(props) {
                                 }
                                 className='input personal-token-input'
                                 fill
-                                autoComplete='false'
+                                autoComplete='new-password'
                               />
                             </div>
                             {tokenTests[patIdx] && (
@@ -595,7 +595,7 @@ export default function ConnectionForm(props) {
                   <InputGroup
                     id='connection-token'
                     type='password'
-                    autoComplete='false'
+                    autoComplete='new-password'
                     inputRef={connectionTokenRef}
                     disabled={isTesting || isSaving || isLocked}
                     placeholder={
@@ -716,7 +716,7 @@ export default function ConnectionForm(props) {
                   id='connection-password'
                   tabIndex={1}
                   type='password'
-                  autoComplete='false'
+                  autoComplete='new-password'
                   inputRef={connectionPasswordRef}
                   disabled={isTesting || isSaving || isLocked}
                   placeholder='Enter Password'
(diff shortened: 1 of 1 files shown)


likyh:
image

'new-password' seems new password will show.  And I find autoComplete="false" used in some other codes. Let's fix all of these in one PR?

mintsweet:
@e2corporation  Hi, Julien, I don't think this modification can completely solve the problem, it will cause firefox to incorrectly fill in the information elsewhere.

I don't think we should revise this, what do you think?

e2corporation:
> image
> 
> 'new-password' seems new password will show. And I find autoComplete="false" used in some other codes. Let's fix all of these in one PR?

I think i covered all the uses of `autoComplete`, _except_ for the `GenerateTokenForm` which is a legacy component for when we offered Base64 encoding for JIRA so that component is not actively being used at the moment so I didn't change it.

e2corporation:
> @e2corporation Hi, Julien, I don't think this modification can completely solve the problem, it will cause firefox to incorrectly fill in the information elsewhere.
> 
> I don't think we should revise this, what do you think?

Can you demonstrate the breakage `new-password` is causing for you? I tested it on local Firefox it appears to be respecting the setting and not auto-filling passwords like Louis was experiencing.

As this is a Browser-induced issue I am also fine with not applying a fix if it's not necessary. In either case we still should modify the `autoComplete` property since `false` is not a desired setting.

mintsweet:
@e2corporation Hi, Julien.

When I autofill to firefox with my current code like this(username and password are filled):
screenshot-20220930-092004
screenshot-20220930-092009

When I modify the **autoComplete** of username to **new-password** (endpoint and password are filled):
screenshot-20220930-091944
screenshot-20220930-091950


This is very unexpected, I think this is a firefox problem, we don't need to write hack code to be compatible with firefox.

likyh:
This PR seems not to block user and there are some disputes. Let's edit it until it fixed. Needn't limit on this Millstone.

e2corporation:
Converting this PR to **Draft** until this issue is further inspected.

PR #3260, opened 2022-09-29
fix: fix migration dialog only appear after refresh

### LINKED ISSUE #3109: [Bug][Config-UI] database migration failed warning after successful mgration
### Search before asking

### What happened

After a successful DB Migration the UI Warning does not go away.

If Migration was successful the DB Migration Warning dialog should not remain.

### How to reproduce

Test a new DB Migration.

### Anything else

_No response_

### Version

main

### Code of Conduct

### LINKED ISSUE #3255: [Bug][config-ui] Db migration dialog doesn't popup on Github connection page
### Search before asking

### What happened

image

db migration dialog pop up

### How to reproduce

1. setup a brand new devlake deployment(new database)
2. open config-ui, try to add a github connection
3. see error

### Anything else

commit sha: bf4de4df3bda079b4cca2ae9e97e8cc9e3d46c54

### Version

main

### Code of Conduct

likyh:
# Summary

1. fix migration dialog only appear after refresh
2. fix migration cannot success

Closes #3255 
Closes #3109

### DIFF
--- a/config-ui/src/hooks/useDatabaseMigrations.jsx
+++ b/config-ui/src/hooks/useDatabaseMigrations.jsx
@@ -73,19 +73,18 @@ function useDatabaseMigrations(Configuration = MigrationOptions) {
     }
   }, [hasMigrationFailed])
 
-  const setNeedMigrateCallback = useCallback((errorObject) => {
-    if (migrationWarning) {
-      console.log(`>>> MIGRATION WARNING DETECTED !!:`, migrationWarning)
-    }
-    setMigrationWarning(errorObject)
-    setWasMigrationSuccessful(false)
-    setIsProcessing(false)
-    setHasMigrationFailed(false)
-  }, [setMigrationWarning])
+  const setNeedMigrateCallback = useCallback(
+    (errorObject) => {
+      setMigrationWarning(errorObject)
+      setWasMigrationSuccessful(false)
+      setIsProcessing(false)
+      setHasMigrationFailed(false)
+    },
+    [setMigrationWarning]
+  )
   useEffect(() => {
     MigrationOptions.setNeedMigrateCallback(setNeedMigrateCallback)
   }, [setNeedMigrateCallback])
-
 
   return {
     migrationWarning,
--- a/config-ui/src/hooks/useDatabaseMigrations.jsx
+++ b/config-ui/src/hooks/useDatabaseMigrations.jsx
@@ -33,9 +33,11 @@ function useDatabaseMigrations(Configuration = MigrationOptions) {
     setIsProcessing(true)
     const migrate = async () => {
       const m = await request.get(Configuration.apiProceedEndpoint)
-      setWasMigrationSuccessful(m?.status === 200 && m?.data?.success === true)
-      setIsProcessing(false)
-      setHasMigrationFailed(m?.status !== 200)
+      setTimeout(() => {
+        setWasMigrationSuccessful(m?.status === 200 && m?.data?.success === true)
+        setIsProcessing(false)
+        setHasMigrationFailed(m?.status !== 200)
+      }, 3000)
     }
     migrate()
   }, [Configuration.apiProceedEndpoint])
--- a/config-ui/src/hooks/useDatabaseMigrations.jsx
+++ b/config-ui/src/hooks/useDatabaseMigrations.jsx
@@ -24,9 +24,7 @@ import { ToastNotification } from '@/components/Toast'
 function useDatabaseMigrations(Configuration = MigrationOptions) {
   const [isProcessing, setIsProcessing] = useState(false)
 
-  const [migrationWarning, setMigrationWarning] = useState(
-    localStorage.getItem(Configuration.warningId)
-  )
+  const [migrationWarning, setMigrationWarning] = useState()
   const [migrationAlertOpened, setMigrationAlertOpened] = useState(false)
   const [wasMigrationSuccessful, setWasMigrationSuccessful] = useState(false)
   const [hasMigrationFailed, setHasMigrationFailed] = useState(false)
@@ -36,17 +34,14 @@ function useDatabaseMigrations(Configuration = MigrationOptions) {
     const migrate = async () => {
       const m = await request.get(Configuration.apiProceedEndpoint)
       setWasMigrationSuccessful(m?.status === 200 && m?.data?.success === true)
-      setTimeout(() => {
-        setIsProcessing(false)
-        setHasMigrationFailed(m?.status !== 200)
-      }, 3000)
+      setIsProcessing(false)
+      setHasMigrationFailed(m?.status !== 200)
     }
     migrate()
   }, [Configuration.apiProceedEndpoint])
 
   const handleCancelMigration = useCallback(() => {
     setIsProcessing(true)
-    localStorage.removeItem(Configuration.warningId)
     setMigrationAlertOpened(false)
     setIsProcessing(false)
     ToastNotification.clear()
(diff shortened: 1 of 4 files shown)


config-ui/src/hooks/useDatabaseMigrations.jsx
```diff
@@ -36,17 +34,14 @@ function useDatabaseMigrations(Configuration = MigrationOptions) {
     const migrate = async () => {
       const m = await request.get(Configuration.apiProceedEndpoint)
       setWasMigrationSuccessful(m?.status === 200 && m?.data?.success === true)
-      setTimeout(() => {
```
e2corporation, line 39: This timeout is for latency and shouldn't be removed, by removing it the user will not see the migration message instead it will just "flash" and be a confusing state. This gives the user at least a 3-second window, it does not block migration from actually happening.

likyh: setTimeout must write without async and await...

e2corporation:
This ESLint warning should be resolved as well (missing dependency)
```
/home/runner/work/incubator-devlake/incubator-devlake/config-ui/src/hooks/useDatabaseMigrations.jsx
[16](https://github.com/apache/incubator-devlake/actions/runs/3151698589/jobs/5125983010#step:5:17)
  86:5  warning  React Hook useCallback has a missing dependency: 'migrationWarning'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

config-ui/src/hooks/useDatabaseMigrations.jsx
e2corporation: Line 35 is synchronous  as await is being used `const m = await request.get(Configuration.apiProceedEndpoint)`.

e2corporation: The timeout gives the UI at least a 3second window for migration to complete. If migration takes longer than 3 seconds then the delta is still [MIGRATION_TIME] + [3SEC TIMEOUT]

e2corporation: You can clearly see the effect on the UI if you remove this timeout, the user can't read the migration message.

likyh:
> @likyh The reason for a refresh being needed is so the local storage key is picked up. If the user navigates away or reloads the page the warning will happen, for a more instant warning we could just try to use a `history.push` to send the user to /dashboard or some alt route so the migration warning local storage key is detected.

But now it can not appear when first open. So I use a new way to get if migration needed.

Now it show appear when first request sent. (Usually is `/versions`)

config-ui/src/hooks/useDatabaseMigrations.jsx
likyh: Let's write a common helper to sovle this problem. Because if add setTimeout out of async migrate, I must write some logic to deal which one finish first between setTimeout and `request.get`.

likyh:
> I like your proposed refactored solution, however I think if we tried to add a history push and re-route the user to another location when we receive the migration warning my original design of using local storage key will work as expected.

You means add history push in https://github.com/apache/incubator-devlake/blob/main/config-ui/src/utils/request.js#L43 ?
https://github.com/apache/incubator-devlake/c732e7bddeb1f498873c6cc807e66772deb205e9/main/config-ui/src/utils/request.js#L43

config-ui/src/hooks/useDatabaseMigrations.jsx
e2corporation: The timeout is not doing much... we wait 3 seconds then 1) Turn off the Processing flag, 2) Set the Migration Failed flag based on the response. `m` is awaited, so it should be blocking. If the request to migrate takes 3 or 30 seconds, the timeout should execute _after_ that time not before.

e2corporation:
> > > @likyh The reason for a refresh being needed is so the local storage key is picked up. If the user navigates away or reloads the page the warning will happen, for a more instant warning we could just try to use a `history.push` to send the user to /dashboard or some alt route so the migration warning local storage key is detected.
> > 
> > 
> > But now it can not appear when first open. So I use a new way to get if migration needed.
> > Now it show appear when first request sent. (Usually is `/versions`)
> 
> I like your proposed refactored solution, however I think if we tried to add a history push and re-route the user to another location when we receive the migration warning my original design of using local storage key will work as expected.



> > I like your proposed refactored solution, however I think if we tried to add a history push and re-route the user to another location when we receive the migration warning my original design of using local storage key will work as expected.
> 
> Do you mean that add history push in `https://github.com/apache/incubator-devlake/blob/main/config-ui/src/utils/request.js#L43` ?
> 
> https://github.com/apache/incubator-devlake/blob/c732e7bddeb1f498873c6cc807e66772deb205e9/config-ui/src/utils/request.js#L40-L50

We can't use a Hook call at this layer as it breaks core hook rules regarding usage, (Additionally, I left this utility Pure and it is Not a React component) what we need to do is find out where we can check the storage key first and send the appropriate signal.

config-ui/src/hooks/useDatabaseMigrations.jsx
e2corporation: If you restore this timeout does your new logic break?

e2corporation:
@likyh If this new solution is working for now, I'll approve and we can try to fix the `localStorage` method later. It would be ideal to restore the timeout and fix the eslint dependency warning however.

e2corporation (approved):
LGTM (Please resolve ESLint dependency warning with `useDatabaseMigrations.jsx`)

config-ui/src/hooks/useDatabaseMigrations.jsx
likyh: old logic can not run actually. setTimeout must write without async and await... Now setTimeout only start when migrations success and it will wait like this:
image

Do you means you want show loading time = 3s + actually time?

e2corporation: That screenshot is after you commented/disabled the timeout or with the timeout enabled?

likyh: > That screenshot is after you commented/disabled the timeout or with the timeout enabled?

It's fix the typo `wasSuccessful={wasMigrationSuccessful}` on old code.

likyh: you want show loading time = 3s + actuallyTime or time = Max(3s, actuallyTime) or time = Min(3s, actuallyTime)?

e2corporation: The intent here is the dialog should display for at least 3 seconds so user can see the migration happening. If migration completes in 0.05 seconds, we still want the 3 second window so user can read the message. If DB Migration request is still processing and takes 30 seconds to complete, the dialog should remain open in the migrating state for 30SECONDS + 3SECONDS so 33seconds total

likyh: So loading time = 3s + actuallyTime?

e2corporation: If the timeout is truly causing a break in the flow we can comment/disable it for now. But my understanding of the await call is that line should be blocking. So that initial request whether it takes 1 second or 30 seconds to receive a response from backend, it should still be blocking until time timeout fires

likyh: Just tell me if loading time = 3s + actuallyTime ?  Other information I have gotten.

e2corporation: yes 3s + actual time

e2corporation:
@likyh Also remember to "npm run lint" locally to fix formatting and indentation etc..

config-ui/src/hooks/useDatabaseMigrations.jsx
likyh: ok

e2corporation (approved):
LGTM

PR #3287, opened 2022-09-30
fix(config-ui): change jira board selection box to be searchable

### LINKED ISSUE #3254: [Question][Blueprints] Can't get all Jira Boards
## Question
When setting the jira data scope，I can only get 50 jira boards to choose, actually we have more than 50 boards.How can I get the others?

mintsweet:
# Summary

Change the Jira board selection box to be searchable, and avoid the interface only returning 50 boards limit.

Closes #3254

### Screenshots
screenshot-20220930-120743

### Other Information
Nothing.

### DIFF
--- a/config-ui/src/components/blueprints/DataScopesGrid.jsx
+++ b/config-ui/src/components/blueprints/DataScopesGrid.jsx
@@ -171,8 +171,8 @@ const DataScopesGrid = (props) => {
                     padding: 0
                   }}
                 >
-                  {c.boards.map((board, bIdx) => (
-                    <li key={`list-item-key-${bIdx}`}>{board}</li>
+                  {c.boardsList.map((board, bIdx) => (
+                    <li key={`list-item-key-${bIdx}`}>{board.title}</li>
                   ))}
                 </ul>
               )}
--- a/config-ui/src/hooks/useDataScopesManager.jsx
+++ b/config-ui/src/hooks/useDataScopesManager.jsx
@@ -458,14 +458,15 @@ function useDataScopesManager({
     (boardIds = [], boardListItems = []) => {
       return boardIds.map((bId, sIdx) => {
         const boardObject = boardListItems.find(
-          (apiBoard) => Number(apiBoard.id) === Number(bId)
+          (apiBoard) =>
+            Number(apiBoard.id) === Number(bId) || +apiBoard.boardId === +bId
         )
         return new JiraBoard({
           ...boardObject,
           id: boardObject?.id || bId || sIdx + 1,
           key: sIdx,
           value: bId,
-          title: boardObject?.name || `Board ${bId}`,
+          title: boardObject?.name || boardObject?.title || `Board ${bId}`,
           type: boardObject?.type || 'scrum',
           location: { ...boardObject?.location }
         })
--- a/config-ui/src/pages/blueprints/blueprint-settings.jsx
+++ b/config-ui/src/pages/blueprints/blueprint-settings.jsx
@@ -117,6 +117,8 @@ const BlueprintSettings = (props) => {
   const [runTasks, setRunTasks] = useState([])
   const [runTasksAdvanced, setRunTasksAdvanced] = useState([])
 
+  const [boardSearch, setBoardSearch] = useState()
+
   const {
     // eslint-disable-next-line no-unused-vars
     activeStep,
@@ -671,8 +673,11 @@ const BlueprintSettings = (props) => {
                 connectionsList,
                 [Providers.JIRA].includes(c.plugin)
                   ? getJiraMappedBoards(
-                      c.scope.map((s) => s.options?.boardId),
-                      allJiraResources?.boards
+                      c.scope?.map((s) => s.options?.boardId),
+                      [
+                        ...allJiraResources?.boards,
+                        ...c.scope.map((s) => s.options)
+                      ]
                     )
                   : []
               )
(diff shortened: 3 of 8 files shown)


config-ui/src/hooks/useDataScopesManager.jsx
```diff
@@ -458,14 +458,15 @@ function useDataScopesManager({
     (boardIds = [], boardListItems = []) => {
       return boardIds.map((bId, sIdx) => {
         const boardObject = boardListItems.find(
-          (apiBoard) => Number(apiBoard.id) === Number(bId)
+          (apiBoard) =>
+            Number(apiBoard.id) === Number(bId) || +apiBoard.boardId === +bId
```
likyh, line 462: what difference between these 2 conditions?

config-ui/src/pages/blueprints/blueprint-settings.jsx
```diff
@@ -672,7 +674,10 @@ const BlueprintSettings = (props) => {
                 [Providers.JIRA].includes(c.plugin)
                   ? getJiraMappedBoards(
                       c.scope.map((s) => s.options?.boardId),
-                      allJiraResources?.boards
+                      [
+                        ...allJiraResources?.boards,
+                        ...c.scope.map((s) => s.options)
```
likyh, line 679: this change for what?

config-ui/src/pages/blueprints/blueprint-settings.jsx
```diff
@@ -897,35 +905,64 @@ const BlueprintSettings = (props) => {
...
-                  boardsList: jiraResourcesResponse?.boards
-                    ? getJiraMappedBoards(
-                        c.boardIds,
-                        jiraResourcesResponse?.boards
-                      )
-                    : []
-                })
-            )
+      fetchIssueTypes()
+      fetchFields()
```
likyh, line 909: why fetch these two when fetch boards?

likyh (commented):
some other bugs:
1. when open modal agian, the board name lose
2. (main also exist) When I delete a board or adjust the order of the boards, the transformation of the second page will not match. For example, the original requirement of board1 is tagA, and the requirement of board2 is tagB. After 1 is deleted, board2 becomes tagA

likyh:
Maybe we can change this tip "Select the boards you would like to sync."

likyh:
I find a bug again:

3. type mapping is wrong when edit board
image

likyh (approved):
I test it and cannot find new bugs compared with the main branch. So Approved.

config-ui/src/components/blueprints/DataScopesGrid.jsx
```diff
@@ -171,8 +171,8 @@ const DataScopesGrid = (props) => {
                     padding: 0
                   }}
                 >
-                  {c.boards.map((board, bIdx) => (
```
e2corporation, line 174: @mintsweet  @likyh This is a bug, this should not have been changed. `boards` Represents the Boards Associated with the Blueprints. `boardsList` represents ALL Boards

likyh: It looks like boards is string array and boardsList is object list. not boardsList represents ALL Boards but boards represents a part of.

e2corporation:
> some other bugs:
> 
> 1. when open modal agian, the board name lose
@likyh This was an advanced timing issue that's existed for sometime. JIRA API data is not ready on the first modification request so the active board names are not applied, on the second modification request the proxy data is ready so true board name are displayed. Our Design requirements have only ever identified Boards using "Board [ID]", our blueprint only stores board ID as well. So this is why I've left this behavior intact until the timing issue is resolved. The Important thing is that Board ID is still preserved, regardless if actual name or generic board name is displayed.

> 2. (main also exist) When I delete a board or adjust the order of the boards, the transformation of the second page will not match. For example, the original requirement of board1 is tagA, and the requirement of board2 is tagB. After 1 is deleted, board2 becomes tagA
This was recently fixed by my last PR

config-ui/src/components/blueprints/DataScopesGrid.jsx
e2corporation: `boards` is the retrieved list of stored boards on the blueprint, designed for Display on the Data Scopes Grid.

likyh:
> > some other bugs:
> > 
> > 1. when open modal agian, the board name lose
> 
> @likyh This was an advanced timing issue that's existed for sometime. JIRA API data is not ready on the first modification request so the active board names are not applied, on the second modification request the proxy data is ready so true board name are displayed. Our Design requirements have only ever identified Boards using "Board [ID]", our blueprint only stores board ID as well. So this is why I've left this behavior intact until the timing issue is resolved. The Important thing is that Board ID is still preserved, regardless if actual name or generic board name is displayed.
> 
> > 2. (main also exist) When I delete a board or adjust the order of the boards, the transformation of the second page will not match. For example, the original requirement of board1 is tagA, and the requirement of board2 is tagB. After 1 is deleted, board2 becomes tagA
> 
> This was recently fixed by my last PR for "Blueprint Settings Degraded..."

1 is after the bug fix you said and it's a new bug on this PR and it fixed.

2 is https://github.com/apache/incubator-devlake/issues/3294

config-ui/src/components/blueprints/DataScopesGrid.jsx
e2corporation: This needs to be reverted to `c.boards.map((board, bIdx)`

e2corporation:
Where does the board limit of 50 come from in the first place?

Issue #3294, opened 2022-09-30
[Bug][config-ui] board transfomation match wrong when edit jira blueprint


likyh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

When I delete a board or adjust the order of the boards, the transformation of the second page will not match. For example, the original requirement of board1 is tagA, and the requirement of board2 is tagB. After 1 is deleted, board2 becomes tagA

### What you expected to happen

After 1 is deleted, board2 keep tagB

### How to reproduce

![image](https://user-images.githubusercontent.com/3294100/193201534-1631b9a2-add6-4a2b-b70b-ea7de46482b2.png)
![image](https://user-images.githubusercontent.com/3294100/193201503-dd4f866b-9b94-4045-bb6c-4635752cc232.png)
![image](https://user-images.githubusercontent.com/3294100/193201636-c4516c3b-07bb-4a10-a919-87d505898412.png)
----
![image](https://user-images.githubusercontent.com/3294100/193201757-e32f0b2f-02d5-46bd-b973-1f0d9a9c9668.png)
![image](https://user-images.githubusercontent.com/3294100/193202353-977724fe-5537-4364-a40f-bd8c041945d1.png)

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

e2corporation:
@likyh This was already resolved with #3275

likyh:
> @likyh This was already resolved with #3275

No, I test on latest main. Please do a test yourself.

PR #676, opened 2021-11-08
#672 Fix: JIRA Settings - Use New Format for Type Mappings

e2corporation:
### Key Points

- [x] Send `typeMappings` request payload as Object Structure to the Lake API for the 3 main types, **Requirement**, **Incident** and **Bug**
  - Do not include `statusMappings` key (deprecated)
- [x] Comment out legacy format (string-delimited format)
- [x] Types can be cleared & set to **blank**

### Description
This MR resolves outstanding issues on the **JIRA Settings Configuration** page that prevent **Type Mappings** from being correctly stored.

```
# SAMPLE REQUEST (JIRA Settings)
"typeMappings": {
            "Story": {
                "standardType": "Requirement",
            },
            "Incident": {
                "standardType": "Incident",
            },
            "Bug": {
                "standardType": "Bug",
            }
        }
```

### Does this close any open issues?
#672 

### Current Behavior
Type Mappings cannot be saved or re-loaded for a JIRA connection instance.

### New Behavior
Type Mappings are now correctly saved and re-parsed for edit/modify mode when a connection instance is loaded.

### Screenshots
Screen Shot 2021-11-08 at 1 16 33 PM
Screen Shot 2021-11-08 at 1 15 36 PM

PR #3297, opened 2022-09-30
fix(config-ui): transformation not updating correctly when switching boards

### LINKED ISSUE #3295: [Bug][config-ui] add board but type mapping is wrong when edit jira blueprint
### Search before asking

### What happened

image

new boards have correct type mapping. It should equal user select.

### How to reproduce

1. find an existing blueprint with jira.
2. add a board in jira
3. select an issue type for bugs
4. save
5. look at the body of request

### Anything else

_No response_

### Version

main

### Code of Conduct

mintsweet:
# Summary

transformation not updating correctly when switching boards.

Closes #3295

### Screenshots
Nothing.

### Other Information
Nothing.

### DIFF
--- a/config-ui/src/pages/configure/settings/jira.jsx
+++ b/config-ui/src/pages/configure/settings/jira.jsx
@@ -177,58 +177,35 @@ export default function JiraSettings(props) {
   }, [typeMappingAll, onSettingsChange, configuredBoard?.id])
 
   useEffect(() => {
-    if (typeMappingBug && typeMappingIncident && typeMappingRequirement) {
-      const RequirementMappings =
-        typeMappingRequirement !== ''
-          ? typeMappingRequirement.map((r) =>
-              createTypeMapObject(r.value, MAPPING_TYPES.Requirement)
-            )
-          : []
-      const IncidentMappings =
-        typeMappingIncident !== ''
-          ? typeMappingIncident.map((i) =>
-              createTypeMapObject(i.value, MAPPING_TYPES.Incident)
-            )
-          : []
-      const BugMappings =
-        typeMappingBug !== ''
-          ? typeMappingBug.map((b) =>
-              createTypeMapObject(b.value, MAPPING_TYPES.Bug)
-            )
-          : []
-      const CombinedMappings = [
-        ...RequirementMappings,
-        ...IncidentMappings,
-        ...BugMappings
-      ].filter((m) => m !== null)
-      const MappingTypeObjects = CombinedMappings.reduce((pV, cV) => {
-        return { ...cV, ...pV }
-      }, {})
-      setTypeMappingAll(MappingTypeObjects)
-      console.log(
-        '>> INCIDENT TYPE MAPPING OBJECTS....',
-        RequirementMappings,
-        IncidentMappings,
-        BugMappings
-      )
-      console.log('>> ALL MAPPINGS COMBINED...', CombinedMappings)
-      console.log(
-        '>> FINAL MAPPING OBJECTS FOR API REQUEST...',
-        MappingTypeObjects
-      )
-    }
-  }, [typeMappingBug, typeMappingIncident, typeMappingRequirement])
+    setTypeMappingAll((ma) => ({
+      ...ma,
+      ...(typeMappingBug || [])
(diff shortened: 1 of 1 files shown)


config-ui/src/pages/configure/settings/jira.jsx
```diff
@@ -177,58 +177,35 @@ export default function JiraSettings(props) {
...
-    console.log('>> CONN SETTINGS OBJECT ', connection)
-    if (connection && connection.id) {
-      // Parse Type Mappings (V2)
-      // setStatusMappings([])
-    }
-  }, [connection])
+    setTypeMappingAll((ma) => ({
+      ...ma,
+      ...(typeMappingIncident || [])
+        .map((r) => createTypeMapObject(r.value, MAPPING_TYPES.Requirement))
```
e2corporation, line 192: Is this supposed to be `MAPPING_TYPES.Incident` ?

config-ui/src/pages/configure/settings/jira.jsx
```diff
@@ -177,58 +177,35 @@ export default function JiraSettings(props) {
...
-      console.log(
-        '>> FINAL MAPPING OBJECTS FOR API REQUEST...',
-        MappingTypeObjects
-      )
-    }
-  }, [typeMappingBug, typeMappingIncident, typeMappingRequirement])
+    setTypeMappingAll((ma) => ({
+      ...ma,
+      ...(typeMappingBug || [])
+        .map((r) => createTypeMapObject(r.value, MAPPING_TYPES.Requirement))
```
e2corporation, line 183: Is this supposed to be MAPPING_TYPES.Bug ?

config-ui/src/pages/configure/settings/jira.jsx
```diff
@@ -177,58 +177,35 @@ export default function JiraSettings(props) {
...
+  }, [typeMappingIncident])
+
+  useEffect(() => {
+    setTypeMappingAll((ma) => ({
+      ...ma,
+      ...(typeMappingRequirement || [])
+        .map((r) => createTypeMapObject(r.value, MAPPING_TYPES.Requirement))
+        .reduce((c, p) => ({ ...c, ...p }), {})
+    }))
+  }, [typeMappingRequirement])
```
likyh, line 204: why not merge to one `useEffect`?     Will it add more than one duplicate for typeMappingXXX in typeMappingAll

mintsweet: Independent dependencies allow **react** to distinguish which needs to be recalculated. If merged, individual changes will cause the whole to be recalculated, wasting performance.

e2corporation: Effects run asynchronously, there's a reason why the existing logic was using 1 Effect in the first place, `typeMappingAll` may encounter collisions as a result. The 3 separate requirement groups should all be processed at the same time, not separately.

e2corporation: Type Mappings have been running stable inside **One** Effect since this PR https://github.com/apache/incubator-devlake/pull/676

mintsweet: The reason why this bug occurs is because when the board is switched, the three dependencies have not changed.

mintsweet: Of course, merging them together can work, but sometimes they do not change at the same time, and the other two variables will also change, which is not necessary. Splitting them can monitor their own related content changes so that changes among them Previously do not cause additional content to be recalculated.

e2corporation: @mintsweet In the past I recalled using 3 separate Effects only to find out that it caused a sync-issue on the UI when selectors were used rapidly, resulting in `typeMappingAll` being overwritten. I am willing to try 3 separate Effects again, assuming it directly addresses the bug reported on issue 3295.

likyh: It resolves #3295 but #3294 exists.  I think we can merge it and I'll try to refactor here to delete `configuredBoard` in the detail page and resolve #3294

likyh (approved):
It looks approved if it works. And I want to change it in #3371. So merged.

PR #3298, opened 2022-09-30
fix: fix 3277

### LINKED ISSUE #2400: [Bug][jira] connection test not work with username
### Search before asking

### What happened

image
image

As we see.
We use an wrong username with the right token password. it is test right.
But when we use it to collect data. It told us that authentication failed.

Maybe we would better to use another api for test like /user/search ? 
image
image

Or at least let user to check his Username but not his AccessToken.
image

### How to reproduce

1. set an right token.
2. set an wrong username.
3. save connection
4. run pipeline.

### Anything else

_No response_

### Version

main

### Code of Conduct

### LINKED ISSUE #3277: [Bug][Jira] test connection api return success with a pair of any username and a valid token
### Search before asking

### What happened

I am using test connection api to check jira integration info before synchronizing.It always return success if the token is valid, even if the username is not exists.  
But the jira tasks will fail with the wrong username.

Test connection api should keep consistent behaviour with sync tasks.  
I mean, if test connection api is passed, sync tasks should not failed for authentication reason.

### How to reproduce

Call test connection api with right endpoint, password but wrong username. 

### Anything else

This issue occurs both in jira server and jira cloud version.

### Version

v0.12.0

### Code of Conduct

mappjzc:
# Summary
connection test api/2/status to agile/1.0/board

Closes #3277 
Closes #2400 

### Screenshots
image
image
image
image

### Other Information

### DIFF
--- a/plugins/jira/api/connection.go
+++ b/plugins/jira/api/connection.go
@@ -20,11 +20,12 @@ package api
 import (
 	"context"
 	"fmt"
-	"github.com/apache/incubator-devlake/errors"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
+
+	"github.com/apache/incubator-devlake/errors"
 
 	"github.com/apache/incubator-devlake/plugins/core"
 	"github.com/apache/incubator-devlake/plugins/helper"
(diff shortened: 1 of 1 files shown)


mindlesscloud (approved):
LGTM

PR #3303, opened 2022-09-30
feat: Added Responsive Sidebar

### LINKED ISSUE #3106: [Config-UI] Responsive Sidebar for Better View on Smaller Windows
### Search before asking

## Takes Unnecessary Space on smaller screens
- While Setting up Lake, placing the windows side by side, you can see that the sidebar takes up a lots of unnecessary space and the actual `route pages` are kind of hidden.

devlakeSidebarIssue

## Mobile View is Mostly hidden by the sidebar 
- When Comes to small mobile view screens, you can see that the sidebar completely hides the viewport, in that case the user if he want to make a quick change or something with his phone, it would be impossible for him to do that.

image

### Use case

I want to implement a responsive sidebar feature for DevLake, mostly for the above two use cases
- **When the user is setting up Lake**
   A customer might need that he has documentation in one half of the window and Lake in the another half, which gives a very smooth setup experience with a responsive sidebar.
   
- **When the user is using smaller screen size like Tablets**
  According to what I think, there might not be a huge number of customers using mobiles for Lake, but there can a plenty of customers using Tablets, it would be a good option to add responsiveness to the sidebar so that those users can take full advantage the main view.
 
   

### Related issues

_No response_

### Code of Conduct

henit-chobisa:
# Summary

## Additions : 
- Added `UIContext` for Universal State Management of the Sidebar.
- Added Conditional Rendering to the Sidebar.
- Added new `Navbar Item`, `menu` button, that toggles the sidebar on smaller window sizes.

## Effect : 
- UI Context Added for further UI State Management.
- Responsive Sidebar would be implemented in the Config-UI.
- There would be a toggle button for seen in the smaller screen sizes.

Closes #3106 

### Screenshots
devlakeSidebarSolution
devlakeSidebarSolution2

### Other Information

### DIFF
--- a/config-ui/src/components/Nav.jsx
+++ b/config-ui/src/components/Nav.jsx
@@ -15,73 +15,104 @@
  * limitations under the License.
  *
  */
-import React from 'react'
-import { Alignment, Position, Popover, Navbar, Icon } from '@blueprintjs/core'
+import React, { useContext, useEffect, useState } from 'react'
+import { Position, Popover, Navbar, Icon } from '@blueprintjs/core'
 import '@/styles/nav.scss'
 import { ReactComponent as SlackIcon } from '@/images/slack-mark-monochrome-black.svg'
 import { ReactComponent as SlackLogo } from '@/images/slack-rgb.svg'
+import UIContext from '@/store/UIContext'
+import useWindowSize from '@/hooks/useWIndowSize'
 
 const Nav = () => {
+  const uiContext = useContext(UIContext)
+  const [menuClass, setMenuClass] = useState('navbarMenuButton')
+  const size = useWindowSize()
+
+  const toggleSidebarOpen = (open) => {
+    uiContext.changeSidebarVisibility(open)
+    setMenuClass((_) =>
+      open ? 'navbarMenuButtonSidebarOpened' : 'navbarMenuButton'
+    )
+  }
+
+  useEffect(
+    () =>
+      uiContext.desktopBreakpointWidth <= size.width
+        ? !uiContext.sidebarVisible && toggleSidebarOpen(true)
+        : toggleSidebarOpen(false),
+    [size]
+  )
+
   return (
     <Navbar className='navbar'>
-      <Navbar.Group align={Alignment.RIGHT}>
-        <a
-          href='https://github.com/apache/incubator-devlake'
-          rel='noreferrer'
-          target='_blank'
-          className='navIconLink'
-        >
-          <Icon icon='git-branch' size={16} />
-        </a>
-        <Navbar.Divider />
-        <a
-          href='mailto:hello@merico.dev'
-          rel='noreferrer'
-          target='_blank'
-          className='navIconLink'
-        >
-          <Icon icon='envelope' size={16} />
-        </a>
-        <Navbar.Divider />
-        {/* DISCORD: !DISABLED! */}
-        {/* <a href='https://discord.com/invite/83rDG6ydVZ' rel='noreferrer' target='_blank' className='navIconLink'>
+      <Navbar.Group className={menuClass}>
+        <Icon
+          icon={uiContext.sidebarVisible ? 'menu-closed' : 'menu-open'}
+          onClick={(e) => toggleSidebarOpen(!uiContext.sidebarVisible)}
+          size={16}
+        />
+      </Navbar.Group>
+      {!uiContext.sidebarVisible && (
+        <Navbar.Group className='navbarItems'>
+          <a
+            href='https://github.com/apache/incubator-devlake'
+            rel='noreferrer'
+            target='_blank'
+            className='navIconLink'
+          >
+            <Icon icon='git-branch' size={16} />
+          </a>
+          <Navbar.Divider />
+          <a
+            href='mailto:hello@merico.dev'
+            rel='noreferrer'
+            target='_blank'
+            className='navIconLink'
+          >
+            <Icon icon='envelope' size={16} />
+          </a>
+          <Navbar.Divider />
+          {/* DISCORD: !DISABLED! */}
+          {/* <a href='https://discord.com/invite/83rDG6ydVZ' rel='noreferrer' target='_blank' className='navIconLink'>
           <DiscordIcon className='discordIcon' width={16} height={16} />
         </a> */}
-        {/* SLACK: ENABLED (Primary) */}
-        <Popover position={Position.LEFT}>
-          <SlackIcon
-            className='slackIcon'
-            width={16}
-            height={16}
-            style={{ cursor: 'pointer' }}
-          />
-          <>
-            <div
-              style={{ maxWidth: '200px', padding: '10px', fontSize: '11px' }}
-            >
-              <SlackLogo
-                width={131}
-                height={49}
-                style={{ display: 'block', margin: '0 auto' }}
-              />
-              <p style={{ textAlign: 'center' }}>
-                Want to interact with the <strong>Merico Community</strong>?
-                Join us on our Slack Channel.
-                <br />
-                <a
-                  href='https://join.slack.com/t/devlake-io/shared_invite/zt-17b6vuvps-x98pqseoUagM7EAmKC82xQ'
-                  rel='noreferrer'
-                  target='_blank'
-                  className='bp3-button bp3-intent-warning bp3-elevation-1 bp3-small'
-                  style={{ marginTop: '10px' }}
-                >
-                  Message us on&nbsp;<strong>Slack</strong>
-                </a>
-              </p>
-            </div>
-          </>
-        </Popover>
-      </Navbar.Group>
+          {/* SLACK: ENABLED (Primary) */}
+          <Popover position={Position.LEFT}>
+            <SlackIcon
+              className='slackIcon'
+              width={16}
+              height={16}
+              style={{ cursor: 'pointer' }}
+            />
+            <>
+              <div
+                style={{ maxWidth: '200px', padding: '10px', fontSize: '11px' }}
+              >
+                <SlackLogo
+                  width={131}
+                  height={49}
+                  style={{ display: 'block', margin: '0 auto' }}
+                />
+                <p style={{ textAlign: 'center' }}>
+                  Want to interact with the <strong>Merico Community</strong>?
+                  Join us on our Slack Channel.
+                  <br />
+                  <a
+                    href='https://join.slack.com/t/devlake-io/shared_invite/zt-17b6vuvps-x98pqseoUagM7EAmKC82xQ'
+                    rel='noreferrer'
+                    target='_blank'
+                    className='bp3-button bp3-intent-warning bp3-elevation-1 bp3-small'
+                    style={{ marginTop: '10px' }}
+                  >
+                    Message us on&nbsp
+                    <strong>Slack</strong>
+                  </a>
+                </p>
+              </div>
+            </>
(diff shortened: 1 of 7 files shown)


henit-chobisa:
@e2corporation 
Julien, following you, I made a new PR, taking care of all the linter issues, commits and coding standards of Lake.
Kindly Review.
Thank you... 😄

config-ui/src/components/Nav.jsx
```diff
@@ -15,73 +15,104 @@
...
+import useWindowSize from '@/hooks/useWIndowSize'
 
 const Nav = () => {
+  const uiContext = useContext(UIContext)
+  const [menuClass, setMenuClass] = useState('navbarMenuButton')
+  const size = useWindowSize()
+
+  const toggleSidebarOpen = (open) => {
+    uiContext.changeSidebarVisibility(open)
+    setMenuClass((_) =>
```
e2corporation, line 33: There is an underscore `_` here being used to indicate the current state of `menuClass`. `mC` maybe a better alias, but since we're not referencing it we could just drop the underscore altogether.

```suggestion
    setMenuClass(() =>
```

e2corporation:
> @e2corporation
> Julien, following you, I made a new PR, taking care of all the linter issues, commits and coding standards of Lake.
> Kindly Review.
> Thank you... 😄

Thanks for getting the PR refreshed Henit! Nice work.

e2corporation:
@henit-chobisa It appears the last commit contains a space before the colon, so it's breaking the **lint-commit-message** CI Check. Are you able to use amend functionality and reword it? https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message

Screen Shot 2022-09-30 at 12 49 09 PM

e2corporation:
@henit-chobisa Also it's important to resolve any React `react-hooks/exhaustive-deps` dependency warnings. Please resolve _only_ the warning(s) associated with your files, leave any other warning regarding `DataTransformations`.

It looks like one of the effects in `Nav.jsx` needs the dependencies added.

Screen Shot 2022-09-30 at 12 54 51 PM

henit-chobisa:
> @henit-chobisa Also it's important to resolve any React `react-hooks/exhaustive-deps` dependency warnings. Please resolve _only_ the warning(s) associated with your files, leave any other warning regarding `DataTransformations`.
> 
> It looks like one of the effects in `Nav.jsx` needs the dependencies added.
> 
> Screen Shot 2022-09-30 at 12 54 51 PM

Actually Julien, the suggested dependencies cannot be added to the array, as it will show adverse side effects, I have tried it before pushing.

e2corporation:
> > @henit-chobisa Also it's important to resolve any React `react-hooks/exhaustive-deps` dependency warnings. Please resolve _only_ the warning(s) associated with your files, leave any other warning regarding `DataTransformations`.
> > It looks like one of the effects in `Nav.jsx` needs the dependencies added.
> > Screen Shot 2022-09-30 at 12 54 51 PM
> 
> Actually Julien, the suggested dependencies cannot be added to the array, as it will show adverse side effects, I have tried it before pushing.

@henit-chobisa This is actually an indicator that the logic inside the Effect is flawed in some way causing a circular loop, and the logic flow should be altered and reworked.

e2corporation:
@henit-chobisa Out of the 3 missing dependencies in the array, add them 1 at a time so you can isolate which _exact_ dependency is the culprit responsible for creating the circular loop situation then start from there.

config-ui/src/components/Nav.jsx
```diff
@@ -15,73 +15,104 @@
...
 import { ReactComponent as SlackLogo } from '@/images/slack-rgb.svg'
+import UIContext from '@/store/UIContext'
+import useWindowSize from '@/hooks/useWIndowSize'
 
 const Nav = () => {
+  const uiContext = useContext(UIContext)
+  const [menuClass, setMenuClass] = useState('navbarMenuButton')
+  const size = useWindowSize()
+
+  const toggleSidebarOpen = (open) => {
```
e2corporation, line 31: @henit-chobisa We may need to wrap this whole `toggleSidebarOpen` function with a `useCallback` hook, this function being a loose constant maybe partially responsible for the circular loop as it will cause a re-render.

henit-chobisa: While I was trying to resolve this, I did the same but after this, the `useCallback` hook, ask for the UIContext dependency and that is the infinite loop.
image

henit-chobisa: the below `useEffect` logic I figured out by just removing the `uiContext.sidebarVisible` check, that was unnecessary.

e2corporation: @henit-chobisa For the toggleSidebar callback, which dependencies are being requested?

henit-chobisa: Should I break `toggleSidebar` to two different functions ?
This may solve the issue

e2corporation: @henit-chobisa For that last main Effect, instead of providing the entire `size` object, be more specific since only width is needed, change `size` to `size?.width` instead for more granularity and to prevent an unwanted re-render.

e2corporation: > Should I break `toggleSidebar` to two different functions ?
> This may solve the issue

That may be unlikely and cause a similar condition with only 2 more functions to troubleshoot. I think if you troubleshoot this original function further it can be resolved.

e2corporation: @henit-chobisa If you have resolved the main Effect without having to add a callback to toggleSidebar... then we can remove the callback there if it isn't required.

henit-chobisa: > @henit-chobisa for toggle sidebar func which dep are you providing? The entire `uiContext` object? or just `uiContext.changeSidebarVisibility` ?

Initially I gave, `uiContext.changeSidebarVisibility`, everything working nicely, but the linter still ask me for the whole uiContext object.

henit-chobisa: > @henit-chobisa If you have resolved the main Effect without having to add a callback to toggleSidebar... then we can remove the callback there if it isn't required.

No Julien, I resolved it with the help of useCallback

config-ui/src/components/Nav.jsx
```diff
@@ -15,73 +15,104 @@
...
+import React, { useContext, useEffect, useState } from 'react'
+import { Position, Popover, Navbar, Icon } from '@blueprintjs/core'
 import '@/styles/nav.scss'
 import { ReactComponent as SlackIcon } from '@/images/slack-mark-monochrome-black.svg'
 import { ReactComponent as SlackLogo } from '@/images/slack-rgb.svg'
+import UIContext from '@/store/UIContext'
+import useWindowSize from '@/hooks/useWIndowSize'
 
 const Nav = () => {
+  const uiContext = useContext(UIContext)
```
e2corporation, line 27: @henit-chobisa Does it help the situation any if you used De-structuring here?
```suggestion
  const { changeSidebarVisibility, desktopBreakpointWidth, sidebarVisible } = useContext(UIContext)
```

Then instead of referencing the Entire `UIContext` object, we can just utilize the De-structured children props?

config-ui/src/hooks/useWIndowSize.jsx
```diff
@@ -0,0 +1,47 @@
...
+          width: window.innerWidth,
+          height: window.innerHeight
+        })
+      }
+
+      window.addEventListener('resize', handleResize)
+
+      handleResize()
+
+      return () => window.removeEventListener('resize', handleResize)
```
e2corporation, line 41: The cleanup function for this Effect should not be inside the `IF` statement, it should be  moved out.

config-ui/src/hooks/useWIndowSize.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+function useWindowSize() {
+  const [windowSize, setWindowSize] = useState({
+    width: 320,
+    height: 240
+  })
+
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
+      function handleResize() {
```
e2corporation, line 29: This `handleResize` Closure should also be lifted out of the conditional and be declared on Ln `28`, which will also fix the reference issue when the cleanup function is moved out of the conditional (see comment on `Ln` 41)

config-ui/src/components/Nav.jsx
henit-chobisa: Yes Julien, it actually resolves the issue.
No side-effects and no linter issues, 
Thank you so much!!!

config-ui/src/hooks/useWIndowSize.jsx
henit-chobisa: Resolved 🫡

henit-chobisa:
@e2corporation 
Julien, thank you so much for keeping up with me.
I have resolved the dependency issue in `Nav.jsx` and also the `useWindowSize` hook suggestions.
Kindly Review.

e2corporation:
> @e2corporation
> Julien, thank you so much for keeping up with me.
> I have resolved the dependency issue in `Nav.jsx` and also the `useWindowSize` hook suggestions.
> Kindly Review.

@henit-chobisa Thanks Henit! This is great news, I will continue review and run some local tests to determine if any other optimizations are needed. Thanks for troubleshooting the problem, this will be beneficial when we create additional contexts in the future now that we know that de-structuring played an important factor in how dependencies are treated with this implementation.

config-ui/src/hooks/useWIndowSize.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { useEffect, useState } from 'react'
+
+function useWindowSize() {
+  const [windowSize, setWindowSize] = useState({
+    width: 320,
+    height: 240
+  })
+
+  function handleResize() {
```
e2corporation, line 27: @henit-chobisa I see that this was lifted completely out of the Effect, although it could have been kept inside the Effect as a local function (just outside the IF statement as recommended).

Given that this is now a loose function outside the Effect, I wonder if it has any negative side effects since it's not wrapped in a callback, it also doesn't appear to be a dependency on `Ln 40`?

It may be better to move the `handleResize` _inside_ the Effect, directly after `Ln 34`, but _outside_ the `IF` statement.

henit-chobisa: Sure Julien, I do the same.

henit-chobisa:
> > @e2corporation
> > Julien, thank you so much for keeping up with me.
> > I have resolved the dependency issue in `Nav.jsx` and also the `useWindowSize` hook suggestions.
> > Kindly Review.
> 
> @henit-chobisa Thanks Henit! This is great news, I will continue review and run some local tests to determine if any other optimizations are needed. Thanks for troubleshooting the problem, this will be beneficial when we create additional contexts in the future now that we know that de-structuring played an important factor in how dependencies are treated with this implementation.

Yes Absolutely, de-structuring was a very smart way to resolve the problem.

e2corporation (approved):
LGTM

klesh:
@henit-chobisa @e2corporation Great Collaboration! Glad to see this merged. Thank you all.

henit-chobisa:
> @henit-chobisa @e2corporation Great Collaboration! Glad to see this merged. Thank you all.

Looking forward for more like this! 🔥🔥
Grateful to the community and @e2corporation

PR #3306, opened 2022-10-04
feat: enable rate limit per hour toggle options

### LINKED ISSUE #3229: [Question][config-ui] why the rate limit of Jira defaults to 0
## Question
Why the rate limit of Jira defaults to 0? I understand that there might be a default rate limit there, but it can't be 0?

## Screenshots
image

## Additional context
Add any other context here.

e2corporation:
### ⚠️ Config-UI / Connections / Rate Limit "Toggle Options"

- [ ] Add **Toggle Switch** to control Rate Limit Options

Closes #3229

### Screenshots
Screen Shot 2022-09-30 at 10 07 33 PM
Screen Shot 2022-09-30 at 10 09 43 PM


klesh:
what if users set a value for it, and then disable it afterward, would this sent a value of `0` to the API ?

e2corporation:
> what if users set a value for it, and then disable it afterward, would this sent a value of `0` to the API ?

@klesh I need to update this so disabling will reset value to 0

e2corporation:
@klesh Fixed, user must hit **Save** as usual after _Disabling_ rate limit switch, and it will be correctly reset to 0

e2corporation:
@klesh This is ready for `main` branch next.

e2corporation:
@klesh @likyh This needs to be merged to `main` next to prevent conflicts with upcoming features.

config-ui/src/data/Providers.js
```diff
@@ -105,6 +105,28 @@ const ProviderFormLabels = {
     rateLimitPerHour: (
       <>
         Rate Limit <sup>(per hour)</sup>
+        <Tooltip
```
likyh, line 108: Let's don't copy these codes. Maybe we can add a var to reduce duplication.

e2corporation: This will be replaced by JSON Configuration in Plugin Registry in the future, I'll move to a constant in the meantime.

e2corporation: @likyh Fixed, Rate Limit Tooltip Extracted `@data/ConnectionTooltips.js` so logic isn't duplicated.

Issue #3346, opened 2022-10-09
[Question][Blueprints] subtask get jira issue get errors


greenpean:
## Question
When collector execution collect issue ended by exceptions

## Screenshots
![image](https://user-images.githubusercontent.com/34735006/194751621-135c35f9-17d7-45eb-960a-12aaf754109e.png)

greenpean:
According to the log may be this is the reason for the exception
![image](https://user-images.githubusercontent.com/34735006/194787189-af2446f5-7647-4dfb-a9a8-1507d9aff7b2.png)

klesh:
Seems like the api requests were timeouted.
Can it be reproduced stably?

Startrekzky:
Hey @greenpean  which version are you using?

greenpean:
@klesh Every job fails caused by the same reason
![image](https://user-images.githubusercontent.com/34735006/194970983-8d81eaa7-d75d-43d6-919b-070bfca7b87c.png)

greenpean:
@Startrekzky v8.16.0

klesh:
@greenpean Would you mind doing some benchmarking against your jira server API ?
Steps to take:

1. download the script from PR #3370 on the server that you deployed the `Apache Devlake`
2. fill in the USER/PASSWORD/HOST and BOARD in the script
3. execute the script and paste the output here

The script would capture the outgoing/incoming headers if failed, and the authentication information should be filtered out automatically, however, please double-check before you post the output, and make sure there is no sensitive information gets exposed.

Thanks

greenpean:
![image](https://user-images.githubusercontent.com/34735006/195243536-31d2ee27-c19d-4eab-9e1a-92492749724c.png)

klesh:
Thanks, that looks alright...
Can you try to dial down the ratelimit for the jira connection? Try 1800 first, see if error persists

![image](https://user-images.githubusercontent.com/61080/195249805-83212c4d-567a-4b16-a19e-f2f037dfa368.png)

klesh:
Thanks, that looks alright...
Can you try to dial down the ratelimit for the jira connection? Try 1800 first, see if error persists

![image](https://user-images.githubusercontent.com/61080/195249805-83212c4d-567a-4b16-a19e-f2f037dfa368.png)

greenpean:
@klesh thks for solving the problem.the issue can be collected successful.

klesh:
Glad to help.
Closing this issue, feel free to reopen if needed.

PR #3370, opened 2022-10-11
feat: add jira api benchmarking script

klesh:
# Summary

Some of our users keep running into jira API timeout, this script would help us identify the root cause of the problem.

Related to #3346


mindlesscloud (approved):
lgtm

PR #3477, opened 2022-10-18
refactor: add providerId to entity type data models

### LINKED ISSUE #3411: [Refactor][Config-UI] Add providerId to Entity Type Data Models
## Config-UI / Entity `Data Models`

The "Entity" Type Data Models should have a hard-coded `providerId` property added/defined which will provide easier access to the Data Provider ID without having to always depend on a lookup to `$activeProvider`

Add `providerId` to the following Data Models (`models/`):

- GitHubProject
- GitlabProject
- JiraBoard
- JenkinsJob

e2corporation:
### Config-UI / Data Models / `Entity` Models

- [x] Add `providerId` prop to `GitHubProject`, `GitlabProject`, `JiraBoard` and `JenkinsJob`
- [x] Create `Variants` Global Data Constant (`@data/Variants.js`)

Closes #3411

### DIFF
--- a/config-ui/src/models/GithubProject.js
+++ b/config-ui/src/models/GithubProject.js
@@ -29,6 +29,7 @@
  * @property {string?} repo
  * @property {boolean?} useApi
  * @property {project|board?} variant
+ * @property {string?} providerId
  */
 class GitHubProject {
   constructor(data = {}) {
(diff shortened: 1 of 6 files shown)


config-ui/src/models/GithubProject.js
```diff
@@ -47,6 +48,7 @@ class GitHubProject {
 
     this.useApi = data?.useApi || false
     this.variant = data?.variant || 'project'
```
likyh, line 50: why we need variant and providerId? class `GitHubProject` have already tell us what the type is.

e2corporation: Yes we could do `typeof` check as an alternate means.

PR #3501, opened 2022-10-20
fix cherry-pick #3497. ConvertJobsMeta is only defined in main(0.15).

likyh:
# Summary

fix cherry-pick #3497. `ConvertJobsMeta` is only defined in main(0.15).


klesh (approved):
LGTM

PR #3532, opened 2022-10-21
fix: fix lake crushed when pipeline.beginAt is nil

### LINKED ISSUE #3531: [Bug][backend-framework] lake crushed when pipeline.beginAt is nil
### Search before asking

### What happened

Isdxz6O9WV
PUtBV7uqfG

don't crushed and log/save original error

### How to reproduce
@narrowizard 
img_v2_e9cd6024-47f0-449c-8475-771ae615b33g
1. run EE's pipeline

### Anything else

_No response_

### Version

main

### Code of Conduct

likyh:
# Summary

fix lake crushed when pipeline.beginAt is nil

Closes #3531

### DIFF
--- a/services/pipeline_runner.go
+++ b/services/pipeline_runner.go
@@ -122,7 +122,10 @@ func runPipeline(pipelineId uint64) errors.Error {
 	// finished, update database
 	finishedAt := time.Now()
 	pipeline.FinishedAt = &finishedAt
-	pipeline.SpentSeconds = int(finishedAt.Unix() - pipeline.BeganAt.Unix())
+	if pipeline.BeganAt != nil {
+		// BeganAt may be nil caused by err
+		pipeline.SpentSeconds = int(finishedAt.Unix() - pipeline.BeganAt.Unix())
+	}
 	if err != nil {
 		pipeline.Status = models.TASK_FAILED
 		if lakeErr := errors.AsLakeErrorType(err); lakeErr != nil {


services/pipeline_runner.go
```diff
@@ -122,7 +122,10 @@ func runPipeline(pipelineId uint64) errors.Error {
 	// finished, update database
 	finishedAt := time.Now()
 	pipeline.FinishedAt = &finishedAt
```
keon94, line 124: In my opinion we should avoid a nil BeganAt value altogether, and fix the logic here instead:
https://github.com/apache/incubator-devlake/blob/7853386b27f68ebba7eb266b854a731f60db6d76/services/pipeline_helper.go#L39

We should set the BeganAt field on DbPipeline when we create it.

services/pipeline_runner.go
```diff
@@ -122,7 +122,10 @@ func runPipeline(pipelineId uint64) errors.Error {
...
-	pipeline.SpentSeconds = int(finishedAt.Unix() - pipeline.BeganAt.Unix())
+	if pipeline.BeganAt != nil {
```
keon94, line 125: I'm actually not finding any place in our code where BeganAt is ever set. Has this always been nil then?

likyh: https://github.com/apache/incubator-devlake/blob/68095c5e4bd95f7afeb1347bcd6232e98dc133ea/runner/run_pipeline.go#L69

Here is the only setter.

Issue #3451, opened 2022-10-17
[Bug][TAPD] We only have original value for changelogs when field name is status


warren830:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

![img_v2_67a02ff0-9b3c-4c26-b4ae-a36f9a78ac7g](https://user-images.githubusercontent.com/39366025/196126763-d4e90c4e-e424-402a-94a9-6e1fc7c107fd.jpg)
didn't parse changelog's status value to standard statuis

### What you expected to happen

should have std value

### How to reproduce

check issue_changelogs

### Version

main

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

warren830:
fixed by #3534

PR #3534, opened 2022-10-21
fix(tapd): translate changelog status value to chinese before standard

warren830:
# Summary

bug changelog's status field used original english value, we need to translate it to chinese first, then standardize it

Relate to  #3451 

### Screenshots

### Other Information


abeizn (approved):
LGTM

PR #3554, opened 2022-10-24
feat: cover orderby/since parameter for jira-api-benchmark

klesh:
# Summary

Enhance jira api benchmarking script to cover orderby/since parameters


mindlesscloud (approved):
LGTM

PR #3563, opened 2022-10-25
fix: delete 0.15 protection in .asf.yaml

likyh:
# Summary

delete 0.15 protection in `.asf.yaml`. It need add back when release-v0.15 is created.


klesh (approved):
LGTM

Issue #3574, opened 2022-10-26
[Bug][plugin] incorrect mapping from blueprint options to dbt argument


usharerose:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

In option description, `projectPath` is `user dbt project directory`. However, the value points to `dbt run`'s argument `--profiles-dir` which is

> --profiles-dir SUB_PROFILES_DIR
Which directory to look in for the profiles.yml file. If not set, dbt will look in the current working directory first, then HOME/.dbt/

### What you expected to happen

The value of `projectPath` should point to `--project-dir` which is

> --project-dir PROJECT_DIR
Which directory to look in for the dbt_project.yml file. Default is the current working directory and its parents.

### How to reproduce

Nothing

### Version

main

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #3575, opened 2022-10-26
fix: pass the value of projectPath to dbt project-dir argument

usharerose:
# Summary
pass the value of `projectPath` from dbt plugin to the argument `--project-dir` of the command `dbt run` instead of `--profiles-dir`

Closes https://github.com/apache/incubator-devlake/issues/3574


abeizn (approved):
LGTM

PR #3581, opened 2022-10-26
chore: increase the default value for API_TIMEOUT from 10s to 120s

### LINKED ISSUE #3580: [Feature][helper] Increase the default value of API_TIMEOUT from 10s to 120s
### Search before asking

### Description

Currently, we have our API_TIMEOUT default to `10s` to halt the pipeline ASAP so users may correct and rerun quickly.

However, in some environments, the JIRA SERVER API response time is far slower, we received multiple reports indicating the elapsed time could be higher than 40s.

To accommodate situations like that, I suggest that we increase the default API_TIMEOUT value to `120s`. Although it prolongs the time for the "correct and rerun" scenario, it is worth it because the chance of the scenario is quite low and negligible.

### Use case

Users with slow JIRA API response may still collect data without error

### Related issues

_No response_

### Code of Conduct

klesh:
# Summary

chore: increase the default value for API_TIMEOUT from 10s to 120s

Closes #3580


likyh (approved):
LGTM

PR #3608, opened 2022-10-28
feat: add bitbucket issueStatusOther for bitbucket transformationrules

### LINKED ISSUE #3368: [Feature][Bitbucket] add Bitbucket Issue Tracking transformationRules
### Search before asking

### Description

_No response_

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

thenicetgp:
# Summary
BitBucket does not have priority, component because the BitBucket issue has priority, component as semantically explicit attributes that do not need to be extracted from the issue labels like GitHub does.

BitBucket does not have a type configuration because
a. BitBucket issue also has a type field, so there is no need to extract it from labels like GitHub does.
b. BitBucket types are only "enhancement", "proposal", "task", "bug", and cannot be customized by users. So by default we transform "enhancement", "proposal" and "task" to REQUIREMENT and "bug" to BUG to save users' time.

BitBucket does not have severity because BitBucket does not have related fields, so let's put it aside for a while and do it later.

BitBucket has "status mapping" because unlike GitHub issues which only have 2 statuses "open" and "closed", BitBucket has a lot more. We can't transform them to the 3 standard issue statuses "TODO", "IN_PROGRESS" and "DONE" on behalf of users. Thus, we add the "status mapping" configuration to let users decide.

now add bitbucket a new Status mapping `issueStatusOther` in order to let users define some status they do not want to define.

Closes #3368 

### Screenshots

### Other Information

### DIFF
--- a/models/domainlayer/ticket/issue.go
+++ b/models/domainlayer/ticket/issue.go
@@ -65,4 +65,5 @@ const (
 	TODO        = "TODO"
 	DONE        = "DONE"
 	IN_PROGRESS = "IN_PROGRESS"
+	OTHER       = "OTHER"
 )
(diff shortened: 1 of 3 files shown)


warren830:
> Is this PR ready?

Seems like this is ready

warren830 (approved):
LGTM

Issue #3637, opened 2022-11-01
[Bug][plugin] raise exception when no input to the argument selectedModels for dbt plugin options


usharerose:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

According to the implementation of `dbt` plugin, `selectedModels` is an optional argument.

Currently, the task would fail when triggered according to the plan without this argument.

### What do you expect to happen

The task would run successfully under the condition that no specific models selected via this argument.

### How to reproduce

Trigger the pipeline with the following plan,
```
[
    [
        {
            "plugin": "dbt",
            "subtasks": null,
            "options": {
                "projectName": "demo_project",
                "projectPath": "/home/ubuntu/demo_project",
                "projectTarget": "dev"
            }
        }
    ]
]
```

and log shows that,
```
time="2022-11-01 04:42:24" level=info msg=" [pipeline service] [pipeline #61] [task #69] [dbt] [DbtConverter] dbt run script: %!(EXTRA *exec.Cmd=/usr/bin/dbt run --project-dir /home/ubuntu/demo_project --select)"
time="2022-11-01 04:42:29" level=error msg=" [task service] task failed
	caused by: Error running task 69.
	Wraps: (2) subtask DbtConverter ended unexpectedly
	Wraps: (3) exit status 2
	Wraps: (4) exit status 2
	Error types: (1) *hintdetail.withDetail (2) *hintdetail.withDetail (3) *hintdetail.withDetail (4) *exec.ExitError"
time="2022-11-01 04:42:29" level=error msg=" [pipeline service] [pipeline #61] run tasks failed
	caused by: Error running task 69."
```
which we can see that there is an extra `--select` without relative values

### Version

main

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

klesh:
@abeizn please take a look at this

PR #3638, opened 2022-11-01
fix: skip --select argument if input is empty

usharerose:
# Summary
When convert to `dbt` command line, skip `--select` if,
1. no input in the option (nil)
2. the slice is empty

Closes https://github.com/apache/incubator-devlake/issues/3637

### DIFF
--- a/plugins/dbt/tasks/convertor.go
+++ b/plugins/dbt/tasks/convertor.go
@@ -119,7 +119,7 @@ func DbtConverter(taskCtx core.SubTaskContext) errors.Error {
 		dbtExecParams = append(dbtExecParams, "--vars")
 		dbtExecParams = append(dbtExecParams, string(jsonProjectVars))
 	}
-	if models != nil {
+	if len(models) > 0 {
 		dbtExecParams = append(dbtExecParams, "--select")
 		dbtExecParams = append(dbtExecParams, models...)
 	}


plugins/dbt/tasks/convertor.go
```diff
@@ -119,8 +119,10 @@ func DbtConverter(taskCtx core.SubTaskContext) errors.Error {
 		dbtExecParams = append(dbtExecParams, "--vars")
 		dbtExecParams = append(dbtExecParams, string(jsonProjectVars))
 	}
-	dbtExecParams = append(dbtExecParams, "--select")
-	dbtExecParams = append(dbtExecParams, models...)
+	if models != nil {
```
mindlesscloud, line 122: The slice `models` could be empty but non-nil, I suggest using `len(models) > 0` instead of `models != nil`

usharerose: Hi, @mindlesscloud . Agree with your point that need to handle the status when the slice is empty. But it is also possible that `models` is null when not declaring `selectedModels` in `dbt` Options. So I guess the condition should be `models != nil && len(models) > 0`.

usharerose: agreed, updated.

mindlesscloud (approved):
LGTM

PR #3676, opened 2022-11-04
fix(plugins): add status RUNNING

### LINKED ISSUE #3660: [Bug][CICD] Keep result as empty for unfinished cicd tasks/pipelines
### Search before asking

### What happened

I just checked, all unfinished pipeline/tasks are also success 
At first, we tried to add `RUNNING` to indicate unfinished tasks/pipelines
But after deeper research, we found that for gitlab job, when it's unfinished, it has much more status like "created", "waiting_for_resource", "preparing", "pending", "running", "scheduled", jenkins also has a lot status to indicate unfinished.
More important, we found that in our original design, we should keep it empty.

keep their result as empty

### How to reproduce

none

### Anything else

_No response_

### Version

main

### Code of Conduct

warren830:
# Summary

use `RUNNING` to indicate that pipeline/task is unfinished
update e2e data for new status

closes #3660

### Screenshots

### Other Information


models/domainlayer/devops/cicd_pipeline.go
```diff
@@ -53,6 +53,7 @@ func (CICDPipelineRelationship) TableName() string {
 
 const (
 	SUCCESS     = "SUCCESS"
+	RUNNING     = "RUNNING"
```
abeizn, line 56: What is the difference between RUNNING and IN_PROGRESS?

warren830: RUNNING is for result, IN_PROGRESS is for status

warren830: Maybe we can keep it as empty?

abeizn: > RUNNING is for result, IN_PROGRESS is for status

then i think RUNNING and IN_PROGRESS maybe written separately, so that users can understand.

abeizn (dismissed):

abeizn (approved):
LGTM

PR #3681, opened 2022-11-07
feat(config-ui): support typescript and format webpack-related files

mintsweet:
# Summary

support typescript and format webpack-related files.

Prepare for the miller-columns component.

### Screenshots
Nothing.

### Other Information
Nothing.


mintsweet:
Why was typescript introduced in devlake?

Currently, many props in our code have been passed through many layers, and we cannot quickly understand the content and function of these props in subsequent use. As a result, we need to spend a lot of time to understand the original position of props, if we use typescript, we can clearly understand the content and function of each props.

likyh:
I really agree with it [Laugh and Cry]. It's not good that no TS is used in the transmission. Some places use parameters that are not passed in, and some pass in variables not defined before. But would you like to inform everyone about this?

likyh (approved):

PR #3732, opened 2022-11-14
fix(framework): modify table name for migrationscript

### LINKED ISSUE #3718: [Bug][Gitextractor] Report error
### Search before asking

### What happened
The plugin gitextractor failed right after it started. I tried several times, and it reported the same error
The error was shown as below
image

This error was reported at main branch,  the commit sha of head is `9e3d1ba886730107276788ecd7552f11c53bb2fa`
After checking log: 
image
And I found that there is a table like:
image

Then I checked related migration script:
image
So here we should use: commit_file_components

none

### How to reproduce

collect github/gitlab by default

### Anything else

_No response_

### Version

main

### Code of Conduct

warren830:
# Summary

According to #3718 , we only made a small change:
`commit_files` to `commit_file_components`

Closes #3718 

### Screenshots

### Other Information

### DIFF
--- a/models/migrationscripts/20220913_fix_commitfile_id_toolong.go
+++ b/models/migrationscripts/20220913_fix_commitfile_id_toolong.go
@@ -77,7 +77,7 @@ func (script *fixCommitFileIdTooLong) Up(basicRes core.BasicRes) errors.Error {
 	return migrationhelper.TransformTable(
 		basicRes,
 		script,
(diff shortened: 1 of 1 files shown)


likyh (approved):
This PR LGTM. And as it looks cause by #3176, let check #3176 again?

warren830:
> 

image
already checked, came from #3176

PR #3761, opened 2022-11-18
Don't check network connection when service is behind a proxy

albinvass:
### ⚠️ Pre Checklist

### Summary
Currently the NewApiClient function errors if the service is behind a proxy since it cannot establish a direct connection.
Instead skip the network test when a proxy is configured.

### DIFF
--- a/plugins/helper/api_client.go
+++ b/plugins/helper/api_client.go
@@ -84,10 +84,13 @@ func NewApiClient(
 	if err != nil {
 		return nil, errors.Default.New("Failed to resolve Port")
 	}
-	err = utils.CheckNetwork(parsedUrl.Hostname(), port, 10*time.Second)
-	if err != nil {
-		return nil, errors.Default.Wrap(err, "Failed to connect")
-	}
+
+    if proxy != "" {
+        err = utils.CheckNetwork(parsedUrl.Hostname(), port, 10*time.Second)
+        if err != nil {
+            return nil, errors.Default.Wrap(err, "Failed to connect")
+        }
+    }
 	apiClient := &ApiClient{}
 	apiClient.Setup(
 		endpoint,
--- a/plugins/helper/api_client.go
+++ b/plugins/helper/api_client.go
@@ -85,7 +85,7 @@ func NewApiClient(
 		return nil, errors.Default.New("Failed to resolve Port")
 	}
 
-    if proxy != "" {
+    if proxy == "" {
         err = utils.CheckNetwork(parsedUrl.Hostname(), port, 10*time.Second)
         if err != nil {
             return nil, errors.Default.Wrap(err, "Failed to connect")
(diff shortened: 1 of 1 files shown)


plugins/helper/api_client.go
```diff
@@ -84,10 +84,13 @@ func NewApiClient(
 	if err != nil {
 		return nil, errors.Default.New("Failed to resolve Port")
 	}
-	err = utils.CheckNetwork(parsedUrl.Hostname(), port, 10*time.Second)
-	if err != nil {
-		return nil, errors.Default.Wrap(err, "Failed to connect")
-	}
+
+    if proxy != "" {
```
keon94, line 88: I'm not sure we'd want to skip checking the connection altogether. Could we not just call CheckNetwork() on the proxy instead?

albinvass: That we could do.

likyh:
But why not CheckNetwork by proxy?

albinvass:
> But why not CheckNetwork by proxy?

Well currently the network check is just trying to set up a connection to the server, which isn't possible because it's behind an http proxy. To test by proxy the helper object would need to do an http request. Not sure if that can be done in a generic enough way to support any http api (if ApiClient is supposed to be a generic helper). Maybe it's better to let the different plugins check connectivity to the service?

keon94:
> > But why not CheckNetwork by proxy?
> 
> Well currently the network check is just trying to set up a connection to the server, which isn't possible because it's behind an http proxy. To test by proxy the helper object would need to do an http request. Not sure if that can be done in a generic enough way to support any http api (if ApiClient is supposed to be a generic helper). Maybe it's better to let the different plugins check connectivity to the service?

Looking at [this post](https://stackoverflow.com/questions/14661511/setting-up-proxy-for-http-client), looks like there is a generic way to handle proxy calls (via the http.DefaultTransport modification). Perhaps we can leverage that?

plugins/helper/api_client.go
```diff
@@ -107,6 +110,16 @@ func NewApiClient(
...
 
 	if proxy != "" {
+        var pu, err := url.Parse(proxy)
+        if err != nil {
+            return nil, errors.BadInput.Wrap(err, fmt.Sprintf("Invalid Proxy URL: %s", proxy))
+        }
+
+        err = utils.CheckNetwork(pu.Hostname(), port, 10*time.Second)
+        if err != nil {
+            return nil, errors.Default.Wrap(err, "Failed to connect to proxy")
```
keon94, line 120: are you still encountering timeouts after making this change?

klesh (commented):
Thanks for the contribution, appreciated.
Seems like the code is testing the Proxy only, rather than testing the actual host via the proxy.
I think we should fix the `utils.CheckNetwork` first to make it support proxy.
In order to fix your problem ASAP, I will fix the problem in another PR and make a hotfix today, will keep you posted.

PR #3754, opened 2022-11-17
feat: add project to blueprint

mappjzc:
### Summary
To associate blueprint with project.
We need add project_name to blueprints table, and update API/Service accordingly.
Add migration script about blueprint.

***bp_get.sh***
```bash
#!/bin/sh
curl --location --request GET 'devlake:8080/blueprints/1' \
--header 'Content-Type: application/json' \
--data-raw '{
}'

echo "\r\n"
```

***bp_patch.sh***
```bash
#!/bin/sh
curl --location --request PATCH 'devlake:8080/blueprints/1' \
--header 'Content-Type: application/json' \
--data-raw '
{
    "name":"Test BluePrint",
    "projectName":"TestProject",
    "cronConfig":"0 0 * * *",
    "plan":{},
    "settings":
    {
        "version":"1.0.0",
        "connections":
        [
            {
                "plugin":"gitlab",
                "connectionId":1,
                "scope":[
                    {
                        "transformation": {"productionPattern":""},
                        "options":{"projectId":37,"title":"思码逸 / Lake"},
                        "entities":["CODE","TICKET","CODEREVIEW","CROSS","CICD"]
                    }
                ]
            }
        ]
    },
    "enable":true,
    "mode":"NORMAL",
    "isManual":true
}'

echo "\r\n"
```

***bp_post_refuse.sh***
```bash
#!/bin/sh
curl --location --request POST 'devlake:8080/blueprints' \
--header 'Content-Type: application/json' \
--data-raw '
{
    "name":"Test BluePrint",
    "projectName":"TestProjectRefuse",
    "cronConfig":"0 0 * * *",
    "plan":{},
    "settings":
    {
        "version":"1.0.0",
        "connections":
        [
            {
                "plugin":"gitlab",
                "connectionId":1,
                "scope":[
                    {
                        "transformation": {"productionPattern":""},
                        "options":{"projectId":37,"title":"思码逸 / Lake"},
                        "entities":["CODE","TICKET","CODEREVIEW","CROSS","CICD"]
                    }
                ]
            }
        ]
    },
    "enable":true,
    "mode":"NORMAL",
    "isManual":true
}'

echo "\r\n"
```

***bp_post.sh***
```bash
#!/bin/sh
curl --location --request POST 'devlake:8080/blueprints' \
--header 'Content-Type: application/json' \
--data-raw '
{
    "name":"Test BluePrint",
    "cronConfig":"0 0 * * *",
    "plan":{},
    "settings":
    {
        "version":"1.0.0",
[...]

PR #3764, opened 2022-11-18
Issues/3234 add parallelLabels for blueprint/pipeline

### LINKED ISSUE #3234: [Refactor][pipeline] pipeline need a mode with higher controllability than PIPELINE_MAX_PARALLEL
### Search before asking

### Description

The pipeline now supports PIPELINE_MAX_PARALLEL to limit running parallel. But it's controllability is not enough. I'll use the following example to support my argument:

there are 2 blueprints:
1. blueprint 1: [ [jira], [customJiraPlugin], [starrocks] ]
2. blueprint 2: [ [github(repo1), gitextractor(repo1)], [github(repo2), gitextractor(repo2)], [starrocks] ]

Now, if I set PIPELINE_MAX_PARALLEL=1, sometimes the blueprint cannot run immediately. If I set PIPELINE_MAX_PARALLEL >= 2, the plugin startrocks may run parallel and it will cause some errors.

So I think we can add a label in plugins to solve this problem.

```go
func (plugin Refdiff) PapallelLabels() map[string]string {
        return string[]{
            "file" // it means all plugin with the label "file" cannot run parallel
        }
}
func (plugin Icla) PapallelLabels() map[string]string {
        return string[]{
            "$pluginName" // it means this plugin cannot run parallel
        }
}
func (plugin Jira) PapallelLabels() map[string]string {
        return string[]{
            "$connectionId" // it means the plugin have the pluginName and same connection id cannot run parallel
        }
}
```
The plugins with the same label are controlled by the same lock.

If we add this, we will avoid another case.

The blueprint is like the following example:
`[ [jira1, jira2], [starrocks] ]`

```go
func (plugin Jira) PapallelLabels() map[string]string {
        return string[]{
            "$connectionId" // it means the plugin have the pluginName and same connection id cannot run parallel
        }
}
```
We can add magic labels like the above one to prevent jira1 and jira2 from using the same rate limit at the same time. 

The env `PIPELINE_MAX_PARALLEL` will use keep and it means how many pipelines can run parallel.
A new env `TASK_MAX_PARALLEL` will be added for how many tasks can run parallel.

Example:
```env
PIPELINE_MAX_PARALLEL=1
TASK_MAX_PARALLEL=2
```
It means 1 pipeline can run together and at most 2 tasks at the same stage in the running pipeline can be started.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

likyh:
### Summary
1. add labels for blueprint/pipeline
2. use labels which start with `parallel/*` to control how pipeline run parallel.
3. [config-ui] use api `trigger` to run blueprint
4. delete unused param `retry`
5. fix a bug imported by https://github.com/apache/incubator-devlake/pull/3754: the response body from `/blueprints` changed.

image

Now the pipelines with the same labels which start with `parallel/*` (just need one) can only run one by one.
all labels in the blueprint will pass to the pipeline when triggered.

Closes #3234

### Screenshots
image
image
image

image
parallel_label in `Get /pipelines` must be a complete label name to filter.
image

### DIFF
--- a/models/migrationscripts/20221115_add_parallel_labels.go
+++ b/models/migrationscripts/20221115_add_parallel_labels.go
@@ -1,0 +1,69 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package migrationscripts
+
+import (
+	"github.com/apache/incubator-devlake/errors"
+	"github.com/apache/incubator-devlake/plugins/core"
+	"time"
+)
+
+type DbPipelineParallelLabel20221115 struct {
+	CreatedAt  time.Time `json:"createdAt"`
+	UpdatedAt  time.Time `json:"updatedAt"`
+	PipelineId uint64    `json:"pipeline_id" gorm:"primaryKey"`
+	Name       string    `json:"name" gorm:"primaryKey"`
+}
+
+func (DbPipelineParallelLabel20221115) TableName() string {
+	return "_devlake_pipeline_parallel_labels"
+}
+
+type DbBlueprintParallelLabel20221115 struct {
+	CreatedAt   time.Time `json:"createdAt"`
+	UpdatedAt   time.Time `json:"updatedAt"`
+	BlueprintId uint64    `json:"blueprint_id" gorm:"primaryKey"`
+	Name        string    `json:"name" gorm:"primaryKey"`
+}
+
+func (DbBlueprintParallelLabel20221115) TableName() string {
+	return "_devlake_blueprint_parallel_labels"
+}
+
+type addParallelLabels struct{}
+
+func (*addParallelLabels) Up(res core.BasicRes) errors.Error {
+	db := res.GetDal()
+	err := db.AutoMigrate(&DbPipelineParallelLabel20221115{})
+	if err != nil {
+		return err
+	}
+	err = db.AutoMigrate(&DbBlueprintParallelLabel20221115{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (*addParallelLabels) Version() uint64 {
+	return 20221115000034
+}
+
+func (*addParallelLabels) Name() string {
+	return "UpdateSchemas for addParallelLabels"
+}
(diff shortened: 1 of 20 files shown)


models/migrationscripts/20221115_add_parallel_labels.go
```diff
@@ -0,0 +1,69 @@
...
+
+type addParallelLabels struct{}
+
+func (*addParallelLabels) Up(res core.BasicRes) errors.Error {
+	db := res.GetDal()
+	err := db.AutoMigrate(&DbPipelineParallelLabel20221115{})
+	if err != nil {
+		return err
+	}
+	err = db.AutoMigrate(&DbBlueprintParallelLabel20221115{})
```
warren830, line 56: why there are two automigrate for the same table?

models/migrationscripts/20221115_add_parallel_labels.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return nil
+}
+
+func (*addParallelLabels) Version() uint64 {
+	return 20221115000034
+}
+
+func (*addParallelLabels) Name() string {
+	return "UpdateSchemas for addParallelLabels"
```
warren830, line 68: `add _devlake_blueprint_parallel_labels` is better?

warren830, line 68: what about `add parallel labels' schema for blueprint and pipeline`

likyh: fixed

mappjzc (dismissed):

models/migrationscripts/20221115_add_parallel_labels.go
```diff
@@ -0,0 +1,69 @@
...
+
+type addLabels struct{}
+
+func (*addLabels) Up(res core.BasicRes) errors.Error {
+	db := res.GetDal()
+	err := db.AutoMigrate(&DbPipelineLabel20221115{})
+	if err != nil {
+		return err
+	}
+	err = db.AutoMigrate(&DbBlueprintLabel20221115{})
```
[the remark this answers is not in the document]
likyh: good idea

services/pipeline.go
```diff
@@ -165,29 +167,57 @@ func RunPipelineInQueue(pipelineMaxParallel int64) {
...
+				pipelineParallelLabels = append(pipelineParallelLabels, dbLabel.Name)
+			}
+		}
+		runningParallelLabels = append(runningParallelLabels, pipelineParallelLabels...)
+
+		go func(pipelineId uint64, parallelLabels []string) {
 			defer sema.Release(1)
-			globalPipelineLog.Info("run pipeline, %d", pipelineId)
+			defer func() {
+				runningParallelLabels = utils.SliceRemove(runningParallelLabels, parallelLabels...)
```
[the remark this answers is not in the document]
likyh: fix, add lock for this

mintsweet (approved):
LGTM.

klesh (dismissed):
LGTM

services/pipeline.go
```diff
@@ -167,27 +171,58 @@ func RunPipelineInQueue(pipelineMaxParallel int64) {
 		dbPipeline := &models.DbPipeline{}
 		for {
 			cronLocker.Lock()
+			// prepare query to find an appropriate pipeline to execute
 			db.Where("status IN ?", []string{models.TASK_CREATED, models.TASK_RERUN}).
+				Joins(`left join _devlake_pipeline_labels ON
+                  _devlake_pipeline_labels.pipeline_id = _devlake_pipelines.id AND
```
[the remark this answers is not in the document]
likyh: fixed

klesh (approved):
LGTM

PR #3766, opened 2022-11-21
feat: refactor impl for all plugins

### LINKED ISSUE #3762: [Refactor][plugins] Unified interfaces of all plugins are implemented in impl
## What and why to refactor
Now the most of plugins are implemented in impl.
But there are still a few plugins do it on the main file,like starrocks、refdiff and so on.
image
image

It would better to Unified all of it in impl.

Move all of them into the impl.
At the same time,refactor the all code of plugin into the right package.

mappjzc:
### Summary
Most of plugins were implement relevant interfaces under impl.

But there are still few plugins were implement relevant interfaces under there main file.

It is better to move the implement of them to impl.

The current plugins for collating impl are as follows：
- dbt
- gitextractor
- github_graphql
- ida
- refdiff
- starrocks

Closes #3762


likyh (approved):
LGTM

Issue #3671, opened 2022-11-04
[Feature][Config UI] Change the layout of Set Data Scope from tabs to a table


yumengwang03:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

We have received many complaints on how users may skip the tabs on the Set Data Scope and Set Transformations page, and therefore we want to update this layout to be a table.

Figma link: https://www.figma.com/file/Eux4Dhorxp7M0TwzOnL6Yh/Config-UI-for-V1.0?node-id=2372%3A16637

Current:
<img width="840" alt="Screen Shot 2022-11-04 at 11 09 36 AM" src="https://user-images.githubusercontent.com/9535430/199876764-b4430149-908d-4764-835a-56a77444f0b2.png">

The new layout:
<img width="625" alt="Screen Shot 2022-11-04 at 11 10 01 AM" src="https://user-images.githubusercontent.com/9535430/199876807-bd337e39-1faf-4aad-9df4-45e31b9fc538.png">

### Related issues

https://github.com/apache/incubator-devlake/issues/2271
https://github.com/apache/incubator-devlake/issues/3670

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Startrekzky:
@mintsweet Is this issue also included in v0.15?

mintsweet:
Already done.

Issue #3468, opened 2022-10-18
[Feature][dora] Support DORA metrics stage 2


klesh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

#### Documentation

1. apache/incubator-devlake-website#298 @klesh 
2. Update [DORA manuals](https://devlake.apache.org/docs/UserManuals/DORA#how-to-implement-dora-metrics-with-apache-devlake ) based on the `Project` design in DORA stage II @Startrekzky 
3. Update [Webhook](https://devlake.apache.org/docs/Plugins/webhook#add-a-new-webhook) manuals for reference @Startrekzky 
4. Update [Config UI tutorial](https://devlake.apache.org/docs/UserManuals/ConfigUI/Tutorial) to be based on new user flow, eg. project, transformation rules, etc. @yumengwang03 
5. Add interface doc to [development doc](https://devlake.apache.org/docs/DeveloperManuals/PluginImplementation) 
6. Update `options` for all plugins, eg. scope_id, transformation_rule_id, etc.
7. Update Jenkins doc. Use Jenkins Job as the data scope.
8. Update the old interface in the [migration scripts](https://devlake.apache.org/docs/DeveloperManuals/DBMigration)
9. Update the [domain layer schema](https://devlake.apache.org/docs/DataModels/DevLakeDomainLayerSchema)
10. Add a doc `Blueprint Plan Generation Version v2.0.0`
11.  Update the calculation logic in [Change Lead Time](https://devlake.apache.org/docs/Metrics/LeadTimeForChanges)https://github.com/apache/incubator-devlake-website/pull/310
12. Update the [refdiff](https://devlake.apache.org/docs/Plugins/refdiff) to support `project`

#### Implementation

1. `project` support on the framework level: project is a set of `scopes` @mappjzc 
   1. Define an interface named `PluginMetric` with methods:
      a. `RequiredDataEntities` returns a list of required data entities and expected features. i.e. DORA should return sth like `[{ "model": "cicd_tasks", "requiredFields": {"column": "type", "execptedValue": "Deployment"}}, ...]` (no concreate usage at this point, can be used in config-ui to describe dependency in the future.)
      b. `IsProjectMetric` returns if the metric depends on Project for calculation. Currently, only `dora` would return true.
      c. `RunAfter` indicates which plugins must be executed before executing this one. (no concreate usage at this point)
[...]

Issue #2845, opened 2022-08-26
[Feature][Config UI] Bulk action for adding/editing transformation rules


yumengwang03:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Right now it's difficult to set transformation rules if the users have many repos/boards/etc., because the UI only allows them to add/edit one by one. To solve this problem, we have designed a transformation templates prototype so that users can flexibly add/edit/unlink transformation rules.

Figma is here: https://www.figma.com/file/Eux4Dhorxp7M0TwzOnL6Yh/Config-UI-for-V1.0?node-id=2734%3A22297
Please note, this issue is relate to https://github.com/apache/incubator-devlake/issues/2271 (The prototypes are in the same place so that you can see the entire workflow.)

### Related issues

https://github.com/apache/incubator-devlake/issues/2271

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

yumengwang03:
Figma is here: https://www.figma.com/file/Eux4Dhorxp7M0TwzOnL6Yh/Config-UI-for-V1.0?node-id=2372%3A16637
Please note, this issue is relate to #2271 (The prototypes are in the same place so that you can see the entire workflow.)

PR #3770, opened 2022-11-21
feat(config-ui): encapsulate base layout component

mintsweet:
### Summary
1. encapsulate **base layout** component.
2. adjust page routing to support the **global** layout.
3. remove old content about **nav** and **sidebar**.

### Related Issues

#2845 #3671 #3468 

### Screenshots
screenshot-20221122-133602

### Other Information
Nothing.


config-ui/src/registry/index.ts
```diff
@@ -0,0 +1,51 @@
...
+import { default as Feishu } from './plugins/feishu.json'
+import { default as Gitee } from './plugins/gitee.json'
+import { default as GitExtractor } from './plugins/gitextractor.json'
+import { default as GitHub } from './plugins/github.json'
+import { default as GitLab } from './plugins/gitlab.json'
+import { default as Jenkins } from './plugins/jenkins.json'
+import { default as JIRA } from './plugins/jira.json'
+import { default as Refdiff } from './plugins/refdiff.json'
+import { default as Starrocks } from './plugins/starrocks.json'
+import { default as TAPD } from './plugins/tapd.json'
```
likyh, line 33: It's same to `import TAPD from './plugins/tapd.json'` ?

config-ui/src/App.js
```diff
@@ -57,66 +63,104 @@ function App(props) {
...
+      <Switch>
+        <Route exact path='/offline' component={() => <Offline />} />
+
+        <Route>
+          <BaseLayout>
+            <Switch>
+              <Route
+                path='/'
+                exact
+                component={() => <Redirect to='/integrations' />}
```
likyh, line 75: Is it the same that writing in component and children?

likyh:
It becomes thinner, right?
image

config-ui/src/App.js
mintsweet: This way of writing is the same, but what I want is `component={<XXX />}`

However, there is a problem with the export of the current page. It can only be compatible with the writing method and will be improved in the future.

config-ui/src/registry/index.ts
mintsweet: hhhh, It's the same.

My original plan was to use `export { default as xx } from 'xx',` but later I found it was still inconvenient, so I changed it directly to `import`.

Updated.

PR #3781, opened 2022-11-22
fix: proxy connectivity checking

klesh:
### Summary

The `NewApiClient` function checked API Server connectivity without taking the `proxy` into consideration, which leads to a usability issue, PR #3761 tried to fix the problem by checking the Proxy Server connectivity instead, this PR solve the problem with a different approach:

1. When `proxy` was given, we check the connectivity by sending out a request to the API Server because it is not possible to distinguish DNS Resolution / Port Connectivity while `proxy` is in play. (yes, I learned it the hard way)
2. When `proxy` was not given, we fallback to the good old Checking Mechanism because it helps when our system was deployed to an internal K8s cluster, it is more likely for a k8s cluster to run into a DNS Resolution / Port Connectivity situations

Closes #3761

### DIFF
--- a/plugins/helper/api_client.go
+++ b/plugins/helper/api_client.go
@@ -69,25 +69,6 @@ func NewApiClient(
 	proxy string,
 	br core.BasicRes,
 ) (*ApiClient, errors.Error) {
-	parsedUrl, err := url.Parse(endpoint)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


likyh (approved):
Emm, `CheckNetwork` and `CheckDNS` provide more information about why the network failed, but they can not work with proxy now. So they move into `if` and request API by proxy.

PR #3791, opened 2022-11-23
Feat dora implement plugin blueprint v200

warren830:
### Summary
Implement plugin blueprint v200 as described in issue #3468 

relate to #3468 

### Screenshots

### Other Information


plugins/dora/api/blueprint_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+)
+
+func TestMakePipelinePlan(t *testing.T) {
+	const projectName = "TestMakePlanV200-project"
+	const productionPattern = "production--"
+	// mock dora plugin as a metric plugin
+	option := map[string]interface{}{
+		"projectName": projectName,
+		"transformationRules": map[string]interface{}{
```
klesh, line 33: cases:
1. when dora is executed as a metric plugin, it should not accept transformation rules.
2. when dora is invoked by another plugin, it then would have transformation rules.

`MakeMetricPluginPipelinePlanV200` is for case #1. no transformation rules needed here.

plugins/dora/api/blueprint.go
```diff
@@ -0,0 +1,54 @@
...
+package api
+
+import (
+	"encoding/json"
+	"github.com/apache/incubator-devlake/errors"
+	"github.com/apache/incubator-devlake/plugins/core"
+	"github.com/apache/incubator-devlake/plugins/dora/tasks"
+)
+
+func MakePipelinePlan(projectName string, options json.RawMessage) (core.PipelinePlan, errors.Error) {
```
klesh, line 27: The `dora` plugin doesn't support `PluginBlueprintV100`, no need to separate the logic into 2 functions.
I suggest that we move these logic to `MakeMetricPluginPipelinePlanV200`

warren830: fixed

plugins/dora/api/blueprint_test.go
warren830: fixed

klesh (approved):
LGTM

PR #3797, opened 2022-11-23
refactor: add transformation_rule_id to _tool_jira_boards

mindlesscloud:
### Summary
In addition to fixing some omissions in #3753, this PR contains the following modification:
1. add `transformation_rule_id` to `_tool_jira_boards`
2. provide `PATCH` API for Jira scope

No, but partially fulfilled issue #3468

### Screenshots
the scope API
image

### Other Information

### DIFF
--- a/plugins/jira/api/scope.go
+++ b/plugins/jira/api/scope.go
@@ -24,16 +24,10 @@ import (
 	"github.com/apache/incubator-devlake/errors"
 	"github.com/apache/incubator-devlake/plugins/core"
 	"github.com/apache/incubator-devlake/plugins/core/dal"
+	"github.com/apache/incubator-devlake/plugins/helper"
 	"github.com/apache/incubator-devlake/plugins/jira/models"
 	"github.com/mitchellh/mapstructure"
 )
-
-type putBoardRequest struct {
-	ProjectId uint   `json:"projectId"`
-	Name      string `json:"name"`
-	Self      string `json:"self"`
-	Type      string `json:"type"`
-}
 
 // PutScope create or update jira board
 // @Summary create or update jira board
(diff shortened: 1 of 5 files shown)


plugins/jira/api/scope.go
```diff
@@ -24,52 +24,66 @@ import (
...
+// @Accept application/json
+// @Param connectionId path int false "connection ID"
+// @Param boardId path int false "board ID"
+// @Param scope body models.JiraBoard true "json"
+// @Success 200  {object} models.JiraBoard
+// @Failure 400  {object} shared.ApiBody "Bad Request"
+// @Failure 500  {object} shared.ApiBody "Internal Error"
+// @Router /plugins/jira/connections/{connectionId}/scopes/{boardId} [PATCH]
+func UpdateScope(input *core.ApiResourceInput) (*core.ApiResourceOutput, errors.Error) {
+	board, err := extractBoard(input)
```
klesh, line 69: This is not needed. What we need to do:
1. load the board from db
2. DecodeMapStruct from input.body to the board (this is essentially a patch operation)
3. verify and save

plugins/jira/api/scope.go
```diff
@@ -122,3 +149,16 @@ func extractParam(params map[string]string) (uint64, uint64) {
...
 }
+
+func verifyBoard(board *models.JiraBoard) errors.Error {
+	if board.ConnectionId == 0 {
+		return errors.BadInput.New("invalid connectionId")
+	}
+	if board.BoardId == 0 {
+		return errors.BadInput.New("invalid boardId")
+	}
+	if board.ScopeId != strconv.FormatUint(board.BoardId, 10) {
```
likyh, line 160: why not set `board.ScopeId = board.BoardId` instead reading it from frontend?

klesh: I think it is for config-ui to display the scope in a uniform way, especially on the pipeline page.
config-ui can display scope id/name directly without detecting plugin.

klesh: come to think about it, do we have `ScopeName` as well? config-ui would need this for a human-readable representation.

likyh: ScopeName is easy to set. But ScopeId maybe has some complex rules to calculate.

mindlesscloud: I wish to expose `scopeId`, `transformationRuleId`, and `connection_id` as common fields among all of the plugin's scope API

Issue #3789, opened 2022-11-23
[Bug][Zentao] execution no data


luna6677:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

I choose zentao execution‘s board ,but have no data
![5DgUBL27Or](https://user-images.githubusercontent.com/48425466/203496244-3c8ec6e4-8d0f-4ae2-8f9e-828bae5c4160.jpg)

### What do you expect to happen

need data

### How to reproduce

always

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #3798, opened 2022-11-23
fix(zentao): fix GetTotalPages

warren830:
### Summary
According to issue #3789 , we also found we can not collect enough data due to a bug in `GetTotalPagesFromResponse`
This pr is to fix this bug. You can find the detail in `Files changed`

Relate to #3789 

### Screenshots

### Other Information


likyh (approved):
It looks like `page` does exist in response, so use Total and PageSize.

Issue #3685, opened 2022-11-07
[Feature][refdiff] enhance refdiff to support project


abeizn:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Enhance refdiff to support project: refdiff should take the project_name from taskOptions, and offer the ability to calculate refs_commits_diffs between 2 consecutive deployments under the same scope, namely cicd_scopes.

Attack plan:
1. Improvements to migration scripts.  pr: https://github.com/apache/incubator-devlake/pull/3703
2. Dashboard adaptation using refs_commits_diffs. pr:https://github.com/apache/incubator-devlake/pull/3736
3. Program adaptation calculated using the refs_commits_diffs table. pr: https://github.com/apache/incubator-devlake/pull/3726
4. The adaptation of e2e.
5. Updated domain layer documentation and plugin documentation.
6. Adaptation of code that supports Project. pr: https://github.com/apache/incubator-devlake/pull/3684

### Related issues

#3468 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #3801, opened 2022-11-24
feat: add new table ref_commits for the refdiff plugin

abeizn:
### ⚠️ Pre Checklist

### Summary
1. add new table ref_commits for the refdiff plugin.
2. update some struct name

related to #3685 

### Screenshots

### Other Information

### DIFF
--- a/models/domainlayer/code/commits_diff.go
+++ b/models/domainlayer/code/commits_diff.go
@@ -28,11 +28,22 @@ func (CommitsDiff) TableName() string {
 	return "commits_diffs"
 }
 
-type FinishedCommitsDiffs struct {
+type RefCommit struct {
+	NewRefId     string `gorm:"primaryKey;type:varchar(255)"`
+	OldRefId     string `gorm:"primaryKey;type:varchar(255)"`
+	NewCommitSha string `gorm:"type:varchar(40)"`
+	OldCommitSha string `gorm:"type:varchar(40)"`
+}
+
+func (RefCommit) TableName() string {
+	return "ref_commits"
+}
+
+type FinishedCommitsDiff struct {
 	NewCommitSha string `gorm:"primaryKey;type:varchar(40)"`
 	OldCommitSha string `gorm:"primaryKey;type:varchar(40)"`
 }
 
-func (FinishedCommitsDiffs) TableName() string {
+func (FinishedCommitsDiff) TableName() string {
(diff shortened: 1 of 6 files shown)


models/domainlayer/code/commits_diff.go
```diff
@@ -28,11 +28,22 @@ func (CommitsDiff) TableName() string {
 	return "commits_diffs"
 }
 
-type FinishedCommitsDiffs struct {
+type RefCommit struct {
```
likyh, line 31: Why does this table need to be old and new? It seems that one refId and one commitSha are enough.

plugins/refdiff/tasks/commit_diff_calculator.go
```diff
@@ -41,18 +42,29 @@ func CalculateCommitsDiff(taskCtx core.SubTaskContext) errors.Error {
...
+				newRefId := fmt.Sprintf("%s:%s", repoId, pair[2])
+				oldRefId := fmt.Sprintf("%s:%s", repoId, pair[3])
+				if pair[2] != newRefId || pair[3] != oldRefId {
+					refCommit.NewCommitSha = pair[0]
+					refCommit.OldCommitSha = pair[1]
+					refCommit.NewRefId = newRefId
+					refCommit.OldRefId = oldRefId
+				}
+
 				commitPairs = append(commitPairs[:i], commitPairs[i+1:]...)
```
mindlesscloud, line 68: Removing items from the `commitPairs` would affect the underlying `data.Options.AllPairs` which is shared among all subtasks.

models/domainlayer/code/commits_diff.go
abeizn: yes, one is not enough

likyh: Maybe adding column in finished_commit_diffs is better. The solution that create A new table is just move duplicate row to new table.

abeizn: no, because the deployment does not use it in finished_commit_diffs

klesh: If one needs to know which `commit-sha` a `ref` is pointing to, one should use the `refs` table.
This table is designed to store the pairs that are filtered by the `refdiff` tag regex.

mindlesscloud (approved):
LGTM

Issue #3805, opened 2022-11-25
[Feature][Jira] Provide a time filter to limit the Jira data to collect


Startrekzky:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

As an product owner, I want to collect Jira data with DevLake to measure issue-tracking metrics. It takes ages to finish the collection of all historical data; while I'm not going to use the data before 6 months.

### Description

Similar to #3688, the core idea is to limit the data to collect by selecting a time range. As a result, the collection speed of Jira data will be shortened evidently.

DevLake will collect the last six months of Jira data by default. Users can modify the date range. 

#### Affected entities
This time filter will affect the following entities by the corresponding time field:
- issues: `created_date`. All comments, changelogs, and worklogs under these issues will be collected.
- sprints: `started_date`.

Non-affected entities:
- boards
- accounts: it's fine to collect the full Jira account list

#### Prototype
Figma: https://www.figma.com/file/Eux4Dhorxp7M0TwzOnL6Yh/Config-UI-for-V1.0?node-id=2610%3A22925
<img width="910" alt="Screen Shot 2022-11-28 at 5 23 55 PM" src="https://user-images.githubusercontent.com/9535430/204241023-3fdfa958-ca86-4c3d-8b84-5dabb6990c10.png">

### Related issues

#3688
#3492 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

klesh:
Seems like the BE part is ready, @mintsweet please pick it up for the config-ui

likyh:
closed by https://github.com/apache/incubator-devlake/pull/3916

mdapke-insight:
Is there a way the time Range is always set to 'last 6 months'? So, at any given point in time, the data that devlake is collecting is the last 6 months without us needed to manually set it again and again

d4x1:
@mdapke-insight Your advice makes sense. You can submit a new feature request issue where we can discuss it in detail.

PR #3813, opened 2022-11-28
feat: add collector_meta and CreatedDateAfter for jira

likyh:
### Summary
1. add OnSuccessPluginTask for meta saving
2. add collector meta to save the latest collecting's meta such as `latestUpdated` and `CreatedDateAfter`
3. collect incrementally only when the `CreatedDateAfter` this time is equal to the latest time.
4. only incrementally collect remoteLinks from the issues matched `meta.latestUpdated<issue.updated` to reduce request number.
5. add nil checker for TransformationRules

Related #3805

### DIFF
--- a/plugins/core/plugin_task.go
+++ b/plugins/core/plugin_task.go
@@ -104,8 +104,14 @@ type PluginTask interface {
 	PrepareTaskData(taskCtx TaskContext, options map[string]interface{}) (interface{}, errors.Error)
 }
 
-// CloseablePluginTask Extends PluginTask, and invokes a Close method after all subtasks are done
+// CloseablePluginTask Extends PluginTask, and invokes a Close method after all subtasks are done or fail
 type CloseablePluginTask interface {
 	PluginTask
 	Close(taskCtx TaskContext) errors.Error
+}
+
+// OnSuccessPluginTask Extends PluginTask, and invokes a Close method after all subtasks are done
+type OnSuccessPluginTask interface {
+	PluginTask
+	OnSuccess(taskCtx TaskContext) errors.Error
 }
--- a/plugins/core/plugin_task.go
+++ b/plugins/core/plugin_task.go
@@ -109,9 +109,3 @@ type CloseablePluginTask interface {
 	PluginTask
 	Close(taskCtx TaskContext) errors.Error
 }
-
-// OnSuccessPluginTask Extends PluginTask, and invokes a Close method after all subtasks are done
-type OnSuccessPluginTask interface {
-	PluginTask
-	OnSuccess(taskCtx TaskContext) errors.Error
-}
--- a/plugins/helper/api_collector_with_state.go
+++ b/plugins/helper/api_collector_with_state.go
@@ -81,18 +81,14 @@ func (m *ApiCollectorStateManager) InitCollector(args ApiCollectorArgs) (err err
 
 // Execute the embedded collector and record execute state
 func (m ApiCollectorStateManager) Execute() errors.Error {
-	executeErr := m.ApiCollector.Execute()
+	err := m.ApiCollector.Execute()
+	if err != nil {
+		return err
+	}
 
 	db := m.Ctx.GetDal()
 	m.LatestState.LatestSuccessStart = &m.ExecuteStart
 	m.LatestState.StartFrom = m.StartFrom
-	saveErr := db.CreateOrUpdate(&m.LatestState)
-	if saveErr != nil {
-		if executeErr != nil {
-			return errors.Default.Combine([]error{executeErr, saveErr})
-		} else {
-			return errors.Default.Wrap(saveErr, "error on saving JiraLatestCollectorMeta")
-		}
-	}
-	return executeErr
+	err = db.CreateOrUpdate(&m.LatestState)
+	return err
 }
--- a/plugins/helper/api_collector_with_state.go
+++ b/plugins/helper/api_collector_with_state.go
@@ -25,15 +25,16 @@ import (
 	"time"
 )
 
+// ApiCollectorStateManager save collector state in framework table
 type ApiCollectorStateManager struct {
 	RawDataSubTaskArgs
 	*ApiCollector
-	db           dal.Dal
 	LatestState  models.CollectorLatestState
 	StartFrom    *time.Time
 	ExecuteStart time.Time
 }
 
+// NewApiCollectorWithState create a new ApiCollectorStateManager
 func NewApiCollectorWithState(args RawDataSubTaskArgs, startFrom *time.Time) (*ApiCollectorStateManager, errors.Error) {
 	db := args.Ctx.GetDal()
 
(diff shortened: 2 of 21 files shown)


plugins/jira/models/collector_meta.go
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+package models
+
+import (
+	"github.com/apache/incubator-devlake/models/common"
+	"time"
+)
+
+type JiraLatestCollectorMeta struct {
```
warren830, line 25: Can you add some comment to describe this struct? And what about change `JiraLatestCollectorMeta` to `JiraLatestCollectionRecord`

plugins/jira/tasks/epic_collector.go
```diff
@@ -53,12 +53,7 @@ func CollectEpics(taskCtx core.SubTaskContext) errors.Error {
 	if err != nil {
 		return err
 	}
-	since := data.Since
 	jql := "ORDER BY created ASC"
-	if since != nil {
-		// prepend a time range criteria if `since` was specified, either by user or from database
-		jql = fmt.Sprintf("updated >= '%s' %s", since.Format("2006/01/02 15:04"), jql)
-	}
 	collector, err := helper.NewApiCollector(helper.ApiCollectorArgs{
```
warren830, line 57: why do we delete since here and didn't add startFrom?

likyh: Because now `data.startFrom` means `created>%s`. But issue created after `data.startFrom` may use the epics created before `data.startFrom`.

In the old logic, Incremental is always false, So I think it is not correct here.

And epic can collect 100 per page and it's not too big to collect incrementally. So I delete them.

plugins/jira/tasks/issue_changelog_collector.go
```diff
@@ -60,12 +60,10 @@ func CollectIssueChangelogs(taskCtx core.SubTaskContext) errors.Error {
 		dal.Join("LEFT JOIN _tool_jira_issue_changelogs c ON (c.connection_id = i.connection_id AND c.issue_id = i.issue_id)"),
 		dal.Where("i.updated > i.created AND bi.connection_id = ?  AND bi.board_id = ? AND i.std_type != ? ", data.Options.ConnectionId, data.Options.BoardId, "Epic"),
 		dal.Groupby("i.issue_id, i.updated"),
-		dal.Having("i.updated > max(c.issue_updated) OR  (max(c.issue_updated) IS NULL AND COUNT(c.changelog_id) > 0)"),
 	}
-	// apply time range if any
-	since := data.Since
-	if since != nil {
-		clauses = append(clauses, dal.Where("i.updated > ?", *since))
+	incremental := data.Meta.StartFrom != nil && data.Meta.StartFrom.Equal(*data.StartFrom)
```
warren830, line 64: data.Meta.StartFrom.Equal(*data.StartFrom) 
what does this mean?

plugins/jira/models/collector_meta.go
likyh: ok

plugins/jira/tasks/issue_changelog_collector.go
likyh: If the StartFrom passed by options （`data.StartFrom`） equals to the latest collection's `StartFrom`, the collectors can work incrementally if they want. If not, it means `StartFrom` changed and all data need to be deleted.

plugins/jira/tasks/issue_collector.go
```diff
@@ -55,34 +55,34 @@ func CollectIssues(taskCtx core.SubTaskContext) errors.Error {
...
 	// IMPORTANT: we have to keep paginated data in a consistence order to avoid data-missing, if we sort issues by
 	//  `updated`, issue will be jumping between pages if it got updated during the collection process
+	startFrom := data.StartFrom
+	incremental := false
 	jql := "ORDER BY created ASC"
-	if since != nil {
+	if startFrom != nil {
 		// prepend a time range criteria if `since` was specified, either by user or from database
-		jql = fmt.Sprintf("updated >= '%v' %v", since.Format("2006/01/02 15:04"), jql)
+		jql = fmt.Sprintf("created >= '%v' %v", startFrom.Format("2006/01/02 15:04"), jql)
```
warren830, line 79: why `created` not `updated`?

plugins/jira/tasks/issue_collector.go
```diff
@@ -55,34 +55,34 @@ func CollectIssues(taskCtx core.SubTaskContext) errors.Error {
...
+	}
+	if latestUpdated.IssueId > 0 {
+		incremental = data.Meta.StartFrom != nil && data.Meta.StartFrom.Equal(*data.StartFrom)
+		if incremental {
+			jql = fmt.Sprintf("updated >= '%v' %v", latestUpdated.Updated.Format("2006/01/02 15:04"), jql)
```
warren830, line 84: Seems like we should consider startFrom OR latestUpdated? Not consider both?
For example, if startFrom == nil, we calculate latestUpdated and use it. Otherwise, we can use startFrom directly?

plugins/jira/jira.go
```diff
@@ -32,12 +32,12 @@ func main() {
 	boardId := cmd.Flags().Uint64P("board", "b", 0, "jira board id")
 	_ = cmd.MarkFlagRequired("connection")
 	_ = cmd.MarkFlagRequired("board")
-	since := cmd.Flags().StringP("since", "s", "", "collect data that are updated after specified time, ie 2006-05-06T07:08:09Z")
+	startFrom := cmd.Flags().StringP("start_from", "f", "", "collect data that are updated after specified time, ie 2006-05-06T07:08:09Z")
```
warren830, line 35: I think `s` for short hand is better?

plugins/jira/tasks/issue_collector.go
likyh: Emm, the old line is not deleted, and now it is in Line#84.

plugins/jira/jira.go
likyh: ok

plugins/jira/tasks/issue_collector.go
likyh: startFrom is used when the user just want to collect the issue created after `startFrom` (such as 2 months). latestUpdated is used for incremental collect.

If `startFrom` and `latestUpdated` set both, that means the plugin need to collect the issue created after `startFrom` and have updated after last collection(`latestUpdated`).

plugins/core/plugin_task.go
```diff
@@ -104,8 +104,14 @@ type PluginTask interface {
...
 
-// CloseablePluginTask Extends PluginTask, and invokes a Close method after all subtasks are done
+// CloseablePluginTask Extends PluginTask, and invokes a Close method after all subtasks are done or fail
 type CloseablePluginTask interface {
 	PluginTask
 	Close(taskCtx TaskContext) errors.Error
 }
+
+// OnSuccessPluginTask Extends PluginTask, and invokes a Close method after all subtasks are done
+type OnSuccessPluginTask interface {
```
klesh, line 114: As we discussed, we should store `startFrom` and `lastUpdated` on framework level `_devlake_collector_state` maybe?
In that case, we don't need this interface.

likyh:
For reivew: a new `ApiCollectorStateManager` to save collector in framework table `_devlake_collector_latest_state`

plugins/jira/models/collector_meta.go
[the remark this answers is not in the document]
likyh: saved in new table `_devlake_collector_latest_state`

plugins/helper/api_collector_with_state.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Execute the embedded collector and record execute state
+func (m ApiCollectorStateManager) Execute() errors.Error {
+	executeErr := m.ApiCollector.Execute()
+
+	db := m.Ctx.GetDal()
+	m.LatestState.LatestSuccessStart = &m.ExecuteStart
+	m.LatestState.StartFrom = m.StartFrom
+	saveErr := db.CreateOrUpdate(&m.LatestState)
```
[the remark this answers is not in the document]
likyh: fixed

klesh (approved):
LGTM

PR #3613, opened 2022-10-28
feat: defining PluginBlueprintV200 to support project

klesh:
# Summary

**Epic**: https://github.com/apache/incubator-devlake/issues/3468
**Section**: Implementation No. 3 - Blueprint Plan Generation Version v2.0.0
**Content**:
1. Defining `Scope` interface with methods `ScopeId()/ScopeName()/TableName()`, and update `repos` `boards` `cicd_scopes`(not yet exists, skipped) to implement the interface.
2. `PluginBlueprintV200` takes only `ScopeIds`(`connectionId`/`transformationRuleId` can be inferred) as input, and returns `PipelinePlan` and `[]Scope` to indicate the scopes belonging to the project.

### Does this close any open issues?
Part of #3468

PR #3826, opened 2022-12-01
feat: support bp v200 for jenkins

likyh:
### Summary
1. This is the implementation of Jenkins for data source plugins to support the PluginBlueprintV200 interface.
You can know more detail about the PluginBlueprintV200 interface on issue https://github.com/apache/incubator-devlake/pull/3613
2. fix some bug when put & patch scope

Related to #3613


mindlesscloud (approved):
LGTM

PR #3848, opened 2022-12-03
feat(config-ui): add global connections store

mintsweet:
### Summary

Add global **connections** store.

### Why add this store？

Connections are used in many places throughout the config UI. And every time it is used, a request will be sent to get this information. This results in meaningless requests and performance loss. Maintaining connections in multiple places also causes inconsistency.

### DIFF
--- a/config-ui/src/store/connections/use-connections.ts
+++ b/config-ui/src/store/connections/use-connections.ts
@@ -25,6 +25,7 @@ import { ConnectionStatusEnum } from './types'
 import { getConnection, testConnection } from './api'
 
 export const useConnections = (plugin?: string | string[]) => {
+  const [loading, setLoading] = useState(false)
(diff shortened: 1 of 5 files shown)


config-ui/src/store/connections/use-connections.ts
```diff
@@ -0,0 +1,147 @@
...
+  const allConnections = useMemo(
+    () =>
+      Plugins.filter((p) => p.type === 'integration').filter((p) => {
+        if (!plugin) return true
+        return Array.isArray(plugin) ? plugin.includes(p.id) : p.id === plugin
+      }),
+    [plugin]
+  )
+
+  const handleRefresh = useCallback(async () => {
```
likyh, line 39: Does `handleRefresh` need a flag `loading`?

likyh (approved):
LGTM

PR #3851, opened 2022-12-05
feat(config-ui): add new page projects

mintsweet:
### Summary
1. fix request params error when the method equal **get**.
2. add new component **toast2**.
3. add new utils **operator**.
4. support parameter **okLoading** in **dialog**.
5. support parameter **style** and **size** in **loading**.
6. add new page **projects**.

### Related Issues
#3468

#### init page:
screenshot-20221205-112719

#### has some test data:
screenshot-20221205-112626

#### add project dialog:
screenshot-20221205-112641


config-ui/src/components/toast2/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { Toaster, Position } from '@blueprintjs/core'
+
+export const Toast = Toaster.create({
```
likyh, line 21: What difference between Toast and Toast2? just position?

likyh (approved):
LGTM. these comment can be changed on next PR.

config-ui/src/components/toast2/index.tsx
mintsweet: Because the toast name is already taken, I need to export other content in this component.

PR #3856, opened 2022-12-05
feat: add scope for cicd in webhook

likyh:
### Summary
add scope for cicd in webhook. Just create a virtual scope to keep pace with other plugin and blueprint v200.

This is the implementation of Webhook for data source plugins to support the PluginBlueprintV200 interface.
You can know more detail about the PluginBlueprintV200 interface on issue https://github.com/apache/incubator-devlake/pull/3613

Related to https://github.com/apache/incubator-devlake/pull/3613

### DIFF
--- a/plugins/webhook/api/blueprint_v200.go
+++ b/plugins/webhook/api/blueprint_v200.go
@@ -1,0 +1,57 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


plugins/webhook/api/blueprint_v200.go
```diff
@@ -0,0 +1,57 @@
...
+
+func MakeDataSourcePipelinePlanV200(_ []core.SubTaskMeta, connectionId uint64, bpScopes []*core.BlueprintScopeV200) (core.PipelinePlan, []core.Scope, errors.Error) {
+	// get the connection info for url
+	connection := &models.WebhookConnection{}
+	err := connectionHelper.FirstById(connection, connectionId)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	plan := make(core.PipelinePlan, 0, len(bpScopes))
```
klesh, line 38: webhost doesn't have a `plan`, we can return nil for the `plan` directly.

plugins/webhook/api/blueprint_v200.go
```diff
@@ -0,0 +1,57 @@
...
+	scopes := make([]core.Scope, 0, len(bpScopes))
+	for _, bpScope := range bpScopes {
```
klesh, line 40: webhook doesn't have scopes, I believe that the `bpScopes` in this case should nil or empty and we should use the `connectionId` to generate a the `scopes`.

klesh (approved):
LGTM

PR #3889, opened 2022-12-08
docs: explain why we chose git2go rather than go-git

mindlesscloud:
### Summary
Explain why we chose git2go rather than go-git. The git2go is a golang binding of the C library libgit2.

No


warren830 (approved):
LGTM

PR #3892, opened 2022-12-09
fix(github): modify bp v200

warren830:
### Summary
Change the way to set options in plugin GitHub.
Before, we set all values to tasks.GithubOptions when making plan, right now, we just pass scopeId to options, and set other values when preparing plugin data

relate to #3468 

### Screenshots
we can see that the blueprint can be triggered successfully
image

bp V100:
image

### Other Information

### DIFF
--- a/plugins/github/api/blueprint_v200.go
+++ b/plugins/github/api/blueprint_v200.go
@@ -31,7 +31,6 @@ import (
 	"github.com/apache/incubator-devlake/plugins/github/tasks"
 	"github.com/apache/incubator-devlake/utils"
 	"github.com/go-playground/validator/v10"
-	"github.com/mitchellh/mapstructure"
 	"gorm.io/gorm"
 	"strings"
 	"time"
@@ -86,9 +85,12 @@ func makeDataSourcePipelinePlanV200(
 	apiClient helper.ApiClientGetter,
 ) (core.PipelinePlan, errors.Error) {
 	var err errors.Error
-	var stage core.PipelineStage
-	var repo *tasks.GithubApiRepo
+	var repoRes *tasks.GithubApiRepo
 	for i, bpScope := range bpScopes {
+		stage := plan[i]
+		if stage == nil {
+			stage = core.PipelineStage{}
+		}
 		githubRepo := &models.GithubRepo{}
 		// get repo from db
 		err = basicRes.GetDal().First(githubRepo, dal.Where(`connection_id = ? AND github_id = ?`, connection.ID, bpScope.Id))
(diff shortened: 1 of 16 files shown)


plugins/github/api/blueprint_v200.go
```diff
@@ -123,39 +125,37 @@ func makeDataSourcePipelinePlanV200(
...
 			}
-			stage, err = addGitex(bpScope.Entities, connection, repo, stage)
+			stage, err = addGitex(bpScope.Entities, connection, repoRes, stage)
 			if err != nil {
 				return nil, err
 			}
 		}
 		plan[i] = stage
 	}
+	fmt.Println(plan)
```
likyh, line 158: delete debug code

plugins/github/impl/impl.go
```diff
@@ -159,6 +161,32 @@ func (plugin Github) PrepareTaskData(taskCtx core.TaskContext, options map[strin
 	if err != nil {
 		return nil, errors.Default.Wrap(err, "unable to get github connection by the given connection ID")
 	}
+	var githubRepo models.GithubRepo
+	// for advanced mode which we already set value to onwer/repo
+	if op.Owner != "" && op.Repo != "" {
+		err = taskCtx.GetDal().First(&githubRepo, dal.Where(
+			"connection_id = ? AND name = ? AND owner_login = ",
```
likyh, line 168: it should be "owner_login = ?"

plugins/github_graphql/impl/impl.go
```diff
@@ -137,6 +133,41 @@ func (plugin GithubGraphql) PrepareTaskData(taskCtx core.TaskContext, options ma
 		return nil, errors.Default.Wrap(err, "unable to get github connection by the given connection ID: %v")
 	}
 
+	var githubRepo models.GithubRepo
+	// for advanced mode or others which we already set value to onwer/repo
+	if op.Owner != "" && op.Repo != "" {
+		err = taskCtx.GetDal().First(&githubRepo, dal.Where(
+			"connection_id = ? AND name = ? AND owner_login = ",
```
likyh, line 140: =?

warren830: fixed

plugins/github/impl/impl.go
warren830: fixed

plugins/github/api/blueprint_v200.go
```diff
@@ -86,9 +85,12 @@ func makeDataSourcePipelinePlanV200(
 	apiClient helper.ApiClientGetter,
 ) (core.PipelinePlan, errors.Error) {
 	var err errors.Error
-	var stage core.PipelineStage
-	var repo *tasks.GithubApiRepo
+	var repoRes *tasks.GithubApiRepo
 	for i, bpScope := range bpScopes {
+		stage := plan[i]
```
likyh, line 90: plan and bpScopes have the same order and count?

plugins/github_graphql/impl/impl.go
```diff
@@ -178,6 +178,49 @@ func (plugin GithubGraphql) PrepareTaskData(taskCtx core.TaskContext, options ma
 	}, nil
 }
 
+func EnrichOptions(taskCtx core.TaskContext,
```
likyh, line 181: Defined in github_graphql?

plugins/github_graphql/impl/impl.go
```diff
@@ -178,6 +178,49 @@ func (plugin GithubGraphql) PrepareTaskData(taskCtx core.TaskContext, options ma
...
+		err := taskCtx.GetDal().First(&githubRepo, dal.Where(
+			"connection_id = ? AND name = ? AND owner_login = ?",
+			op.ConnectionId, op.Repo, op.Owner))
+		if err != nil {
+			return nil, errors.Default.Wrap(err, fmt.Sprintf("fail to find repo %s/%s", op.Owner, op.Repo))
+		}
+		op.TransformationRuleId = githubRepo.TransformationRuleId
+	}
+	// for bp v200 which we only set ScopeId for options
+	if &githubRepo == nil && op.ScopeId != "" {
```
likyh, line 198: It looks like `&githubRepo == nil` always true.

warren830: > It looks like `&githubRepo == nil` always true.

yes, already fixed

PR #3901, opened 2022-12-09
refactor: remove scope collector and extractor

mindlesscloud:
### Summary
After the scope refactoring proposed in #3468, the scope would be sent from the config-UI rather than collected by the plugin itself. In this PR, we removed the following subtasks:
- Jira board collector and extractor 
- GitHub repo collector and extractor
- Gitlab project collector and extractor
- Jenkins job collector and extractor

No, but this should be viewed as part of #3468


klesh (approved):
LGTM

PR #3907, opened 2022-12-12
refactor(config-ui): add new component logo

mintsweet:
### Summary
Add new component **logo**.

### Why do you do this?

Because this is just a pre-work, I found that the same picture and style were referenced when I refactored the **error handling** and **offline**, so I encapsulated the content of this part first so that I could refactor the **error handling** and **offline** later.


likyh (approved):
LGTM

PR #3924, opened 2022-12-13
fix(config-ui): connection store not injected into the context

mintsweet:
### Summary
The connections store is not injected into the context, and it will trigger a lot of extra requests.

### DIFF
--- a/config-ui/src/pages/blueprint/create/hooks/use-blueprint-value.ts
+++ b/config-ui/src/pages/blueprint/create/hooks/use-blueprint-value.ts
@@ -19,6 +19,7 @@
 import { useState, useMemo } from 'react'
 import { useHistory } from 'react-router-dom'
 
+import { useStore, ConnectionStatusEnum } from '@/store'
 import { operator } from '@/utils'
 
 import type { BPConnectionItemType, BPScopeItemType } from '../types'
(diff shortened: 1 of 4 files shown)


config-ui/src/pages/blueprint/create/hooks/use-blueprint-value.ts
```diff
@@ -49,6 +50,8 @@ export const useBlueprintValue = ({ from, projectName }: Props) => {
 
   const history = useHistory()
 
+  const store = useStore()
```
likyh, line 53: const { connections } = useStore()

config-ui/src/pages/blueprint/create/step-one/index.tsx
```diff
@@ -40,7 +40,7 @@ import * as S from './styled'
 export const StepOne = () => {
   const [isOpen, setIsOpen] = useState(false)
 
-  const connectionsStore = useConnections()
+  const connectionsStore = useStore()
```
likyh, line 43: const { onTestConnections } = useStore()

config-ui/src/pages/blueprint/create/hooks/use-blueprint-value.ts
mintsweet: Because there is content with the same name here, it cannot be deconstructed, I will modify the problem with the same name.

config-ui/src/pages/blueprint/create/step-one/index.tsx
mintsweet: I will use ** connectionsStore.connections**.

likyh (approved):
I approve because the Naming issue will refactor in the next PR.

PR #3931, opened 2022-12-14
Jenkins add GetAllJobs

mappjzc:
### Summary
Since we need compatibility with makePipelinePlanV100
So we need Jenkins to get all the jobs when makePipelinePlanV100
So we need a method to get all Jobs

Now create the method GetAllJobs and the corresponding unit test.

Closes xx

### Screenshots
image

### Other Information

### DIFF
--- a/plugins/jenkins/api/jobs.go
+++ b/plugins/jenkins/api/jobs.go
@@ -30,7 +30,7 @@ import (
 )
 
 // request all jobs
-func GetAllJobs(apiClient helper.ApiClientGetter, path string, pageSize int, callback func(job *models.Job, isPath bool) errors.Error) errors.Error {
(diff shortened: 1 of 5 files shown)


plugins/jenkins/api/jobs.go
```diff
@@ -0,0 +1,92 @@
...
+
+		for _, rawJobs := range data.Jobs {
+			job := &models.Job{}
+			err1 := json.Unmarshal(rawJobs, job)
+			if err1 != nil {
+				return errors.Convert(err1)
+			}
+
+			job.Path = path
+			job.FullName = path + job.Name
```
likyh, line 68: FullName should read from response.

likyh (approved):

PR #3997, opened 2022-12-21
fix: fix gitlab PUT scopes router

mindlesscloud:
### Summary
move the handler of PUT scopes from `connections/:connectionId/scopes/:projectId` to `connections/:connectionId/scopes`


warren830 (approved):
LGTM

PR #4009, opened 2022-12-21
feat: add time filter for gitlab jobs

likyh:
### Summary
add time filter for gitlab jobs.

GitLab jobs do not support the time filter and time sorting. It can only sort by id desc (not strictly). https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines

So filter by stopping when job's create time before than options.

gitlab's pipeline can sort by time asc, so I haven't add this. Sort asc can avoid new rows qustions.

### snapshots
The result for only collect the rows created after `2022-12-01T00:00:00Z`
image

### DIFF
--- a/plugins/gitlab/tasks/shared.go
+++ b/plugins/gitlab/tasks/shared.go
@@ -94,7 +94,7 @@ func GetRawMessageCreatedAtAfter(createDateAfter *time.Time) func(res *http.Resp
 				return nil, err
 			}
 			if createDateAfter == nil || createDateAfter.Before(apiModel.CreatedAt.ToTime()) {
-				// only finish when all item before `createDateAfter`
+				// only finish when all items created before `createDateAfter`
 				// because gitlab's order may not strict enough
 				isFinish = false
 				filterRawMessages = append(filterRawMessages, rawMessage)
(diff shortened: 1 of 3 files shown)


plugins/gitlab/tasks/shared.go
```diff
@@ -74,6 +75,38 @@ func GetRawMessageFromResponse(res *http.Response) ([]json.RawMessage, errors.Er
...
+		isFinish := true
+		filterRawMessages := []json.RawMessage{}
+		for _, rawMessage := range rawMessages {
+			apiModel := &ApiModel{}
+			err = errors.Convert(json.Unmarshal(rawMessage, apiModel))
+			if err != nil {
+				return nil, err
+			}
+			if createDateAfter == nil || createDateAfter.Before(apiModel.CreatedAt.ToTime()) {
+				// only finish when all item before `createDateAfter`
```
klesh, line 97: The comment is not correct. It should be "all **items created after**"

likyh: ok, only finish when all items are created before `createDateAfter`.
`createDateAfter` means only accepting items created after it, So if all items reject, collect need finish.

klesh (approved):
LGTM

PR #4012, opened 2022-12-22
fix(github): minor bugs

warren830:
### Summary
1. Use uppercase for all status and conclusion for cicd jobs in github
2. update e2e

relate to #3468 

### Screenshots
image
### Other Information


likyh (approved):
It seems like rest return lowercase and graphql return uppercase

PR #4013, opened 2022-12-22
fix(config-ui): some bugs

mintsweet:
### Summary
1. fix cannot use advanced mode in project.
2. adjust the style for bp.
3. fix the push link error.
4. fix the token not check if one input skip.
5. fix cannot display correctly when bp has webhook.
6.  fix cannot display correctly when bp is advanced mode.
7. adjust the style for project detail.

### DIFF
--- a/config-ui/src/pages/blueprint/detail/panel/configuration.tsx
+++ b/config-ui/src/pages/blueprint/detail/panel/configuration.tsx
@@ -22,7 +22,7 @@ import dayjs from 'dayjs'
 
 import { Table, ColumnType } from '@/components'
 import { getCron } from '@/config'
-import { DataScopeList } from '@/plugins'
+import { PluginConfig, PluginType, DataScopeList, Plugins } from '@/plugins'
 
 import { ModeEnum } from '../../types'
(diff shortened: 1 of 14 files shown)


config-ui/src/pages/blueprint/detail/panel/configuration.tsx
```diff
@@ -66,6 +64,25 @@ export const Configuration = ({
...
+          return {
+            icon: plugin.icon,
+            name: plugin.name,
+            connectionId: cs.connectionId,
+            entities: cs.scopes[0].entities,
+            plugin: cs.plugin,
+            scopeIds: cs.scopes.map((sc: any) => sc.id)
+          }
+        })
+        .filter(Boolean),
```
likyh, line 82: It seems like never be null

likyh (approved):

config-ui/src/pages/blueprint/detail/panel/configuration.tsx
mintsweet: Yes

PR #4026, opened 2022-12-26
fix(github gitlab): modify dora and refdiff

warren830:
### Summary
For bp v100, refdiff needs repoId, so in this pr, we added repoId to refdiff options
Again, we now run `enrichTaskEnv` in relevant plugins, but we cannot delete this when making pipeline plan v100 because we will append dora subtasks to the plan in `GeneratePlanJsonV100`. So, we keep `enrichTaskEnv` in plan.stage, but will do nothing about it.

### Issues
relate to #3468 

### Screenshots
image

image

### Other Information


plugins/dora/tasks/cicd_task_env_enricher.go
```diff
@@ -37,49 +33,5 @@ var EnrichTaskEnvMeta = core.SubTaskMeta{
...
-			}
-			return nil, nil
-		},
-	})
-	if err != nil {
-		return err
-	}
-
-	return converter.Execute()
+	return nil
```
klesh, line 36: Please add a comment to mention that the enrichment was moved to data source plugins

klesh (approved):
LGTM

PR #4027, opened 2022-12-26
PoC Jira Server PAT support

klesh:
### Summary
Currently, we support only `Basic Authentication` for Jira Cloud/Server, it worked fine for most of our users.
However, some might run into a situation the their Jira Server support PAT only for security reason which makes it impossible for devlake to collect data.

This is a PoC to verify if it is possible to collect data by using Jira PAT in a hacky way.

### DIFF
--- a/plugins/jira/tasks/api_client.go
+++ b/plugins/jira/tasks/api_client.go
@@ -19,8 +19,9 @@ package tasks
 
 import (
 	"fmt"
-	"github.com/apache/incubator-devlake/errors"
 	"net/http"
+
+	"github.com/apache/incubator-devlake/errors"
 
 	"github.com/apache/incubator-devlake/plugins/core"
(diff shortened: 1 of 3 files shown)


plugins/jira/tasks/api_client.go
```diff
@@ -32,6 +33,9 @@ func NewJiraApiClient(taskCtx core.TaskContext, connection *models.JiraConnectio
 	headers := map[string]string{
 		"Authorization": fmt.Sprintf("Basic %v", connection.GetEncodedToken()),
 	}
+	if connection.Username == "_BEARER_TOKEN_" {
```
keon94, line 36: maybe put line 33 in an else block, so we're not double-writing the same key to the map.

klesh:
@keon94 This is a temporary fix, not meant to be merged

PR #4054, opened 2022-12-28
refactor: improve slice size check for gitlab

mindlesscloud:
### Summary
The following snippet is taken from the codebase. It looks like it would result in the `index out of range error`, but the function `FindStringSubmatch` ensures this will not happen. In this PR, we improved the readability by changing the expression inside the `if` statement from `len(groups) > 0` to `len(groups) > 1`
```go
groups := labelTypeRegex.FindStringSubmatch(label)
	if len(groups) > 0 {
		gitlabMergeRequest.Type = groups[1]
	}
```


warren830 (approved):
LGTM

PR #4089, opened 2023-01-03
gitextractor should delete old data

### LINKED ISSUE #4024: [Bug][gitextractor] Old data (commit_files) does not get deleted and numbers are off
### Search before asking

### What happened

from pr #4022, each commits has 500+ files change which is not correct..
I purged related commit_files and got the following numbers which are closer to accurate according to github commit page:
image
image
image

However, the numbers are still off..

The numbers should match to whatever shows in github-commit-page

### How to reproduce

collect the repo and observe by the following SQL:
select cf.commit_sha, count(*) 
from commit_files cf 
left join repo_commits rc on rc.commit_sha  = cf.commit_sha 
where repo_id = 'github:GithubRepo:1:369331408'
group by cf.commit_sha;

### Anything else

_No response_

### Version

main

### Code of Conduct

klesh:
### Summary

The phenomenon described in #4024 was caused by the fact that `gitextractor` failed to delete previously collected data, it was due to the fact `_raw_data_*` fields were never populated so they would never be deleted.

This PR fixes the problem

Closes #4024

### DIFF
--- a/plugins/gitextractor/impl/impl.go
+++ b/plugins/gitextractor/impl/impl.go
@@ -64,7 +64,7 @@ func (plugin GitExtractor) PrepareTaskData(taskCtx core.TaskContext, options map
 	if err := op.Valid(); err != nil {
 		return nil, err
 	}
-	storage := store.NewDatabase(taskCtx, op.Url)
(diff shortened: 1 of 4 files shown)


likyh (approved):
LGTM

PR #4090, opened 2023-01-03
fix: fix the canceling does not work when SkipOnFail is set to true

mindlesscloud:
### Summary
Fix the bug that canceling does not work when `SkipOnFail` is set to true
- make sure every task returns the error `context.Canceled` when it is canceled
- if the error is`context.Canceled`, return it despite the value of `SkipOnFail`

### Screenshots
image

### DIFF
--- a/plugins/helper/api_collector.go
+++ b/plugins/helper/api_collector.go
@@ -168,7 +168,8 @@ func (collector *ApiCollector) Execute() errors.Error {
 					break
 				}
 			}
-			input, err := iterator.Fetch()
+			var input interface{}
+			input, err = iterator.Fetch()
 			if err != nil {
 				break
 			}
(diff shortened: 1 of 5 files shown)


plugins/helper/api_collector.go
```diff
@@ -168,7 +168,8 @@ func (collector *ApiCollector) Execute() errors.Error {
 					break
 				}
 			}
-			input, err := iterator.Fetch()
+			var input interface{}
+			input, err = iterator.Fetch()
```
warren830, line 172: What does this fix?

warren830: I know there is some issue which I cannot regenerate in jenkins, does this relate to it?

mindlesscloud: this snippet is sitting inside a for loop, the statement `input, err := iterator.Fetch()` would declare a new local variable `err` which shadows the original `err`.

warren830 (approved):
LGTM

PR #4095, opened 2023-01-04
fix(config-ui): adjust the rateLimitPerHour default value

### LINKED ISSUE #4068: [Bug][Plugin] Display the actual default API rate limit instead of 0 on the data connection page
### Search before asking

### What happened

When enabling the Rate Limit configuration in the connection page of any data source, the default value shows 0; while it's not.

image

- [ ] Change the field name from `Rate Limit (per hour)` to `Fixed Rate Limit (per hour)`
- [ ] When the state of the toggle is `disabled`, show text `Use dynamic rate limit` on the right side
- [ ] When the state of the toggle is `enabled`, show default value in the input for each data source as below:
    - [ ] GitHub: default rate limit 4500 (Maximum 5000 per hour, see [details](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting))
    - [ ] GitLab: default rate limit 5,000 (Maximum 120,000 per hour, but many users encountered 429, see [details](https://docs.gitlab.com/ee/user/gitlab_com/index.html#gitlabcom-specific-rate-limits))
    - [ ] Jira: default rate limit 3,000 (Jira cloud uses dynamic rate limit, but someone tested out that currently, if calls from same IP address exceed 600 requests per 60 seconds, calls are blocked for 60 seconds. See [details](https://help.jiraalign.com/hc/en-us/articles/360045371954-Getting-started-with-the-REST-API-2-0))
    - [ ] Jenkins: can't find any doc, let's use the fallback option 10,000
    - [ ] TAPD: default rate limit 3,000 (max: 60req/1min, see [detail](https://o.tapd.cn/document/api-doc/API%E6%96%87%E6%A1%A3/API%E9%94%99%E8%AF%AF%E7%A0%81_cloud.html))
    - [ ] Zentao: can't find any doc, let's use the fallback option 10,000 

### How to reproduce

1. Go to the `connection` page
2. Click `add a connection`
3. Enable 'Rate Limit'

### Anything else

_No response_

### Version

v0.14.2

### Code of Conduct

mintsweet:
### Summary
adjust the **rateLimitPerHour** default value.

Closes #4068

### DIFF
--- a/config-ui/src/pages/connection/form/components/rate-limit/index.tsx
+++ b/config-ui/src/pages/connection/form/components/rate-limit/index.tsx
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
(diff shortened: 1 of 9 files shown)


likyh:
I think the default value of ratelimit does not mean this. It should be 0 by default, that is, the button enable. The default value is used only when the button is off. By the way, there's no hint. @Startrekzky

mintsweet:
@likyh updated.

likyh (approved):

PR #3921, opened 2022-12-13
fix: dbt zombie processes

abeizn:
### Summary
make sure the cmd.Wait() can be executed.

### Does this close any open issues?
Closes #3919 

### Screenshots
before:
Execute three times in a row, process screenshot：

image

after:
Execute three times in a row, process screenshot：
image

### Other Information
Any other information that is important to this PR.

PR #4101, opened 2023-01-04
fix: maybe can fix dbt zombie processes

abeizn:
### Summary
i can not recurrent this bug, but it occur in production env.

same to https://github.com/go-gitea/git/blob/master/commit_info.go#L237.

related to #3921

### Screenshots

### Other Information


mindlesscloud (approved):
Let's take a chance

PR #4105, opened 2023-01-04
fix: fix gitlab missing transformation rules

mindlesscloud:
### Summary
Fix the gitlab plugin missing transformation rules
- set `transformationRuleId` to the `options` in the function `makePipelinePlanV200`
- find the transformation rule in the database by the `transformationRuleId`
- set the transformation rule to task data


plugins/gitlab/api/blueprint_v200.go
```diff
@@ -132,6 +132,7 @@ func makePipelinePlanV200(subtaskMetas []core.SubTaskMeta, scopes []*core.Bluepr
 		options := make(map[string]interface{})
 		options["connectionId"] = connection.ID
 		options["projectId"] = intScopeId
+		options["transformationRuleId"] = transformationRules.ID
```
abeizn, line 135: need to add it at blueprint?

mindlesscloud: After a discussion with @warren830 , we think it's OK to set `transformationRuleId` to options within the function `makePipelinePlanV200`

abeizn (approved):
LGTM

Issue #3607, opened 2022-10-28
[Feature][config-ui] display version information


asterix314:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Previously, the version information was displayed at the page's lower-left corner, like this:

![image](https://user-images.githubusercontent.com/10952409/198491962-f073f086-127e-4a5f-b745-2a1c883c5b9d.png)

But not anymore in v0.14. The version info is either removed or hard to find. Please make it reappear. It is handy for issue/bug reporting among other things.

### Use case

Devlake version easily accessible in config-ui page.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

abeizn:
@asterix314  thanks for your contribution.

Startrekzky:
![image](https://user-images.githubusercontent.com/14050754/205082460-d7c29aa4-a02f-4e32-9782-38ae26d207b2.png)

PR #4108, opened 2023-01-05
fix: fix the style of version

likyh:
### Summary
fix the style of version

closes https://github.com/apache/incubator-devlake/issues/3607

### Screenshots
image

### DIFF
--- a/config-ui/src/layouts/base/base.tsx
+++ b/config-ui/src/layouts/base/base.tsx
@@ -76,8 +76,8 @@ export const BaseLayout = ({ children }: Props) => {
           ))}
         </Menu>
         <div className="copyright">
-          <div>{version}</div>
           <span>Apache 2.0 License</span>
+          <div>{version}</div>
         </div>
       </S.Sider>
       <S.Inner>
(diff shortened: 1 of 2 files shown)


Startrekzky (approved):
The design is acceptable.

config-ui/src/layouts/base/styled.ts
```diff
@@ -73,6 +73,12 @@ export const Sider = styled.div`
     left: 0;
     text-align: center;
     color: rgba(124, 124, 124, 0.7);
+    padding: 0 20px;
+    div {
```
mintsweet, line 77: Don't use div directly to add a class.

config-ui/src/layouts/base/base.tsx
```diff
@@ -76,8 +76,8 @@ export const BaseLayout = ({ children }: Props) => {
           ))}
         </Menu>
         <div className="copyright">
-          <div>{version}</div>
           <span>Apache 2.0 License</span>
+          <div>{version}</div>
```
mintsweet, line 80: You can use `span` and add a class.

mintsweet (approved):
LGTM.

PR #4118, opened 2023-01-05
fix: maybe can fix-3 dbt zombie processes

abeizn:
### Summary
i can not recurrent this bug, but it occur in production env.
quick-fix and i will test on https://env.merico.cn/envList 

related to #3921

### Screenshots

### Other Information


mindlesscloud (approved):
LGTM

PR #4132, opened 2023-01-06
fix: dbt zombie process, set dbt default threads=1

abeizn:
### Summary
The dbt threads default value changed to 4, but it can not release when the project run ended,
so we need set dbt default threads=1.

https://docs.getdbt.com/docs/dbt-versions/release-notes/dec-2022/default-thread-value

related to #3921

### Screenshots

### Other Information


mindlesscloud (approved):
LGTM

PR #4137, opened 2023-01-06
fix: Bitbucket git-extractor repo cloning fix

### LINKED ISSUE #4138: [Bug][Bitbucket] Bitbucket plugin passes wrong URL to git extractor when making pipeline
### Search before asking

### What happened

A user reported in the Slack that on v0.14.2, when invoking Bitbucket plugin (beta) through API following steps in [doc](https://devlake.apache.org/docs/next/Plugins/bitbucket), the git extractor task would fail to clone repos whose owners aren't the user himself. 

Upon examination of our source code, it seems we incorrectly passed `Op.Owner` for authentication instead of using the username from the Bitbucket connection. This is tricky because when `Op.Owner` and the connection's username are the same people, then this logic would work and thus hide the issue. 

Screen Shot 2023-01-06 at 11 04 51 AM

I expect a user A can successfully clone a user B's Bitbucket private repos as long as user A has access to those repos.

### How to reproduce

Please see above.

### Anything else

This bug needs to be fixed on main and the fix needs to be cherry-picked back to v0.14.x.

### Version

main

### Code of Conduct

keon94:
Fixes an issue reported by a Slack user regarding using the git-extractor plugin on Bitbucket. The credential logic for the extractor was incorrect, leading to failed repo clones. This PR fixes that. It'll be ported back to v0.14.8 as well.

Closes #4138

### DIFF
--- a/plugins/bitbucket/api/blueprint.go
+++ b/plugins/bitbucket/api/blueprint.go
@@ -142,7 +142,7 @@ func makePipelinePlan(subtaskMetas []core.SubTaskMeta, scope []*core.BlueprintSc
 			if err != nil {
 				return nil, err
 			}
-			cloneUrl.User = url.UserPassword(op.Owner, connection.Password)
+			cloneUrl.User = url.UserPassword(connection.Username, connection.Password)
 			stage = append(stage, &core.PipelineTask{
 				Plugin: "gitextractor",
 				Options: map[string]interface{}{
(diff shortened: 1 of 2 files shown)


hezyin (approved):
LGTM

hezyin:
The fix has been verified on the main branch. Merging it now.

PR #4174, opened 2023-01-10
fix: fix advance bp cannot edit again

likyh:
### Summary
fix advance bp cannot edit again.

Create request:
image

Edit request:
image

If `setting` haven't been post, the bp display API will got error.

Now I fix it.


likyh:
duplicate with https://github.com/apache/incubator-devlake/pull/4176

PR #4175, opened 2023-01-10
fix: limit the max rate limit = 200 for gitlab

### LINKED ISSUE #3192: [Bug][Gitlab] Pipeline collector report 429
### Search before asking

### What happened

Pipeline collector report 429 for the requests that would get empty response,
This seems like a bug from gitlab

Should not report that error

### How to reproduce

collect gitlab pipeline

### Anything else

_No response_

### Version

main

### Code of Conduct

likyh:
### Summary
limit the max rate limit = 200 for GitLab. Because the pipeline seems to have can only support about 250 requests per minute. Although Gitlab allow 2000 pre minute, I find that Devlake now can only use about 300 per minute by rateLimit header for GitLab. So I think we can set it to 200 at the most.

Closes #3192

### DIFF
--- a/plugins/gitlab/tasks/api_client.go
+++ b/plugins/gitlab/tasks/api_client.go
@@ -19,14 +19,13 @@ package tasks
 
 import (
 	"fmt"
-	"github.com/apache/incubator-devlake/errors"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


klesh (approved):
LGTM

PR #4176, opened 2023-01-10
fix: blueprint api would panic if len(settings) is zero

klesh:
### Summary
Before:

image

After:
image

### Screenshots


warren830 (approved):
LGTM

PR #4183, opened 2023-01-10
fix(framework): dont wrap contextCanceled and use this to calculate s…

warren830:
### Summary
1. If subtask.error is context.Canceled, we don't wrap it and just return 
image

2. Then if taskError is context.Canceled, we assign its status `CANCEL`
image

3.  Calculate pipeline status with `Canceled`
image


warren830:
this mr is no longer needed

PR #4187, opened 2023-01-10
Add Subnet mask to docker-compose

### LINKED ISSUE #4186: [Refactor][docker-compose] Add subnet mask to docker-compose
## What and why to refactor
I am trying to add a subnet mask for the docker containers.

```
networks:
  devlake_network:
    driver: bridge
    ipam:
     config:
       - subnet: 10.6.1.0/24
```

## Related issues
No

## Additional context
No

shubham-cmyk:
### Summary
This PR would add a different network for the container because sometimes the ports are already bound on the localhost which could be fixed and add more flexibility to user to chose a network 

Closes #4186 

### Screenshots
Screenshot from 2023-01-11 02-02-13
Screenshot from 2023-01-11 02-02-30

### Other Information
Container Working good

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,8 +20,11 @@ services:
     volumes:
       - mysql-storage:/var/lib/mysql
     restart: always
-    ports:
-      - 127.0.0.1:3306:3306
+    # ports:
+    #   - 127.0.0.1:3306:3306
+    networks:
+      devlake_network: 
+        ipv4_address: 10.6.1.13
     environment:
       MYSQL_ROOT_PASSWORD: admin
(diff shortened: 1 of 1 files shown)


shubham-cmyk:
@keon94  Can you look at it.

klesh:
@shubham-cmyk Hi, thanks for the hard work.
But I don't understand why port conflict on the host machine would have anything to do with the subnet between containers.
We use `ports` sections when we need to access the `container` from our HOST machine, and we should avoid using the same `port` for different `containers`. Deleting all `ports` would certainly avoid port conflict, but how can we access those ports from a remote computer?

PR #4206, opened 2023-01-12
doc: Replace supported data-sources in Readme with a hyperlink to website

### LINKED ISSUE #3095: [Doc][Documentation] Use hyperlinks to link unlinked Readme.md content to Website
## Documentation Scope
Readme.md

## Describe the Change
The Readme contains duplicated content from the [website](https://github.com/apache/incubator-devlake-website) (Introduction, Datasources, Create Blueprint, etc). This makes managing it cumbersome (we have to track both places). I suggest we simply link these sections to the website as well, like we have with "Getting Started", to make the Readme content minimal. If we want the Readme to contain actual text content we should look into embedding doc pages from the website rather than duplicating content.

keon94:
### Summary
Currently the supported-datasources table is duplicated in both the Readme and the Website docs. This PR will hyperlink the Readme to the website so that we only have a single source. This will ease the maintenance burden.

Closes #3095

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,18 +34,8 @@ Apache DevLake is designed for developer teams looking to make better sense of t
 
 ## 💪 Supported Data Sources
 
-| Data Source                   | Domain(s)                                                  | Supported Versions                                          | Config UI Availability | Triggered Plugins           | Collection Mode       |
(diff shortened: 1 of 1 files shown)


yumengwang03 (approved):
LGTM

PR #4236, opened 2023-01-18
feat: Docker-compose for data-sources

keon94:
### Summary
Creates a new docker compose file to support running supported datasources locally. For starters, I am adding Jenkins, Jira and Bitbucket.

Relates to https://github.com/apache/incubator-devlake-website/issues/385
Potentially closes it too depending on feedback.

Tested locally and all three worked.
image
image
image

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -34,6 +34,7 @@ header:
     - '**/env.example'
     - '**/*.csv'
     - '**/*.json'
+    - '**/*.sql'
     - '**/*.svg'
     - '**/*.png'
     - '.editorconfig'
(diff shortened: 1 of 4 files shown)


hezyin:
@keon94 Two suggestions:

1. Could you please add documentation to "Developer Setup" on what this new docker-compose is for and how to use it?
2. Ideally, we'd like to be able to fire up data source instances that are ready for testing purposes. Right now, they're empty instances that still require creating test data manually. Do you think you could look into that?

hezyin:
@keon94 I'm gonna merge this one now and leave the suggestions/improvements for future work.

PR #4260, opened 2023-01-29
Multiple Authorization Methods Support Mechanism (JIRA supports PAT/BasicAuth)

### LINKED ISSUE #4211: [Feature][Jira] Support bearer authentication
### Search before asking

### Use case

Right now, the Jira plugin only supports basic authentication. However, some users report their Jira server has basic authentication turned off and they don't have admin access to their Jira server. We should consider support bearer authentication method as an alternative.

### Description

_No response_

### Related issues

_No response_

### Code of Conduct

klesh:
### Summary
This PR is a more robust/reasonable/elegant/comprehensive solution for #4027, #4212.

The Mechanism offers the following aid for connections that require Single or Multiple Authorizations

** For Connections Support Single Authorization Method **, define the connection to look like:

```golang
// GitlabConn holds the essential information to connect to the Gitlab API
type GitlabConn struct {
	helper.RestConnection `mapstructure:",squash"`
	helper.AccessToken    `mapstructure:",squash"`
}

// This object conforms to what the frontend currently sends.
// GitlabConnection holds GitlabConn plus ID/Name for database storage
type GitlabConnection struct {
	helper.BaseConnection `mapstructure:",squash"`
	GitlabConn            `mapstructure:",squash"`
}
```

** For Connections Support Multiple Authorization Methods **, define the connection to look like:
```golang
// JiraConn holds the essential information to connect to the Jira API
type JiraConn struct {
	helper.RestConnection `mapstructure:",squash"`
	helper.MultiAuth      `mapstructure:",squash"`
	helper.BasicAuth      `mapstructure:",squash"`
	helper.AccessToken    `mapstructure:",squash"`
}

// SetupAuthentication implements the `IAuthentication` interface by delegating
// the actual logic to the `MultiAuth` struct to help us write less code
func (jc JiraConn) SetupAuthentication(req *http.Request) errors.Error {
	return jc.MultiAuth.SetupAuthenticationForConnection(&jc, req)
}

// JiraConnection holds JiraConn plus ID/Name for database storage
type JiraConnection struct {
	helper.BaseConnection `mapstructure:",squash"`
	JiraConn              `mapstructure:",squash"`
}
```

** Test Connection implementation for both parties be like **:
```golang
	// decode
	var err errors.Error
	var connection models.GitlabConn
	if err = api.Decode(input.Body, &connection, vld); err != nil {
		return nil, err
	}
	// test connection
	apiClient, err := api.NewApiClientFromConnection(context.TODO(), basicRes, connection)
	if err != nil {
		return nil, errors.Convert(err)
	}

// TEST the connection with apiClient
```

Closes #4027 
Closes #4212 
Closes #4211


warren830:
According to backend/plugins/feishu/models/migrationscripts/20220714_add_init_tables.go, we also need to modify this script in jira

backend/plugins/sonarqube/models/connection.go
```diff
@@ -21,6 +21,7 @@ import helper "github.com/apache/incubator-devlake/helpers/pluginhelper/api"
 
 // This object conforms to what the frontend currently sends.
 type SonarqubeConnection struct {
```
warren830, line 23: According to NewApiClientFromConnection, for sonarqube connection, we have to save "TOKEN:" as token, because sonarqube's Authorization is different from other plugins

klesh: Do you mean the Header value should be `Token xxx`?

warren830 (approved):
LGTM

warren830:
Seems like we need to do more work in 20230129_add_multi_auth.go for jira, otherwise, it will report error:
image

klesh:
> Seems like we also need to modify all existed connections in 20230129_add_multi_auth.go for jira, otherwise, it will report error: image

Good catch, fixed.

warren830 (approved):
LGTM

PR #4276, opened 2023-01-30
fix(sonarqube): modify connection to conform pr4260

warren830:
### Summary
According pr #4260, we modify sonarqube's connection to conform the new pattern. Then we can use multi-auth very easily in future

Relate to #4260 

### Screenshots

### Other Information


klesh (approved):
LGTM

PR #4292, opened 2023-02-01
[Bitbucket] Pagination params and merged date on PRs

### LINKED ISSUE #4285: [Bug][Bitbucket] Error when running `collectApiPullRequestCommits`
### Search before asking

### What happened

When I run `collectApiPullRequestCommits` with Bitbucket plugin the API that throws an error (Bad Request) when start to fetching the PR API.
image
image

Fetching on Bitbucket need to run without throwing an error.

### How to reproduce

Configure a Bitbucket connection and run a pipeline with any repo.

### Anything else

I think it's a problem with params that Bitbucket API expect to do pagination. 

### Version

main

### Code of Conduct

jrflga:
### Summary
What does this PR do?
- Fix the pagination params on Bitbucket API and add a feature to get the PR merge datetime (in order to fix the Lead Time feature on DORA for Bitbucket plugin).

Closes #4285

### Screenshots
After these modifications, DORA metrics dashboard shows these metrics.
Imagem 01-02-2023 às 16 57

### DIFF
--- a/backend/plugins/bitbucket/tasks/api_common.go
+++ b/backend/plugins/bitbucket/tasks/api_common.go
@@ -65,8 +65,8 @@ func CreateRawDataSubTaskArgs(taskCtx plugin.SubTaskContext, Table string) (*api
 func GetQuery(reqData *api.RequestData) (url.Values, errors.Error) {
 	query := url.Values{}
 	query.Set("state", "all")
-	query.Set("page", fmt.Sprintf("%v", reqData.Pager.Page))
-	query.Set("pagelen", fmt.Sprintf("%v", reqData.Pager.Size))
+	query.Set("start", fmt.Sprintf("%v", reqData.Pager.Page*reqData.Pager.Size))
+	query.Set("limit", fmt.Sprintf("%v", reqData.Pager.Size))
 
 	return query, nil
 }
--- a/backend/plugins/bitbucket/tasks/api_common.go
+++ b/backend/plugins/bitbucket/tasks/api_common.go
@@ -65,11 +65,23 @@ func CreateRawDataSubTaskArgs(taskCtx plugin.SubTaskContext, Table string) (*api
 func GetQuery(reqData *api.RequestData) (url.Values, errors.Error) {
 	query := url.Values{}
 	query.Set("state", "all")
-	query.Set("start", fmt.Sprintf("%v", reqData.Pager.Page*reqData.Pager.Size))
-	query.Set("limit", fmt.Sprintf("%v", reqData.Pager.Size))
-
+	query.Set("pagelen", fmt.Sprintf("%v", reqData.Pager.Size))
+	if reqData.CustomData != nil {
+		query.Set("page", reqData.CustomData.(string))
+	}
 	return query, nil
 }
+
+func GetNextPageCustomData(reqData *helper.RequestData, prevPageResponse *http.Response) (interface{}, errors.Error) {
+	resBody, err := io.ReadAll(prevPageResponse.Body)
+	// decode json from body
+	if rawMessages.Next == `` {
+		return ``, helper.ErrFinishCollect
+	}
+	u, err := url.Parse(rawMessages.Next)
+	println(u.Query()[`page`][0])
+	return u.Query()[`page`][0], nil
+},
 
 func GetTotalPagesFromResponse(res *http.Response, args *api.ApiCollectorArgs) (int, errors.Error) {
 	body := &BitbucketPagination{}
--- a/backend/plugins/bitbucket/tasks/api_common.go
+++ b/backend/plugins/bitbucket/tasks/api_common.go
@@ -62,26 +62,60 @@ func CreateRawDataSubTaskArgs(taskCtx plugin.SubTaskContext, Table string) (*api
 	return RawDataSubTaskArgs, data
 }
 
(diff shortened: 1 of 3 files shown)


backend/plugins/bitbucket/tasks/pr_convertor.go
```diff
@@ -70,6 +70,17 @@ func ConvertPullRequests(taskCtx plugin.SubTaskContext) errors.Error {
 		},
 		Convert: func(inputRow interface{}) ([]interface{}, errors.Error) {
 			pr := inputRow.(*models.BitbucketPullRequest)
+
+			// Getting the merge reference commit
+			mergeCommit := &models.BitbucketCommit{}
+			err = db.First(mergeCommit, dal.Where("LEFT(sha, 12) = ?", pr.MergeCommitSha))
```
likyh, line 76: I'm not sure if the function `LEFT` can be used in both MySQL and pg. Would like confirm it?

backend/plugins/bitbucket/tasks/api_common.go
```diff
@@ -65,8 +65,8 @@ func CreateRawDataSubTaskArgs(taskCtx plugin.SubTaskContext, Table string) (*api
 func GetQuery(reqData *api.RequestData) (url.Values, errors.Error) {
 	query := url.Values{}
 	query.Set("state", "all")
-	query.Set("page", fmt.Sprintf("%v", reqData.Pager.Page))
-	query.Set("pagelen", fmt.Sprintf("%v", reqData.Pager.Size))
+	query.Set("start", fmt.Sprintf("%v", reqData.Pager.Page*reqData.Pager.Size))
```
likyh, line 68: `reqData.Pager.Skip` = `reqData.Pager.Page*reqData.Pager.Size`

Let's use `reqData.Pager.Skip` directly.

backend/plugins/bitbucket/tasks/pr_convertor.go
```diff
@@ -70,6 +70,17 @@ func ConvertPullRequests(taskCtx plugin.SubTaskContext) errors.Error {
...
+			if err != nil {
+				mergeCommit.CommittedDate = *pr.MergedAt
```
likyh, line 78: Actually, `pr.MergedAt` is always empty. How about set `pr.MergedAt = &mergeCommit.CommittedDate` when `err == nil` directly?

backend/plugins/bitbucket/tasks/api_common.go
likyh: I try to user start/limit, but it just reqeust the first page. That means `start/limit` are not work. https://developer.atlassian.com/cloud/bitbucket/rest/intro/#pagination

So I think `page/pagelen` is right. We need add `GetTotalPages` for collector to avoid this issue.

likyh: https://jira.atlassian.com/browse/BCLOUD-19593

image

It looks like we can not build a query on the client and we can only request it one page by one page.

backend/plugins/bitbucket/tasks/pr_convertor.go
jrflga: That make sense. I'll change this.

backend/plugins/bitbucket/tasks/pr_convertor.go
jrflga: I've tested with MySQL and a PostgreSQL instance, and it worked fine. According to the documentation, the function exists in both databases.
https://www.postgresql.org/docs/current/functions-string.html
https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_left

backend/plugins/bitbucket/tasks/api_common.go
jrflga: You're right; sorry for the change to start/limit. It's used on other collections of routes on Bitbucket API.
Maybe we can change how pagination on `reqData` is mounted because it's different depending on what API is used.
We have no access to the `next` property when building `GetQuery` function in `NewApiCollector`, so we can't fetch the next page.

likyh: Yes, we need a new param in NewApiCollector to request API one by one. I'll add it. Please wait a few days.

likyh: https://github.com/apache/incubator-devlake/pull/4320
Added a new pages `PageInOrder` to support this situation.

And I find the page token is the same when `pagelen` is the same. Maybe we can write a map to translate page to page token instead of using `PageInOrder`.

（no matter which pr is）
when pagelen=100 page=2: nextPage=67sER
when pagelen=100 page=2: nextPage= znuod
……

jrflga: Thank you so much!
I'll change the `NewApiCollector` on Bitbucket to make use of this new param.

Maybe the `page` attribute is defined by some form of encoding?

likyh: Hi, I submit a commit to complete `GetQuery`. Please take a look. If you think it's good, I'' merge this PR and cherry-pick to 0.15

jrflga: @likyh, it's perfect! Thanks for the help.

likyh (approved):

PR #4295, opened 2023-02-02
fix: delete gitlab task in bitbucket

likyh:
### Summary
delete gitlab task in bitbucket.

bitbucket cannot run before this change.


klesh (approved):
LGTM

PR #4297, opened 2023-02-02
Fix(tapd): keep old schemas in init migration script

warren830:
### Summary
This is upgrading for hexun-tapd.
The only diff with release-v0.15.0-rc5 is that we will keep old schemas in tapd init migration script


klesh (approved):
LGTM

PR #4311, opened 2023-02-03
fix: optimize the median calculation logic of sql

### LINKED ISSUE #4291: Getting stream timeout error in most dashboards and huge time in querying.
### Search before asking

### What happened

Hi,
We've tried using this and happen to see a huge latency in querying the results especially under DORA dashboard and as result ended up in stream timeout error, we've increased the limits in the grafana.ini file.
Also able to see huge spike in the mysql process which kills the intances, so de-coupled the Database source to Cloud SQL and connected while querying from there as well there is huge spike in latency and also mysql instances received 100 CPU usage and stop responding we've even increased our mysql size to 12 core CPU but still there is 100% CPU usage, is there any known issue for the same? it would be really great if we could get any help for the same. Thanks much.

Well it's a query from sql and it supposed be getting it response but its taking delay and going into stream timeout in almost all cases.

### How to reproduce

While running DORA metrics dashboard mostly. 

### Anything else

_No response_

### Version

v0.15.0

### Code of Conduct

abeizn:
### Summary
optimize the median calculation logic of sql on dora dashboards

Closes #4291 

### Screenshots
image

### Other Information


klesh:
@Startrekzky Please take look at this when you have time, thanks

Startrekzky (approved):
LGTM

Issue #2305, opened 2022-06-22
[Feature][SonarQube] Integration SonarQube


summergan:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

I hope to integrate Sonarqube into Devlake.
The following data can be obtained:
1. The rating of individual code
2. The issues(Bugs/Code Smells/Vulnerabilities) of individual code
3. The duplications lines of individual code

### Use case

The following data can be obtained:
1. The rating of individual code
2. The issues(Bugs/Code Smells/Vulnerabilities) of individual code
3. The duplications lines of individual code

### Related issues

N/O

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Startrekzky:
Thanks @summergan , this is on our [roadmap](https://devlake.apache.org/docs/Overview/Roadmap). Thanks for the proposal.

aggmoulik:
Hey @Startrekzky I would love to work on a plugin for this, I just have same questions about this.

> What kind of metrics do you want to see from SonarQube?
How would you plan to use it after SonarQube is supported?

summergan:
![image](https://user-images.githubusercontent.com/38934767/177237678-bb728aad-74da-4d46-9bc8-2f821159ab40.png)
hi, I can get overall code metrics including Bugs, Vulnerabilities, and so on in SonarQube Dashboard. I hope I can get each developer's data。
e.g. Developer A, development of two projects. Here hope can get the total bugs, Bugs, Vulnerabilities, and so on (Maybe can follow up on Dashboard data)

Startrekzky:
Thanks @summergan , this is super helpful. Please take a look. @aggmoulik

Startrekzky:
Hi @aggmoulik ,  I've collected requirements from the community members and most of the requirements are similar to what @summergan commented above. 

Have you started on this ticket? If not, I suggest coming up with the data schema for SonorQube first, since DevLake doesn't have the data schema for a quality-domain tool like SonarQube. You can draw a diagram for SonarQube and put it in the comment section so that we can discuss. This is [Apache DevLake's current domain layer schema](https://devlake.apache.org/docs/DataModels/DevLakeDomainLayerSchema) for your reference.

[...]

PR #4364, opened 2023-02-09
fix(sonarqube): fix minor issues

warren830:
### Summary
Fix some minor issues

part of #2305 

### Screenshots
image

### Other Information


backend/core/models/domainlayer/securitytesting/st_issue_code_blocks.go
```diff
@@ -21,14 +21,13 @@ import "github.com/apache/incubator-devlake/core/models/domainlayer"
 
 type StIssueCodeBlock struct {
 	domainlayer.DomainEntity
-	IssueCodeBlockId string `json:"issueCodeBlockId"`
```
abeizn, line 24: is this no longer needed？

warren830: I think this is better to be the same to other entities. we have domain id now, we can delete it

abeizn (approved):
LGTM

PR #4385, opened 2023-02-10
fix: github test connection not working

### LINKED ISSUE #4375: [Bug][github] Github test connection not working correctly
### Search before asking

### What happened

Test Connection always fails after refactoring.

It is caused by the Test Connection passing the `conn` struct directly into `NewApiClientFromConnection` which should be a pointer instead, offending code:
https://github.com/apache/incubator-devlake/blob/d2815d5ffe4168d0080237f3c0f5ebc2240032b6/backend/plugins/github/api/connection.go#L52

Github  test connection should work

### How to reproduce

input valid token for github test connection, see error

### Anything else

_No response_

### Version

main

### Code of Conduct

klesh:
### Summary
fix: github test connection not working

Closes #4375 

### Screenshots

github-testconn

when passing the `connection` to `NewApiClientFromConnection`
image

### DIFF
--- a/backend/helpers/pluginhelper/api/api_client.go
+++ b/backend/helpers/pluginhelper/api/api_client.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


likyh:
Why is there a problem without passing the pointer

likyh (approved):

PR #4416, opened 2023-02-14
fix(config-ui): transformation name is duplicated when upgrading bp

mintsweet:
### Summary
Fixed transformation name is duplicated when upgrading bp.

### DIFF
--- a/config-ui/src/error/components/bp-upgrade/use-bp-upgrade.ts
+++ b/config-ui/src/error/components/bp-upgrade/use-bp-upgrade.ts
@@ -90,27 +90,28 @@ export const useBPUpgrade = ({ id, onResetError }: UseBPUpgradeProps) => {
   };
 
   const upgradeScope = async (plugin: string, connectionId: ID, scope: any) => {
+    // get data scope detail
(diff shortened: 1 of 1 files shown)


config-ui/src/error/components/bp-upgrade/use-bp-upgrade.ts
```diff
@@ -90,27 +90,28 @@ export const useBPUpgrade = ({ id, onResetError }: UseBPUpgradeProps) => {
...
+    const scopeId = getScopeId(plugin, scopeDetail);
+
     let transformationRule;
 
     if (scope.transformation) {
       // create transfromation template
       transformationRule = await API.createTransformation(plugin, {
         ...scope.transformation,
-        name: `upgrade-${plugin}-${connectionId}-${new Date().getTime()}`,
+        name: `upgrade-${plugin}-${connectionId}-${scopeId}`,
```
likyh, line 103: Does it will duplicated if one scope in 2 different bp?

mintsweet: You're right, I'll fixed it.

mintsweet:
> Have you tested it?

Yes.

PR #4424, opened 2023-02-15
fix: creating a connection with the same name will overwrite the previous one

### LINKED ISSUE #4224: [Bug][Connection] Creating a connection with the same name will overwrite the previous one
### Search before asking

### What happened

Creating a connection with the same name will not report an error or prompt, and will overwrite the previous configuration.

The creation failed, prompting that the name already exists, and the previous one cannot be overwritten.

### How to reproduce

1. Create a github connection and named **GitHub**.
2. Create another connection named **GitHub** while changing any parameter.
3. This parameter created later will overwrite the previously created one.

### Anything else

_No response_

### Version

main

### Code of Conduct

mindlesscloud:
### Summary
There is a unique key constraint imposed on the column `name`. Creating a new connection with an existing name should throw an error instead of overwriting the old one.

Closes #4224 

### Screenshots
image

### DIFF
--- a/backend/helpers/pluginhelper/api/connection_helper.go
+++ b/backend/helpers/pluginhelper/api/connection_helper.go
@@ -18,17 +18,15 @@ limitations under the License.
 package api
 
 import (
-	"github.com/apache/incubator-devlake/core/models"
-	"strconv"
-	"strings"
-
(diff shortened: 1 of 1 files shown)


abeizn (approved):
LGTM

PR #4431, opened 2023-02-16
fix(github): process situation without ratelimit header

### LINKED ISSUE #4387: error running blueprint on github enterprise.
### Search before asking

### What happened

If we run the blueprint without manually speciying the fixed rate limit in the github connection the blueprint gives the following error. However the issue fixes if we specify the fixed rate in the connection.
Screenshot 2023-02-10 at 1 53 23 PM

While creating a simple github connection we do not need to specify the fixed rate limit. Similarily when creating the github enterprise connection we should also not need to specify the fixed rate limit

### How to reproduce

Create a github enterprise connection without specifying the Fixed rate limit .
Create a blueprint with the same connection.
Then run the blueprint.

### Anything else

_No response_

### Version

main

### Code of Conduct

warren830:
### Summary
According to issue #4387 , if users are using github enterprise server, the response header does not have `X-RateLimit-Limit`. Under this situation, we can just use globalRatelimit in devlake

Closes #4387

### DIFF
--- a/backend/plugins/github/tasks/api_client.go
+++ b/backend/plugins/github/tasks/api_client.go
@@ -54,9 +54,17 @@ func CreateApiClient(taskCtx plugin.TaskContext, connection *models.GithubConnec
 			}
 			return remaining * len(tokens), time.Unix(int64(reset), 0).Sub(date), nil
 			*/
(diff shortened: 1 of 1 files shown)


klesh (approved):
LGTM

PR #4440, opened 2023-02-17
fix(bamboo): simplify test connectino and use json instead of xml

warren830:
### Summary
We can use RESOURCE_NAME.json to get json data.
One more thing, instead of getting repository to test connection, we'd better get server info to support multiple versions of bamboo in future

Closes xx

### Screenshots
image

### Other Information


mappjzc (approved):

klesh:
Nice work

Issue #4403, opened 2023-02-13
[Bug][Jira] The behaviour of createdDateAfter option does not fit the user scenario


narrowizard:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

In our situation, we do need the limited history sync feature introduced by 0.15.  
With solution implemented in 0.15, we lost issues which is created before but updated after `CreatedDateAfter` time.  
Those issues (created before but updated after) are useful when analysing, because the updating operation means it is picked up again recently.

### What do you expect to happen

We need all data which has updation after the time we specified.

### How to reproduce

No

### Version

v0.15.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

klesh:
# Plan of attack for v0.16

1. Add `TimeAfter` support alongside `CreatedDateAfter` for JIRA ( including framework modification), review, and merge.
2. Update all plugins one by one in terms of multiple PRs
3. Cleanup: Rename `CreatedDateAfter` to `TimeAfter` in the database and remove all deprecated code
4. Update config-ui  @mintsweet

PR #4442, opened 2023-02-17
feat: github runs collector supports timeFilter/diffSync

klesh:
### Summary
github runs collector supports timeFilter/diffSync

Due to the fact that github runs API doesn't support `since`(filter by updated_at) parameter, 
this PR uses another approach to support timeFilter and diffSync:

1. Collect newly added records by examining the `created_at` field
2. Load unfinished records in the database and collect their details from API if diffSync is in action

Part of #4403

### Self test:

I took the following steps to verify:

1. I set the `timeAfter`, and collect data, and observed the database, it worked as expected
2. I then reran the task with the same parameters as the previous step, and no data changed which is correct
3. I then modified the `timeAfter` parameter to a later date and the database stay the same which is correct
4. lastly, I pushed `timeAfter` to an older date and records increased as expected

### DIFF
--- a/backend/plugins/github/tasks/cicd_run_collector.go
+++ b/backend/plugins/github/tasks/cicd_run_collector.go
@@ -20,15 +20,25 @@ package tasks
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
+	"reflect"
+	"time"
 
+	"github.com/apache/incubator-devlake/core/dal"
 	"github.com/apache/incubator-devlake/core/errors"
 	"github.com/apache/incubator-devlake/core/plugin"
 	helper "github.com/apache/incubator-devlake/helpers/pluginhelper/api"
+	"github.com/apache/incubator-devlake/plugins/github/models"
 )
 
 const RAW_RUN_TABLE = "github_api_runs"
+
+// Although the API accepts a maximum of 100 entries per page, sometimes
+// the response body is too large which would lead to request failures
+// https://github.com/apache/incubator-devlake/issues/3199
+const PAGE_SIZE = 30
 
 var CollectRunsMeta = plugin.SubTaskMeta{
 	Name:             "collectRuns",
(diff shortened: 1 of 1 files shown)


backend/plugins/github/tasks/cicd_run_collector.go
```diff
@@ -52,48 +62,131 @@ func CollectRuns(taskCtx plugin.SubTaskContext) errors.Error {
...
+			Params: GithubApiParams{
+				ConnectionId: data.Options.ConnectionId,
+				Name:         data.Options.Name,
+			},
+			Table: RAW_RUN_TABLE,
+		},
+		ApiClient:   data.ApiClient,
+		Input:       iterator,
+		Incremental: true,
+		UrlTemplate: "repos/{{ .Params.Name }}/actions/runs/{{ .Input.ID }}",
```
likyh, line 176: How many runs will collect one by one in a collector?

backend/plugins/github/tasks/cicd_run_collector.go
```diff
@@ -52,48 +62,131 @@ func CollectRuns(taskCtx plugin.SubTaskContext) errors.Error {
...
+func collectUnfinishedRuns(taskCtx plugin.SubTaskContext) errors.Error {
+	data := taskCtx.GetData().(*GithubTaskData)
+	db := taskCtx.GetDal()
+
+	// load unfinished runs from the database
+	cursor, err := db.Cursor(
+		dal.Select("id"),
+		dal.From(&models.GithubRun{}),
+		dal.Where(
+			"repo_id = ? AND connection_id = ? AND status IN ('ACTION_REQUIRED', 'STALE', 'IN_PROGRESS', 'QUEUED', 'REQUESTED', 'WAITING', 'PENDING')",
```
likyh, line 151: Will the extractor be confuse if there are more than one raw layer record about one github run?

backend/plugins/github/tasks/cicd_run_collector.go
klesh: What do you mean? All unfinished records would be collected in parallel.

backend/plugins/github/tasks/cicd_run_collector.go
klesh: TDLR: No,  it won't, we are expecting `raw_table` to have multiple records for one run.


We actually are counting on this behavior, the extractor is working in a `CreateOrUpdate` manner so `Incremental Collection` could function correctly because we are deleting the whole BATCH before extraction which means we have to keep the OLDER data in the `raw_table` since Incremental collection would only fetch the NEWER data.

likyh (approved):

Issue #4448, opened 2023-02-17
[Bug][Config-UI] Config-UI only lists public repos for a user


hezyin:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

In v0.15, Config-UI uses `/users/{username}/repos` to list repos for a GitHub user. However, according to GitHub API, this endpoint only list **public** repos of that user: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-a-user

We should probably use this endpoint (`/user/repos`) instead: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-the-authenticated-user

### What do you expect to happen

Users are able to see their private repos in Config-UI as long as their PAT has the right permission.

### How to reproduce

See above

### Version

v0.15.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #4453, opened 2023-02-20
fix(config-ui): github does not list private membership organizations

### LINKED ISSUE #4439: [Bug][Config-UI] Config-UI not listing private membership organizations for GitHub
### Search before asking

### What happened

So some users report in Slack that they don't see their organizations showing up in the first column of the miller column in Config-UI. 

The issue is that when Config-UI fetches GitHub orgs, it uses /users/<username>/orgs, which only list public organization memberships, regardless of authentication. To list all organizations (public and private) for the authenticated user, we would have to use another endpoint /user/orgs . 

GitHub API reference: https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#list-organizations-for-the-authenticated-user

Config-UI list all organizations of an user regardless of whether the org membership is public or not.

### How to reproduce

See above

### Anything else

_No response_

### Version

v0.15.0

### Code of Conduct

mintsweet:
### Summary
GitHub does not list private membership organizations.

Closes #4439 

### Screenshots
screenshot-20230220-143830

screenshot-20230220-154145

### DIFF
--- a/config-ui/src/plugins/register/github/api.ts
+++ b/config-ui/src/plugins/register/github/api.ts
@@ -25,8 +25,8 @@ type PaginationParams = {
 
 export const getUser = (prefix: string) => request(`${prefix}/user`);
 
-export const getUserOrgs = (prefix: string, username: string, params: PaginationParams) =>
-  request(`${prefix}/users/${username}/orgs`, {
+export const getUserOrgs = (prefix: string, params: PaginationParams) =>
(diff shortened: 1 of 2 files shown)


likyh:
what different from the prev one?

mintsweet:
@likyh #4439 and #4448

this pr will fix #4439, prev pr fixed #4448

likyh (approved):

PR #4455, opened 2023-02-20
fix: check array length before accessing for github run collector

klesh:
### Summary
fix: check array length before accessing for github run collector


mindlesscloud (dismissed):
LGTM

mindlesscloud (approved):
LGTM

Issue #3322, opened 2022-10-08
[Feature][plugin] Support Bamboo CI v6.8.1


Startrekzky:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Stage 1: Backend support

### Use case

One of the users uses Bamboo CI as the CI tool, they want to see CI metrics via DevLake. The version they use is Bamboo CI v6.8.1

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #4460, opened 2023-02-20
feat(bamboo): add plan and job build

warren830:
### Summary
This pr added job and plan builds.
There is a trick, job and plan builds have the same url but different path pattern, and almost the same structs but different type. Yes, we can process them together and convert them to cicd_pipelines and cicd_tasks according to type, but I think if we process them separately, it will be easier for us to debug or modify.

part of #3322 

### Screenshots
image

image

### Other Information

### DIFF
--- a/backend/plugins/bamboo/tasks/plan_build_collector.go
+++ b/backend/plugins/bamboo/tasks/plan_build_collector.go
@@ -1,0 +1,109 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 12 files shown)


backend/plugins/bamboo/tasks/plan_build_collector.go
```diff
@@ -0,0 +1,109 @@
...
+		ApiClient:   data.ApiClient,
+		PageSize:    100,
+		Input:       iterator,
+		UrlTemplate: "result/{{ .Input.PlanKey }}.json",
+		Query: func(reqData *helper.RequestData) (url.Values, errors.Error) {
+			query := url.Values{}
+			query.Set("showEmpty", fmt.Sprintf("%v", true))
+			query.Set("expand", "results.result.vcsRevisions")
+			query.Set("max-result", fmt.Sprintf("%v", reqData.Pager.Size))
+			query.Set("start-index", fmt.Sprintf("%v", (reqData.Pager.Page-1)*reqData.Pager.Size))
```
likyh, line 70: `(reqData.Pager.Page-1)*reqData.Pager.Size` -> `reqData.Pager.Skip`

likyh (approved):

PR #4477, opened 2023-02-21
fix(sonarqube): fix data loss

warren830:
### Summary
1. In mysql, the default collation is utf8mb4_general_ci is case insensitive. So when I was working on sonarqube dashboard, I found that we missed some data.
This pr mainly add the following to mysql part in docker-compose.yml:
```yaml
    command:
      --character-set-server=utf8mb4
      --collation-server=utf8mb4_bin
```
utf8mb4_bin is case sensitive

2. In addition, because we need to update Sonarproject.LastAnalysisDate, we need to request apiProject whenever we want to collect data. 

part of #2305 

### Screenshots

### Other Information

### DIFF
--- a/backend/plugins/sonarqube/impl/impl.go
+++ b/backend/plugins/sonarqube/impl/impl.go
@@ -123,30 +123,22 @@ func (p Sonarqube) PrepareTaskData(taskCtx plugin.TaskContext, options map[strin
 		Options:   op,
 		ApiClient: apiClient,
 	}
-	if op.ProjectKey != "" {
-		var scope *models.SonarqubeProject
-		var apiProject *tasks.SonarqubeApiProject
-		// support v100 & advance mode
-		// If we still cannot find the record in db, we have to request from remote server and save it to db
-		db := taskCtx.GetDal()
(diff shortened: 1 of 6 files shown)


backend/plugins/sonarqube/impl/impl.go
```diff
@@ -123,30 +123,22 @@ func (p Sonarqube) PrepareTaskData(taskCtx plugin.TaskContext, options map[strin
...
-			scope.ConnectionId = op.ConnectionId
-			err = taskCtx.GetDal().CreateIfNotExist(&scope)
-			if err != nil {
-				return nil, err
-			}
-		}
-		if err != nil {
-			return nil, errors.Default.Wrap(err, fmt.Sprintf("fail to find project: %s", op.ProjectKey))
-		}
+	// even we have project in _tool_sonaqube_projects, we still need to collect project to update LastAnalysisDate
```
likyh, line 126: why? what column updated?

likyh: If the date will be updated each time, maybe we should split the model into 2 model

warren830: but this come from the same api. Hmm, maybe we don't need to update the model, just set the value to taskData

PR #4507, opened 2023-02-23
fix: remove createdDateAfter completely

### LINKED ISSUE #4403: [Bug][Jira] The behaviour of createdDateAfter option does not fit the user scenario

klesh:
### Summary
According to the Plan of  Attack from #4403, this is the last step of the BE side to replace `createdDateAfter` with `timeAfter`

@mintsweet  the `config-ui` may have to be modified accordingly. you may close #4403 from your PR, thanks.

Part of #4403


likyh (approved):

backend/helpers/pluginhelper/api/api_collector_with_state.go
```diff
@@ -66,37 +64,17 @@ func NewApiCollectorWithStateEx(args RawDataSubTaskArgs, createdDateAfter *time.
...
-		return false
-	}
-	// prioritize TimeAfter parameter: collector should filter data by `updated_date`
 	if m.TimeAfter != nil {
 		return m.LatestState.TimeAfter == nil || !m.TimeAfter.Before(*m.LatestState.TimeAfter)
 	}
-	// Deprecating(timeAfter): to be removed
-	// fallback to CreatedDateAfter: collector should filter data by `created_date`
-	return m.LatestState.CreatedDateAfter == nil || m.CreatedDateAfter != nil && !m.CreatedDateAfter.Before(*m.LatestState.CreatedDateAfter)
+	return m.LatestState.LatestSuccessStart != nil
```
likyh, line 77: no

likyh: check LatestSuccessStart first. Because when LatestState exists but the latest collection fails (LatestSuccessStart=nil), it will be true that `m.TimeAfter != nil && m.LatestState.TimeAfter == nil`.

likyh (changes requested):
a logic bug

backend/helpers/pluginhelper/api/api_collector_with_state.go
klesh: You are correct, thanks for pointing our. fixed

likyh (approved):

PR #4508, opened 2023-02-23
fix: format timeAfter to RFC3339 for updated_on

likyh:
### Summary
format timeAfter to RFC3339 for updated_on in bitbucket.


klesh (approved):
LGTM

PR #4510, opened 2023-02-24
fix(config-ui): adjust the api field

### LINKED ISSUE #4403: [Bug][Jira] The behaviour of createdDateAfter option does not fit the user scenario

mintsweet:
### Summary
Adjust the api field.

Closes #4403


config-ui/src/pages/blueprint/components/sync-policy/index.tsx
```diff
@@ -72,7 +72,7 @@ export const SyncPolicy = ({
         <div className="block">
           <h3>Time Filter *</h3>
           <p>Select the data range you wish to collect. DevLake will collect the last six months of data by default.</p>
-          <StartFromSelector value={createdDateAfter} onChange={onChangeCreatedDateAfter} />
+          <StartFromSelector value={timeAfter} onChange={onChangeCreatedDateAfter} />
```
likyh, line 75: should `setCreatedDateAfter` be changed?

mintsweet: ok, you're right.

likyh (approved):

Issue #4486, opened 2023-02-22
[Bug][Core] panic: ERROR: current transaction is aborted, commands ignored until end of transaction block 


narrowizard:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

It seems devlake make deadlock and panic when PIPELINE_MAX_PARALLEL > 1

### What do you expect to happen

It should run pipelines in parallels if PIPELINE_MAX_PARALLEL > 1.

### How to reproduce

set env PIPELINE_MAX_PARALLEL = 3, and create multi pipelines at the same time. some pipelines with label 'parallels/typeA', and others with label 'parallels/typeB'.

### Anything else

```
message:{"log":"\u001b[0m\u001b[33m[0.309ms] \u001b[34;1m[rows:0]\u001b[0m UPDATE \"_devlake_pipelines\" SET \"began_at\"='2023-02-22 07:06:38.72',\"message\"='',\"status\"='TASK_RUNNING',\"updated_at\"='2023-02-22 07:06:38.72' WHERE id = 0\"\n","stream":"stdout","time":"2023-02-22T07:06:38.721141119Z"}	devlake:/var/log/containers/devlake-6fd9b9559f-p64qk_default_devlake-e7b5e6643f117a90e067a5367b366729eb90507c3fa382d5453b11119da450e0.log	t:07:06:39.+0000
message:{"log":"panic: ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)\n","stream":"stderr","time":"2023-02-22T07:06:38.723916307Z"}	devlake:/var/log/containers/devlake-6fd9b9559f-p64qk_default_devlake-e7b5e6643f117a90e067a5367b366729eb90507c3fa382d5453b11119da450e0.log	t:07:06:39.+0000
message:{"log":"Wraps: (2) ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)\n","stream":"stderr","time":"2023-02-22T07:06:38.723929524Z"}	devlake:/var/log/containers/devlake-6fd9b9559f-p64qk_default_devlake-e7b5e6643f117a90e067a5367b366729eb90507c3fa382d5453b11119da450e0.log	t:07:06:39.+0000
[...]

PR #4551, opened 2023-02-28
fix: add some log; wait when query fail in RunPipelineInQueue

likyh:
### Summary
1. add some log; 
2. wait when the query fails in RunPipelineInQueue. The old logic will run pipeline when `err != nil` and err is not RecordNotFound.

Related to #4486


klesh (approved):
LGTM

PR #4583, opened 2023-03-03
fix: sonaqube dashboard fix

mappjzc:
### Summary
Change creation_date to be created_date

Add Default value for rows.


abeizn (approved):
LGTM

PR #4587, opened 2023-03-06
fix(config-ui): some bugs

mintsweet:
### Summary
1. bp not properly filterd.
2. rate limit lost default value.
3. add default value for transformation.
4. adjust the rules for project name.
5. not reset value after project create.
6. adjust the list page parameter pageSize.
7. endpoint subLabel error for gitlab.
8. add plugin customize.
9. add default value for use-operator's formatReason.
10. optimize the creation of webhook in the project.


config-ui/src/pages/project/detail/use-project.ts
```diff
@@ -55,7 +55,7 @@ export const useProject = (name: string) => {
   }, []);
 
   const handleUpdate = async (newName: string, enableDora: boolean) => {
-    if (!/^\w+$/.test(newName)) {
+    if (!/^\w|-\/+$/.test(newName)) {
```
likyh, line 58: It not works, I think it means /(^\w)|(-(\/+))$/.

config-ui/src/pages/project/home/use-project.ts
```diff
@@ -53,7 +53,7 @@ export const useProject = <T>({ name, enableDora, onHideDialog }: Props) => {
   }, []);
 
   const handleSave = async () => {
-    if (!/^\w+$/.test(name)) {
+    if (!/^\w|-\/+$/.test(name)) {
```
likyh, line 56: Same above

mintsweet: thanks.

config-ui/src/pages/project/detail/use-project.ts
mintsweet: thank.

likyh (approved):

PR #4604, opened 2023-03-07
feat: allow protocol used in proxying to be overridden

samcrang:
### Summary
In our deployment Grafana and Devlake are using HTTPS. This change allows us to proxy to those endpoints without having to modify the image ourselves.

This should retain the existing behavior and therefore be backwards compatible.

N/A
### Screenshots
N/A
### Other Information
N/A


klesh (approved):
LGTM

Issue #4653, opened 2023-03-14
[Refactor][Framework] Extract scope/remote helper to avoid duplicated code in every plugin


warren830:
## What and why to refactor
Right now, in our plugins, we have to duplicate a lot of code to write scope/remote api

## Describe the solution you'd like
We can extract scope/remote helper to reduce duplicated code

warren830:
fixed already, please check related prs

PR #4660, opened 2023-03-14
refactor(framework): modify scope helper

warren830:
### Summary
Fix return value which should be one layer struct as below:
image
But it was :
image

part of #4653


likyh (approved):
Let's Try.

PR #4662, opened 2023-03-14
refactor: remove depracted `rand.Seed`

d4x1:
### ⚠️ Pre Checklist

### Summary
rand.Seed is deprecated, more detail https://pkg.go.dev/math/rand#Seed

No

### Screenshots
No need

### Other Information
No need


backend/server/services/notification.go
```diff
@@ -22,17 +22,17 @@ import (
...
 	"math/rand"
 	"net/http"
 	"strings"
 	"time"
+
+	"github.com/apache/incubator-devlake/core/errors"
+	"github.com/apache/incubator-devlake/core/models"
 )
 
 func init() {
```
warren830, line 35: I think you can also delete this function

warren830:
Nice work! I left a comment, let me know whenever you need any help

warren830:
BTW, are you already in our wechat group? @d4x1

backend/server/services/notification.go
d4x1: Of course it should be deleted, I have made a mistake.

d4x1:
Updated. Fix unnecessary `init`.

klesh (changes requested):
Accroding to https://pkg.go.dev/math/rand#Seed
1. The function is deprecated from go v1.20. but we are at v1.19 https://github.com/apache/incubator-devlake/blob/main/backend/go.mod#L3 . We shouldn't remove it unless we upgrade to v1.20
2. If we do so, I think the only thing we need to do is remove the `init` function since it will be seeded randomly automatically.

d4x1:
> Accroding to https://pkg.go.dev/math/rand#Seed
> 
>     1. The function is deprecated from go v1.20. but we are at v1.19 https://github.com/apache/incubator-devlake/blob/main/backend/go.mod#L3 . We shouldn't remove it unless we upgrade to v1.20
> 
>     2. If we do so, I think the only thing we need to do is remove the `init` function since it will be seeded randomly automatically.

Although it was marked deprecated since 1.20, `rand.New()` can work in 1.19 https://pkg.go.dev/math/rand@go1.19#New . So this change is safe .I can see some  cases in main branch https://github.com/apache/incubator-devlake/blob/main/backend/core/plugin/plugin_utils.go#LL148

As for the semantics, according to the comments in `rand.Seed`, `rand.New(source)` is equivalent to `rand.Seed && rand.Int`,  just like this PR did.

Also we can use `rand.Int`to produce random integer, IMO, this two ways are equivalent. The both works.

klesh:
> > Accroding to https://pkg.go.dev/math/rand#Seed
> > ```
> > 1. The function is deprecated from go v1.20. but we are at v1.19 https://github.com/apache/incubator-devlake/blob/main/backend/go.mod#L3 . We shouldn't remove it unless we upgrade to v1.20
> > 
> > 2. If we do so, I think the only thing we need to do is remove the `init` function since it will be seeded randomly automatically.
> > ```
> 
> Although it was marked deprecated since 1.20, `rand.New()` can work in 1.19 https://pkg.go.dev/math/rand@go1.19#New . So this change is safe .I can see some cases in main branch https://github.com/apache/incubator-devlake/blob/main/backend/core/plugin/plugin_utils.go#LL148
> 
> As for the semantics, according to the comments in `rand.Seed`, `rand.New(source)` is equivalent to `rand.Seed && rand.Int`, just like this PR did.
> 
> Also we can use `rand.Int`to produce random integer, IMO, this two ways are equivalent. The both works.

Creating and seeding a new `Rand` instance on each call looks inefficient.
However, I understand your point. I suggest that we do the following:

1. these 2 functions are doing the same thing, we should merge them and move it to the `core/utils` package
2. we can create and seed a new `Rand` at this point, but we should also add a comment to declare that it can be optimized if we upgrade to go v1.20

d4x1:
Agreed! These two functions do the same thing. Of course them can be merged and moved to `core/utils`.
But after doing that, I think there are some options to deal with it.
1.  without `init` function where we can do `rand.Seed` in,  and use `rand.Int` directly in `RandSomeStr`. this will change the original behavior.
2. without `init` function where we can do `rand.Seed` in, and use `rand.New(rand.NewSource)` in `RandSomeStr`.  This will be inefficient.
3. add `init` function, we can  do `rand.Seed` in it, and  `rand.Int` in `RandSomeStr`, add some comments.


I think 1 and 2 are the right way. 3 is temporary and outdated.

klesh:
@d4x1 
Yes, I was saying that we should opt for option 2 for now because it is not a true bottleneck of our system. 
We may opt for 1 in the future after we upgrade to go v1.20 .

klesh:
Aah, I see. Actually, we don't need the `ENCODE_KEY` to be capitalized. I suggest we just remove the `RandomEncKey` completely and replace it with `RandLetterBytes`

d4x1:
> Aah, I see. Actually, we don't need the `ENCODE_KEY` to be capitalized. I suggest we just remove the `RandomEncKey` completely and replace it with `RandLetterBytes`


updated. 
I was busy with my fulltime job these days,  sorry for the late update.

klesh:
@d4x1 No worries, thank you for your hard work. The code looks good to me.
Can you attach some screenshots to demonstrate the new code is working correctly so it can be merged?
Thanks for your contribution.

klesh (approved):

d4x1:
> @d4x1 No worries, thank you for your hard work. The code looks good to me. Can you attach some screenshots to demonstrate the new code is working correctly so it can be merged? Thanks for your contribution.

Do you mean add some testing cases?  It's a basic refactor, I don't know which screenshots I should provide.
It's just a function and it just works.

klesh:
@d4x1 Normally, we would search which functions/features are using the modified code, and verify the related functions/features are working just like before.
In this case, I would set up a brand-new instance to see if the ENCODE_KEY is generated and saved correctly and take screenshots, and so on and so forth.
I can do it if you can not find time, no worries.

d4x1:
I will try to do like this in other PRs. It will not take too long when I become more familiar with the whole project.

Issue #3708, opened 2022-11-10
[Bug][github_graphql] 502 when collecting pull requests


likyh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

![image](https://user-images.githubusercontent.com/3294100/200982345-a8991341-97b2-42b5-a3ea-c155bf21d129.png)

sometimes PrCollector in graphql will raise 502.

### What do you expect to happen

collect success.

### How to reproduce

1. use graphql_github to collect `https://github.com/apache/jackrabbit-oak`
2. it will raise 502.

### Anything else

The reason is too much data on each page. Now 30 pr, 30 * most 100 commits will bind to one graphql query. But I think it will be too slow if set to less than 30 pr per page. In past, the number is 100 pr per page.

Maybe we need a dynamic page size. It could request 100 per page and reduce less if 502 got.

### Version

main

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

yumengwang03:
@likyh Do we want to put this bug in v0.15? cc @klesh

likyh:
> @likyh Do we want to put this bug in v0.15? cc @klesh

No, I don't think so.

Startrekzky:
This is solved.

PR #4668, opened 2023-03-15
fix: decrease page size for github graphql pr collector

likyh:
### Summary
image

fix: decrease page size for github graphql pr collector

It may slow than the old pr collector about 20%.

Related to #3708


warren830 (approved):
LGTM

PR #4678, opened 2023-03-15
Fix type annotations

samcrang:
### Summary
I don't think these are right. Without this change it seems as though the value in the DB will always be set to something like:

```
typing.Optional[pydevlake.domain_layer.devops.CICDResult]
```

N/A

### Screenshots
N/A

### Other Information
N/A


CamilleTeruel (approved):

warren830 (approved):
camile already approved

Issue #4571, opened 2023-03-02
[Feature][Auth] Test connection also tests token permissions


Startrekzky:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

I used an existing GitHub token to collect GitHub data with DevLake. When I input it in the data connection page and clicked `Test Connection`, it showed `success`. However, when I started to run BP, the pipeline failed.

It turned out that the token I used did not meet the permission required by DevLake. It's an old token so I was not sure about its permissions.

I think the better experience is to prompt `fail` when testing a token without the requried permissions.

### Description

Follow the token permissions in the manuals
- [x] Test [GitHub token's permissions](https://devlake.apache.org/docs/next/Configuration/GitHub#github-personal-access-tokensrecommended)
- [ ] Test [GitLab token's permissions](https://devlake.apache.org/docs/next/Configuration/GitLab#gitlab-personal-access-tokens)
- [ ] Test Jira token's permissions @mindlesscloud 
- [ ] or more

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Startrekzky:
@jakezhu9  
- Your approach to checking GitHub permission is good. 
- Not sure what we can do with GitLab. 
- As for Jira, the approach is to add a doc to describe the nuances of token permission settings. See https://devlake.apache.org/docs/Configuration/Jira#auth-token

PR #4686, opened 2023-03-16
feat: check github token permissions when test connection

jakezhu9:
### Summary
When testing the connection using a classic GitHub token, we will now check its permissions. If the token doesn't have the required permissions, the test will fail.

Part of #4571 

### Screenshots
image


klesh:
Well done! I believe we should check the permission for private repos as well.

@hezyin what do you think

klesh (approved):
LGTM

PR #4695, opened 2023-03-16
Improve test connection

CamilleTeruel:
### Summary
- Make connection.name optional when calling test-connection
- Include error messages in response body
- Rename connection's `pat` to `token`
- Allow empty connection's `proxy`

### DIFF
--- a/backend/python/plugins/azuredevops/azuredevops/main.py
+++ b/backend/python/plugins/azuredevops/azuredevops/main.py
@@ -50,21 +50,23 @@ class AzureDevOpsPlugin(Plugin):
             url=git_repo.url
         )
 
-    def remote_scope_groups(self, ctx) -> list[RemoteScopeGroup]:
-        api = AzureDevOpsAPI(ctx.connection)
-        member_id = api.my_profile.json['id']
+    def remote_scope_groups(self, connection) -> list[RemoteScopeGroup]:
+        api = AzureDevOpsAPI(connection)
+        member_id = api.my_profile().json['id']
         accounts = api.accounts(member_id).json
-        orgs = [acc['accountId'] for acc in accounts]
-        for org in orgs:
+        for account in accounts['value']:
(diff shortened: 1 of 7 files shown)


backend/python/plugins/azuredevops/azuredevops/main.py
```diff
@@ -50,21 +50,23 @@ def domain_scopes(self, git_repo: GitRepository):
             url=git_repo.url
         )
 
-    def remote_scope_groups(self, ctx) -> list[RemoteScopeGroup]:
-        api = AzureDevOpsAPI(ctx.connection)
-        member_id = api.my_profile.json['id']
+    def remote_scope_groups(self, connection) -> list[RemoteScopeGroup]:
```
keon94, line 53: please add type hints

backend/server/services/remote/plugin/connection_api.go
```diff
@@ -22,15 +22,22 @@ import (
 
 	"github.com/apache/incubator-devlake/core/errors"
 	"github.com/apache/incubator-devlake/core/plugin"
+	"github.com/apache/incubator-devlake/server/api/shared"
 	"github.com/apache/incubator-devlake/server/services/remote/bridge"
 )
 
 func (pa *pluginAPI) TestConnection(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
 	err := pa.invoker.Call("test-connection", bridge.DefaultContext, input.Body).Err
 	if err != nil {
```
keon94, line 31: we can't assume the error will always be 401. What if the endpoint was offline or gave us another error code? I think TestConnection needs to have a contract where it writes the datasource's response into a standardized output with a Http Code on it. It might as well be the ApiResourceOutput struct, but written for Python.

keon94:
LGTM

keon94 (approved):

PR #4696, opened 2023-03-17
fix(config-ui): missed timeAfter after change data scope

mintsweet:
### Summary
Missed `timeAfter` after change data scope.

### DIFF
--- a/config-ui/src/pages/blueprint/detail/panel/configuration.tsx
+++ b/config-ui/src/pages/blueprint/detail/panel/configuration.tsx
@@ -96,7 +96,7 @@ export const Configuration = ({ blueprint, operating, onUpdate, onRefresh }: Pro
   const handleUpdateConnection = (updated: any) =>
     onUpdate({
       settings: {
-        version: '2.0.0',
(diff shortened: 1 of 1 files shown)


config-ui/src/pages/blueprint/detail/panel/configuration.tsx
```diff
@@ -96,7 +96,7 @@ export const Configuration = ({ blueprint, operating, onUpdate, onRefresh }: Pro
   const handleUpdateConnection = (updated: any) =>
     onUpdate({
       settings: {
-        version: '2.0.0',
+        ...blueprint.settings,
```
likyh, line 99: If the empty time filter hasn't been pass to the backend, I think we should fix this issue in backend. 😂

mintsweet: I tried an empty timeAfter and it still works.

likyh (approved):
Emm, after this change, maybe users will overwrite unchanged data unconsciously. But I have no idea to deal with it.

PR #4706, opened 2023-03-19
feat: set project mapping by plugin org

### LINKED ISSUE #4534: [Feature][infra] Blueprint/Pipeline Advanced Mode Supports Project/DORA
### Search before asking

### Why

DORA is no doubt one of the most important features of our system judged by the feedback from our new users, many of them joined the community because of it.

In our current design (v0.15), to calculate DORA metrics, we need `project_mapping` to be populated before the DORA plugin gets executed. However, this operation happens during `blueprint.settings` to `blueprint.plan` conversion which is only available in Normal Mode. In other words, Advanced Mode accepts `blueprint.plan` directly **without** conversion, thus no DORA for Advanced Mode. It poses a couple of problems:

1.  Not all our plugins support Normal Mode, especially those in WIP/Beta stage like BitBucket/TaPD/Zentao. It is a blocker for users who are using these Data Sources and are willing to help with testing.
2.  Some users might want to arrange their collection/computation in a particular way. i.e. to collect data in one blueprint but compute the DORA metric in another one. However, Normal Mode has a different agenda(guiding new users), which makes Advanced Mode the only suitable way to support this kind of customization.

In conclusion, **Advanced Mode must support Project/DORA.**

### What

The DORA plugin itself merely depends on the `project_mapping` table which is OK, however, the table is managed by the `framework` and omitted from the `bp.plan` which is the root cause of the problem. I hereby proposed the following solution to fix it without too much hassle:

**Let the plugin maintains the `project_mapping` instead of the `framework` so it can be described in the `bp.plan` field**, for example, the `org` plugin:

Pipeline task json example:

```json
{
    "plugin": "org",
    "subtasks": [ "setProjectMapping" ],
    "options": {
        "projectMappings": [{
            "projectName": "myproject",
            "scopes": [
                { "table": "cicd_scope", "rowId": "gitlab:GitlabProject:1:11876178" },
                { "table": "boards", "rowId": "gitlab:GitlabProject:1:11876178" }
            ]
        }]
    }
}
```

Meanwhile, the Normal Mode logic must be updated to produce the `task` instead of writing the table directly.

### How

With this solution, the Project relationship can be fully described in the `bp.plan` field, it offers more freedom to users to do whatever they see fit.

1.  One might test/use the Beta/WIP plugins in Advanced Mode. To that end, we must provide thorough documentation to guide users.
  a.  Documents to describe `scopes` management for WIP/Beta plugins.
  b.  Add a section to the Project Doc to describe how the `plan` works and where the `org task` should be.
  c.  Offer a workable example `plan` with related docs linked.
2.  For those who want collection/computation in separate blueprints, we shall offer a document with examples to guide them through the process.

### Related issues

_No response_

### Code of Conduct

mindlesscloud:
### Summary
The binding between project and scopes is performed by the subtask `SetProjectMapping`.  We defined a new interface `ProjectMapper` implemented by the plugin org. While creating the blueprint, the framework will find the org plugin and invoke its `MapProject` method. 

Closes #4534

### Screenshots

### DIFF
--- a/backend/core/dal/dal.go
+++ b/backend/core/dal/dal.go
@@ -19,6 +19,7 @@ package dal
 
 import (
 	"database/sql"
+	"fmt"
 	"reflect"
 
 	"github.com/apache/incubator-devlake/core/errors"
@@ -239,7 +240,8 @@ func GetPrimarykeyColumnNames(d Dal, dst Tabler) (names []string, err errors.Err
 		return
 	}
 	for _, pkColumn := range pkColumns {
-		names = append(names, pkColumn.Name())
+		// in case the column name is a reserved identifier
+		names = append(names, fmt.Sprintf("%s.%s", dst.TableName(), pkColumn.Name()))
 	}
 	return
 }
--- a/backend/plugins/org/e2e/snapshot_tables/project_mapping.csv
+++ b/backend/plugins/org/e2e/snapshot_tables/project_mapping.csv
@@ -1,0 +1,3 @@
+project_name,table,row_id,_raw_data_params,_raw_data_table,_raw_data_id,_raw_data_remark
+my_project,boards,bitbucket:BitbucketRepo:4:thenicetgp/lake,my_project,,0,
+my_project,repos,github:GithubRepo:1:1,my_project,,0,
--- a/backend/core/plugin/plugin_blueprint.go
+++ b/backend/core/plugin/plugin_blueprint.go
@@ -141,6 +141,11 @@ type MetricPluginBlueprintV200 interface {
 	MakeMetricPluginPipelinePlanV200(projectName string, options json.RawMessage) (PipelinePlan, errors.Error)
 }
 
+// ProjectMapper is implemented by the plugin org, which binding project and scopes
+type ProjectMapper interface {
+	MapProject(projectName string, scopes []Scope) (PipelinePlan, errors.Error)
+}
+
 // CompositeDataSourcePluginBlueprintV200 is for unit test
 type CompositeDataSourcePluginBlueprintV200 interface {
 	PluginMeta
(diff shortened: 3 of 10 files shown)


backend/core/dal/dal.go
```diff
@@ -239,7 +240,8 @@ func GetPrimarykeyColumnNames(d Dal, dst Tabler) (names []string, err errors.Err
 		return
 	}
 	for _, pkColumn := range pkColumns {
-		names = append(names, pkColumn.Name())
+		// in case the column name is a reserved identifier
+		names = append(names, fmt.Sprintf("%s.%s", dst.TableName(), pkColumn.Name()))
```
warren830, line 244: not clear about this, why do we need to add tableName as prefix?

backend/plugins/org/e2e/snapshot_tables/project_mapping.csv
```diff
@@ -0,0 +1,3 @@
+project_name,table,row_id,project_name,row_id,_raw_data_params,_raw_data_table,_raw_data_id,_raw_data_remark
```
warren830, line 1: why there are two row id?

backend/core/dal/dal.go
mindlesscloud: If the column name is a database reserved identifier, it would cause a database error. Take the table `project_mapping` as an example, there is a column named `table` which also happens to be a MySQL reserved word.

backend/plugins/org/e2e/snapshot_tables/project_mapping.csv
mindlesscloud: Good catch!

warren830 (approved):
LGTM

PR #4710, opened 2023-03-20
fix: add db close connection

### LINKED ISSUE #4711: [Bug][starrocks] the starrocks db connections are not released
### Search before asking

### What happened

there are many pg connections that are idle after this query. I suspect that the starrocks db connections are not released.

gorm close

### How to reproduce

use starrocks plugin

### Anything else

_No response_

### Version

main

### Code of Conduct

abeizn:
### Summary
add gorm close.

When using GORM to connect to the database, if you do not manually call the db.Close() method to close the connection, the connection will be returned to the GORM connection pool after the program is executed. At this point, the state of the connection will change to "idle", indicating that the connection is idle and available for other requests.

In GORM, if there is no connection available in the connection pool, then GORM will create a new connection. If there is a free connection in the connection pool, GORM will get a connection from the connection pool and assign it to the request.

Under normal circumstances, there should always be some idle connections in the connection pool, so as to meet the needs of concurrent requests. If the number of idle connections in the connection pool is low, it may cause requests to wait for a long time to obtain a connection, thereby affecting the performance of the application.

Therefore, although it is not necessary to explicitly call the db.Close() method to close the connection when using GORM to connect to the database, it is recommended to call this method explicitly before the program exits to ensure that the connection in the connection pool is released and avoid "idle" Too many connections cause the connection pool to fail, thereby affecting the performance of the application.

Closes #4711 

### Screenshots
img_v2_93e99a5e-a708-4c92-8bfb-8cef800b346g

### Other Information

### DIFF
--- a/backend/plugins/starrocks/tasks/tasks.go
+++ b/backend/plugins/starrocks/tasks/tasks.go
@@ -63,10 +63,22 @@ func ExportData(c plugin.SubTaskContext) errors.Error {
 	config := c.GetData().(*StarRocksConfig)
 
 	// 1. Get db instance
-	db, err := getDbInstance(c)
-	if err != nil {
-		return errors.Convert(err)
+	var db dal.Dal
+	if config.SourceDsn != "" && config.SourceType != "" {
+		o, err := getDbInstance(c)
+		if err != nil {
+			return errors.Convert(err)
+		}
+		db = dalgorm.NewDalgorm(o)
+		sqlDB, err := o.DB()
+		if err != nil {
+			return errors.Convert(err)
+		}
+		defer sqlDB.Close()
+	} else {
+		db = c.GetDal()
 	}
+
 	// 2. Filter out the tables to export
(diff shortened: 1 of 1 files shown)


backend/plugins/starrocks/tasks/tasks.go
```diff
@@ -414,6 +419,11 @@ func getDbInstance(c plugin.SubTaskContext) (db dal.Dal, err error) {
 			return nil, errors.NotFound.New(fmt.Sprintf("unsupported source type %s", config.SourceType))
 		}
 		db = dalgorm.NewDalgorm(o)
+		sqlDB, err := o.DB()
+		if err != nil {
+			return nil, err
+		}
+		defer sqlDB.Close()
```
mindlesscloud, line 426: the connection will be closed when the function `getDbInstance` returns

mindlesscloud (commented):
LGTM

mindlesscloud (approved):

PR #4715, opened 2023-03-20
fix: set _tool_jira_issue_commits.commit_sha during extracting remote…

### LINKED ISSUE #4702: [Bug][Jira] Issue with RemotelinkRepoPattern in Devlake Jira plugin (#4526)
### Search before asking

### What happened

After the addition of the RemotelinkRepoPattern parameter to the Devlake Jira plugin (#4526), it was observed that using this parameter alone leads to an empty issue_repo_commits table.

It is expected that RemotelinkRepoPattern should work alongside the existing RemotelinkCommitShaPattern parameter. Furthermore, it is expected that the data for creating the issue_repo_commits table should be sourced directly from the _tool_jira_remotelinks data rather than _tool_jira_issue_commits which currently uses the RemotelinkCommitShaPattern.

### How to reproduce

In the devlake bp configuration, only set the RemotelinkRepoPattern parameter and do not set the  RemotelinkCommitShaPattern parameter.  
Start devlake and sync jira data.  
Check the issue_repo_commits table in the database and find it empty.  

### Anything else

Recommended solution
The _tool_jira_issue_commits logic should also utilize the RemotelinkCommitShaPattern parameter, or alternatively, the data for creating the issue_repo_commits table should be sourced directly from the _tool_jira_remotelinks.

### Version

main

### Code of Conduct

mindlesscloud:
### Summary
The commits related to issues SHA were extracted twice in two different subtask, in this PR, it is simplified and only performed once in the subtask `ExtractRemotelinks`

Closes #4702


abeizn (approved):
LGTM

mappjzc (approved):

Issue #4429, opened 2023-02-15
[Feature][Testing] Integration tests POC


keon94:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

Currently the broadest automated testing that we do is e2e-testing of the plugins. This requires faking the data of datasources in CSV files. But we should also have tests that hit the actual datasources and run Devlake truly end to end.

### Description

Following https://github.com/apache/incubator-devlake/issues/4353, we can leverage the Devlake Go client to interact with the API and create projects/blueprints and run them from Go tests. The scope of this ticket will be to write at least one test suite for any plugin. There should not need to be data correctness validation for such tests - that is meant to be covered by E2Es and unit tests.

For any datasource that we choose, we'll need a test account with credentials to access it. Credentials should not be hardcoded in the test code and perhaps should be available through environment variables. See this: 
https://github.com/1Password/load-secrets-action

These tests should live inside backend/test.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

keon94:
bump

Startrekzky:
Thanks for the proposal. I have a few questions:
1. Is 'finding bugs related to bp/pipelines in the DevLake backend framework' the major benefit of this approach? 
2. What kind of bugs can be avoided or found earlier after this is in place, can you name a few issue numbers here so I can understand it correctly? 
3. Is this a replacement or complementary of the manual testing via the Config UI before each release?

keon94:
Hi @Startrekzky:
1. That would be one benefit. Another benefit would be that we'll actually be able to test our plugin collectors which are currently untested except by manual testing. As the number of plugins increases the amount of required manual testing will too which will become very infeasible.
2. Bugs related to migrations [example](https://github.com/apache/incubator-devlake/issues/3549), problematic database queries [example](https://github.com/apache/incubator-devlake/issues/4240) endpoint request issues [example](https://github.com/apache/incubator-devlake/issues/4285), and many framework level problems especially during the collection stages or server startup.
[...]

PR #4724, opened 2023-03-21
[feat-4619]: Support for running Python tests and Go-Python tests

### LINKED ISSUE #4619: [Feature][Testing] Integrate Python tests into our CICD
### Search before asking

### Use case

The Python SDK's tests currently do not work with our Github automation. This ticket will add them to the build process.

### Description

_No response_

### Related issues

_No response_

### Code of Conduct

keon94:
### What does this PR do?
1. Rewrite of the lake-builder Dockerfile (v0.0.9) to include Go 1.19.7 and Python 3.11.2, with libgit2 v1.3 support.
2. Github test actions adjusted to run Python tests and the "Remote" test for Go-Python using the Integration testing client (see #4429)
3. A few Python tests skipped because they need some rework to pass (To be done in another ticket)
4. Integration test client enhancements to work properly with Github workflows

Closes #4619 

### Screenshots

### Other Information


.github/workflows/test.yml
```diff
@@ -44,6 +44,17 @@ jobs:
       uses: actions/setup-go@v3
       with:
         go-version: ${{ matrix.go }}
+    - name: Setup Python
```
klesh, line 47: I figured you wanted a python runtime to execute the test.
Notice that we are using container to execute our test, https://github.com/apache/incubator-devlake/blob/243cc8a80aa5b37828e2a142ac9f7e3269b7e1dc/.github/workflows/test.yml#L39 , maybe `actions/setup-python@v4` doesn't support it well.

Maybe we could pack the `python` into `lake-builder` as a workaround, plus, it would be fast if we do so.

.env.example
```diff
@@ -28,7 +28,7 @@ TEMPORAL_TASK_QUEUE=
 # Debug Info Warn Error
 LOGGING_LEVEL=
 LOGGING_DIR=./logs
-ENABLE_STACKTRACE=false
+ENABLE_STACKTRACE=true
```
hezyin, line 31: Do we want to turn this on by default? I thought it's only for debugging purposes?

.github/workflows/test-e2e.yml
```diff
@@ -42,27 +42,33 @@ jobs:
           MYSQL_ROOT_PASSWORD: root
     container: mericodev/lake-builder:latest
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - run: git config --global --add safe.directory $(pwd)
```
hezyin, line 47: What does this git config command do?

hezyin (commented):
Please see my questions.

.env.example
keon94: It'll help us greatly with troubleshooting failing builds if this is on

.github/workflows/test-e2e.yml
keon94: For some strange reason, code seems to get checked out with the wrong permissions, and this would lead to VCS-stamp errors during Golang build. Couldn't understand why nor could recreate it in anyway. I assume it has to do with the new base-image being used. Adding this command fixes that problem.

hezyin (approved):
LGTM

PR #4729, opened 2023-03-21
refactor: simplify diffSync logic for jira worklog/changelog/remoteli…

### LINKED ISSUE #4638: [Refactor][jira] Simplify diffSync logic for worklog/changelog/remotelink collectors
## What and why to refactor

Previously, we relied on the `issue_updated` from worklong/changelog/remotelink tables to filter the issue ids that need to be recollected because it was the only viable way to implement diffSync at the time, to that end, we incorporated some complicated SQL:

https://github.com/apache/incubator-devlake/blob/c42a584cdb900ec09448e2a871b4407eb8c08aa4/backend/plugins/jira/tasks/worklog_collector.go#L61

However, with the support from the `api_collector_with_state`, the logic can be simplified to just utilize the `LatestSuccessStart` to filter the issues.

Get rid of the complicated SQL, filter the issues with `issue.updated_date > LatestSuccessStart`

## Related issues
#4535

abeizn:
### Summary
simplify diffSync logic for jira worklog/changelog/remotelink collectors

Closes #4638 

### Screenshots
image

### Other Information

### DIFF
--- a/backend/plugins/jira/tasks/issue_changelog_collector.go
+++ b/backend/plugins/jira/tasks/issue_changelog_collector.go
@@ -65,28 +65,18 @@ func CollectIssueChangelogs(taskCtx plugin.SubTaskContext) errors.Error {
 		return err
 	}
 
-	// query for issue_ids that needed changelog collection
 	clauses := []dal.Clause{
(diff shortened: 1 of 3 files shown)


mindlesscloud (approved):
LGTM

PR #4778, opened 2023-03-26
test: Support for remote debugging of Python src code

keon94:
### Summary
Allows us to inject breakpoints using a Pycharm debugger into Python source code. This makes debugging Python code possible.

n/a


backend/python/README.md
```diff
@@ -0,0 +1,350 @@
...
+  -d '{...connection JSON object...}'
+```
+
+You should get the created connection with his id (which is 1 for the first created connection) in the response.
+
+Now that a connection for your plugin exists in DevLake database, we can try to run your plugin using `backend/server/services/remote/run/run.go` script:
+
+```console
+cd backend
+go run server/services/remote/run/run.go  -c 1 -p python/plugins/myplugin/myplugin/main.py
```
keon94, line 333: @CamilleTeruel Should we keep this or change it to talk about using the Go integration testing client like we're using with remote_test.go? I think the run.go at this point is outdated anyway given all the recent changes (scopes, txrules, etc)

backend/python/pydevlake/README.md
```diff
@@ -1,337 +1,3 @@
...
-Now that a connection for your plugin exists in DevLake database, we can try to run your plugin using `backend/server/services/remote/run/run.go` script:
-
-```console
-cd backend
-go run server/services/remote/run/run.go  -c 1 -p python/plugins/myplugin/myplugin/main.py
-```
-
-This script takes a connection id (`-c` flag) and the path to your plugin `main.py` file (`-p` flag).
-You can also send options as a JSON object (`-o` flag).
+This code handles calls from the main Golang server code into Python plugins via IPC.
```
CamilleTeruel, line 3: The original README should stay here if we ever want to release pydevlake on pypi.

backend/python/pydevlake/pydevlake/helpers/debugger.py
```diff
@@ -0,0 +1,39 @@
...
+
+import os
+
+from pydevlake import logger
+
+
+def __start__():
+    if os.environ.get("ENABLE_PYTHON_DEBUGGER", default="false").lower() == "true":
+        try:
+            import pydevd_pycharm as pydevd
```
CamilleTeruel, line 24: Does it works only for PyCharm?

backend/python/pydevlake/pydevlake/ipc.py
```diff
@@ -24,6 +24,10 @@
 
 
 def plugin_method(func):
+    # keep this to enable debugging, and don't move this elsewhere or it can cause crashes if it gets called during the bootstrap process.
+    # noinspection PyUnresolvedReferences
+    from pydevlake.helpers import debugger
```
CamilleTeruel, line 29: Shouldn't it be commented by default (or enabled by flag)?

keon94: It is based on a flag. I put the flag check inside debugger.py to minimize boilerplate code in our source code

CamilleTeruel: Indeed, missed it

backend/python/pydevlake/pydevlake/helpers/debugger.py
CamilleTeruel: Sounds good

hezyin (approved):
LGTM

Issue #4779, opened 2023-03-27
[Bug][Core] Cron job failure while running multiple BPs


narrowizard:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

When attempting to execute multiple blueprints (BPs) using the crontab feature in DevLake, we noticed that only the first BP is being executed. Despite scheduling additional BPs, they are triggered but never run.

### What do you expect to happen

We would like the crontab feature to trigger and execute all scheduled blueprints as expected.

### How to reproduce

Schedule a series of BPs to run at specific times using the crontab feature.
Wait for the execution countdown to hit zero.
Observe that only the first scheduled BP runs and executes properly; subsequent BPs fail to start.

### Anything else

We kindly request that this issue be addressed as soon as possible to enable us to access our development data more efficiently. Thank you.  
![e8921cbf-f942-411c-a219-da0ec6ab8ab5](https://user-images.githubusercontent.com/11383928/227815207-7707983d-dbe4-4a2e-bb25-b40d9c668b2f.jpeg)

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Issue #4538, opened 2023-02-28
[Bug][Framework] Removed data scopes are still in effect in sync tasks


hezyin:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

Reported by a user in Slack: 

> When one edits the scopes attached to a blueprint and removes a data scope, the associated sync tasks, don't get removed/disabled. They keep running on the next sync

I haven't tried reproducing this issue locally, whoever picks this up please reproduce the issue first, thanks!

The user is on v0.15.1-beta5

### What do you expect to happen

See above

### How to reproduce

See above

### Version

v0.15.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

likyh:
![image](https://user-images.githubusercontent.com/3294100/223474589-f58393ec-a949-42b7-9bc4-44897c071d76.png)
![image](https://user-images.githubusercontent.com/3294100/223474826-e83f4a1c-ca86-47d1-bff3-ac3d264ff5a1.png)
I try these steps but haven't reproduced this issue:
1. add a new repo / delete a repo / replace a repo in gitlab
2. wait for the next cron run or run it manually
All these pipelines run with new repos(scopes). Do you mean the scope hasn't updated when clicking "rerun failed tasks"?
![image](https://user-images.githubusercontent.com/3294100/223475926-8f20941e-46e4-4eef-a27e-cced0c8beeba.png)
Or maybe you can add some more informations.

klesh:
Closed by #4787

PR #4787, opened 2023-03-27
fix: fix cron

### LINKED ISSUE #4761: [Bug][framework] cron and pipeline failed need more detail logs
### Search before asking

### What happened

When the cron failed we can only get the following:
image
image

But when we checked the databases, every pipeline are not padding.
It is not enough to help us understand what happened.
image

So here need more detail log if it error.

Add more detail when  it is error.

### How to reproduce

set a failed cron.

### Anything else

_No response_

### Version

main

### Code of Conduct

mappjzc:
### Summary
Fix the cron using the error blueprint to run pipelines.

Closes #4761 

### Screenshots
image

### DIFF
--- a/backend/server/services/blueprint.go
+++ b/backend/server/services/blueprint.go
@@ -43,6 +43,20 @@ var (
 	blueprintLog = logruslog.Global.Nested("blueprint")
 )
 
+type BlueprintJob struct {
+	Blueprint *models.Blueprint
+}
+
+func (bj BlueprintJob) Run() {
+	blueprint := bj.Blueprint
+	pipeline, err := createPipelineByBlueprint(blueprint)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


likyh:
Why did this issue happen?

likyh (approved):

klesh:
> Why did this issue happen?

It happened because of the `Variable Scope`, the `AddFunc` is nested inside a `for loop` which effectively are all using the same `blueprint` to create new `pipelines`, when multiple cronjobs get triggered, the first one would succeed but the rest would fail because running multiple pipelines of the same bp is not allowed:

https://github.com/apache/incubator-devlake/blob/1e2b351f4cd2ba86e913c3f11159f409e62718cc/backend/server/services/pipeline_helper.go#L44

This should close the #4779 and #4538 as well

PR #4789, opened 2023-03-27
fix: test teambition connection

coldgust:
### Summary
fix teambition set the wrong code when test connection

Closes xx

### Screenshots

### Other Information

### DIFF
--- a/backend/plugins/teambition/api/connection.go
+++ b/backend/plugins/teambition/api/connection.go
@@ -70,7 +70,8 @@ func TestConnection(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, 
 		return nil, err
 	}
 	if resBody.Code != http.StatusOK {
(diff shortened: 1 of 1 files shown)


klesh:
Hi, @coldgust , thank you for the fix.

Please describe why you think this is the right way, did you find it out on their document? if so, please attach a link to the document, thanks.
Please attach screenshots to show the new code is working correctly.

klesh:
Hi, @coldgust . Thank you for the information, 👍
Seems like their API design is a little bit off.
However, there is no proof to show the `code` inside the response body is `Http Status Code`, we should hardcode our API status code to 400 unless it can be proven.

coldgust:
@klesh , thanks for the reminder.
I think `errorMessage` is helpful for users too.

klesh (approved):
LGTM

coldgust:
#5066 have solved this issues.

PR #4802, opened 2023-03-28
Fix failing python plugin tests

CamilleTeruel:
### ⚠️ Pre Checklist

### Summary
Fix failing skipped test in python_plugin_test.go


backend/server/services/remote/models/conversion.go
```diff
@@ -118,8 +119,10 @@ func getGoType(schema map[string]any) (reflect.Type, errors.Error) {
 	case "boolean":
 		goType = reflect.TypeOf(false)
 	case "string":
-		//TODO: distinguish stypes based on string format
+		//TODO: distinguish types based on string format
```
keon94, line 122: I think we can remove this TODO

CamilleTeruel: Done

backend/core/utils/ipc.go
```diff
@@ -260,7 +260,8 @@ func scanErrorPipe(pipe io.ReadCloser, onReceive func([]byte), outboundChannel c
 			data := make([]byte, len(src))
 			copy(data, src)
 			onReceive(data)
-			outboundChannel <- &ProcessResponse{stderr: data}
+			err := errors.Default.New(string(data))
```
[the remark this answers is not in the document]
CamilleTeruel: Oh right.
I want to make sure that a command is considered a failure whenever python-side calls `logger.error(...)`.
I made the change in `cmd.go:Stream` and `cmd.go:Call` instead.

backend/server/services/remote/bridge/cmd.go
```diff
@@ -74,6 +74,12 @@ func (c *CmdInvoker) Call(methodName string, ctx plugin.ExecContext, args ...any
 			Err: errors.Default.Wrap(err, fmt.Sprintf("failed to invoke remote function \"%s\"", methodName)),
 		}
 	}
+	errMsg := response.GetStderr()
```
keon94, line 77: What would be a scenario where Python would only log an error and we want things to fail on Go side, as opposed to throwing an Exception from Python?

backend/core/utils/ipc.go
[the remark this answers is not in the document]
CamilleTeruel: Well depending on the semantics you put under the term "error", you might want an error that doesn't abort the current command.
This was mostly for convenience and to ensure we have a single failure mode, so not an important change. 
Removed.

keon94 (approved):

PR #4833, opened 2023-03-31
feat: customize plugin support float and integer extraction

### LINKED ISSUE #4828: [Bug][customize] Error mapping null customfield values to domain layer issues table
### Search before asking

### What happened

When using the Customize plugin to map the custom field (customfield_10024) from the _raw_jira_api_issues to the 'issues' table in the domain layer, we encountered an issue. We found out that if the value of customfield_10024 in _raw_jira_api_issues is null, the Customize plugin throws an error while mapping process. The error message is: "invalid input syntax for type double precision: "" (SQLSTATE 22P02)". It should be noted that the target field we are mapping to is of type float.

We expect the Customize plugin to handle null values when mapping customfield_10024 to the 'issues' table in the domain layer. Instead of throwing an error, the plugin should leave the target field's value as null.

### How to reproduce

Ensure that the Customize plugin is configured to map the customfield_10024 from _raw_jira_api_issues to the 'issues' table in the domain layer.
In _raw_jira_api_issues, ensure that the value for customfield_10024 is null.
Run the mapping process.
Observe the error message thrown by the Customize plugin.

### Anything else

image

### Version

main

### Code of Conduct

mindlesscloud:
### Summary
This pull request adds support for extracting both float and integer data types. Users will now be able to retrieve numeric values in a wider range of formats, improving the flexibility and usefulness of our DevLake

Closes #4828


abeizn (approved):
LGTM

Issue #4869, opened 2023-04-06
[Refactor][dora] Change Lead Time doesn't include Jenkins deployment and other data correctness issues


klesh:
## What and why to refactor
1. #4825 indicates that deployment might be a `cicd_pipeline` or `cicd_task` from a user's perspective. Therefore, we need to broaden the scope of the "Deployment Regex" to include the `cicd_pipeline`.
2. Jenkins deployments are not included in Change Lead Time calculation. The `cicd_pipeline_commits` consist of the repositories and commit_shas being built or deployed by a `cicd_pipeline`, we currently rely on the `repo_id` field to distinguish different repos, for data sources like github/gitlabe, filling it is not a issue. However, for specialized tool like Jenkins, it may not be possible, reliable, or easy to provide such information. In such cases, the only dependable field would be the `repo_url`. All data source plugins must be updated to populate this field.
3. There are many ways to orchestrate a deployment, making it challenging to identify all possible scenarios. We have determined that a `cicd_deployment_commits` (new table) is a `cicd_pipeline_commit` if the corresponding `cicd_pipeline` satisfies `type=deployment` or includes any `cicd_tasks` with `type=deployment`.
4. Currently, both `refdiff` and `dora` are identifying previous deployments separately and using different methods, which can lead to issues.
5. To determine the deployment_commit of a pull request, we must identify the commits deployed by examining its preceding SUCCESSFUL deployment. To do this, we must group the deployments by `cicd_scope_id`, `env`, and `repo_url`, then sort them by their `started date`.

## Describe the solution you'd like
![Flowchart](https://user-images.githubusercontent.com/61080/233818262-dbd0d56f-0750-421b-bdf3-dd00294aa952.jpg)

## Related issues
#4825 
#4810

keon94:
@klesh Should this not be for the v0.18 milestone and cherry-picked back to v0.17, or are we delaying v0.17 until this gets done?

klesh:
@keon94 It ought to be fixed ASAP

hezyin:
Hi @klesh, I synced up the refactor plan with @keon94 and @CamilleTeruel today. They were on board with the plan and gave the following feedback:

1. It seems a bit unnatural for the refdiff plugin to generate deployments. That sounds more like the responsibility of the DORA plugin. Can you think of a solution that moves that logic out of the refdiff plugin?
[...]

PR #4886, opened 2023-04-10
Refactor Chage Lead Time - Step 2: enrich type/env for pipelines and tasks

klesh:
### Summary

1. added `TryAdd/ReturnNameIfMatched` to `RegexEnricher` to simplify the usage of the helper
2. data source plugins(bitbucket in this PR) would use `RegexEnricher` to enrich the `type` and `environment` fields for `cicd_pipeline`/`cicd_task`. 
3. `type`/`env` enrichment are moved into `extractors`,  `pipeline_extractor` would run the RegexEnricher against the `pipeline.ref_name` since bitbucket pipelines don't have names. 
4. next step: `dora.generateDeployment` would use `type`/`environment` fields to generate deployment

This approach covers `pipeline`  like [this one](https://bitbucket.org/zhenmianws/helloworldrepo/src/master/bitbucket-pipelines.yml), assuming there is a `production` branch for deploying the repo to the production environment.

Part of #4869 

### Screenshots

when `deploymentPattern=main` and `productionPattern=pipeline`, the `_tool_bitbucket_pipelines` table would produces the result like the following screenshot
image


backend/plugins/bitbucket/tasks/pipeline_extractor.go
```diff
@@ -98,6 +100,8 @@ func ExtractApiPipelines(taskCtx plugin.SubTaskContext) errors.Error {
 				DurationInSeconds:   bitbucketApiPipeline.DurationInSeconds,
 				BitbucketCreatedOn:  api.Iso8601TimeToTime(bitbucketApiPipeline.CreatedOn),
 				BitbucketCompleteOn: api.Iso8601TimeToTime(bitbucketApiPipeline.CompletedOn),
+				Type:                data.RegexEnricher.ReturnNameIfMatched(devops.DEPLOYMENT, bitbucketApiPipeline.Target.RefName),
```
warren830, line 103: I think if would be better do this on domain layer, as some data source platform might have its own `type` and `env`

backend/helpers/pluginhelper/api/enrich_with_regex.go
```diff
@@ -64,3 +67,33 @@ func (r *RegexEnricher) GetEnrichResult(pattern string, v string, result string)
...
+	regex, err := errors.Convert01(regexp.Compile(pattern))
+	if err != nil {
+		return errors.BadInput.Wrap(err, fmt.Sprintf("Fail to compile pattern for regex pattern: %s", pattern))
+	}
+	r.regexpMap[name] = regex
+	return nil
+}
+
+// ReturnNameIfMatched will return name if any of the targets matches the regex with the given name
+func (r *RegexEnricher) ReturnNameIfMatched(name string, targets ...string) string {
```
warren830, line 88: Nice!

backend/plugins/bitbucket/tasks/pipeline_extractor.go
klesh: I'm confused, shouldn't we do it on the tool layer since "some data source platform might have its own type and env" ?

backend/helpers/pluginhelper/api/enrich_with_regex.go
mindlesscloud, line 88: After the refactoring, the `RegexEnricher` is more like a utils function, and has nothing to do with enricher, can you remove `enricher` from its name?

likyh (approved):

backend/helpers/pluginhelper/api/enrich_with_regex.go
klesh: Alright, will do

PR #4907, opened 2023-04-12
fix: dora refdiff and change lead time logic

abeizn:
### Summary
update dora refdiff and change lead time logic, 

refdiff add filter: 
ct.environment =  "PRODUCTION" 
ct.result = "SUCCESS"
pm.table = "cicd_scopes"

change lead time:
finished_commits_diffs left joincicd_pipeline_commits If the commit has the same data, the same commit with left join cicd_tasks will be associated with multiple cicd_task ids, resulting in more data, so change lead time keep the same logic.

```
select 
distinct ct.id as task_id, fcd.new_commit_sha, fcd.old_commit_sha, ct.finished_date as task_finished_date, cpc.repo_id as repo_id, ct.started_date
-- ct.id as task_id, cpc.commit_sha as new_deploy_commit_sha, ct.finished_date as task_finished_date, cpc.repo_id as repo_id
from
finished_commits_diffs fcd
left join cicd_pipeline_commits cpc on fcd.old_commit_sha = cpc.commit_sha
left join cicd_tasks ct on ct.pipeline_id = cpc.pipeline_id
left join project_mapping pm on pm.row_id = ct.cicd_scope_id
where 
pm.project_name = "github1" and
ct.environment = "PRODUCTION"
```
image

NA

### Screenshots
image

### Other Information


mindlesscloud (approved):
LGTM

PR #4491, opened 2023-02-22
feat: add increment support for graphql collector

likyh:
Summary
Rename all CreatedDateAfter to TimeAfter for github_graphql.

Does this close any open issues?
Part of https://github.com/apache/incubator-devlake/issues/4403

PR #4921, opened 2023-04-13
fix: move extract existing graphql data to general position

likyh:
### Summary
move extract existing graphql data to general position

Closes #4491

### Screenshots
image


abeizn (approved):
LGTM

PR #4926, opened 2023-04-13
feat(framework): backend support cognito

warren830:
### Summary
Modify backend to support cognito

Closes xx

### Screenshots
login success
image

get blueprints with correct authorization:
image

get blueprints with wrong authorization:
image

get blueprints with no authorization:
image

### Other Information


backend/server/services/cognito.go
```diff
@@ -0,0 +1,187 @@
+/*
```
keon94, line 1: I'd define a new package called aws, and put all aws-facing code in there. It's more organized that way

klesh: good call.
I would prefer the package to be called `auth`, and it could offer multiple auth mechanisms in the future, including `cognito` and others like `OAuth` and `SAML`

warren830: OK, will fix it in the next pr, as I just created a new pr for frontend to support cognito

PR #4932, opened 2023-04-14
fix: the jql time zone issue

### LINKED ISSUE #4902: [Bug][Jira] System Timezone not Respected
### Search before asking

### What happened

I am testing the Jira plugin and have found that the incremental differential feature does not take the timezone into account. You can see in the following log line that the Jira plugin is looking in the future for Jira issues:

```
time="2023-04-11 13:51:26" level=debug msg=" [pipeline service] [pipeline #2] [task #2] [collectIssues] fetchAsync >>> done for agile/1.0/board/1302/issue map[expand:[changelog] jql:[updated >= '2023/04/11 20:20' AND updated >= '2022/10/11 00:00' AND created is not null ORDER BY created ASC] maxResults:[100] startAt:[0]] <nil>"
```
The the time of the log line was 2023-04-11 13:51:26 while the previous pipeline was executed at 2023/04/11 20:20 which is in the future. The pod time zone is set to America/Los_Angeles, but the Jira plugin is using UTC.

I expect the Jira plugin to use the same timezone as the system along with the same time zone of the lake log timestamps.

### How to reproduce

I had to set the logging level to debug to see the Jira queries where the timestamp is given. I set the log level and timezone using the helm chart flags below:

```
--set "lake.loggingLevel=debug" --set "option.localtime=/usr/share/zoneinfo/America/Los_Angeles"
```

The timezone of my Jira instance is America/Los_Angeles so for someone to test they need to set the pod timezone to a different timezone than the Jira instance. In order to find the Jira server time you can use the HTTP API endpoint `/rest/api/2/serverInfo`.

### Anything else

This issue will cause an inconsistency between ticket changes in Jira and the metrics being stored in MySQL. I just happened to run across the issue because I was making ticket changes and they were not showing up.

The version I'm running is 0.16.0-beta11

### Version

main

### Code of Conduct

mindlesscloud:
## Problem:
A user reported a timezone-related issue with a Jira plugin: https://github.com/apache/incubator-devlake/issues/4902. The most important information is a log that the user provided, which includes a timestamp and some debug information.
```text
time="2023-04-11 13:51:26" level=debug msg=" [pipeline service] [pipeline #2] [task #2] [collectIssues] fetchAsync >>> done for agile/1.0/board/1302/issue map[expand:[changelog] jql:[updated >= '2023/04/11 20:20' AND updated >= '2022/10/11 00:00' AND created is not null ORDER BY created ASC] maxResults:[100] startAt:[0]] <nil>"
```
## Reason:
There are two issues with the log: 
1.  the "updated" field appears twice in the JQL query
2.  the log timestamp is earlier than the "updated" timestamp

```go
if data.TimeAfter != nil {
   jql = fmt.Sprintf("updated >= '%v' AND %v", data.TimeAfter.Format("2006/01/02 15:04"), jql)
}

incremental := collectorWithState.IsIncremental()
if incremental {
   jql = fmt.Sprintf("updated >= '%v' AND %v", collectorWithState.LatestState.LatestSuccessStart.Format("2006/01/02 15:04"), jql)
}
```
Regarding the first issue, the root cause can be found in the code: the "updated" field is included twice because of some conditional logic that checks if a certain variable is not nil.

The second issue is a bit more complex. The timestamp in the log comes from the variable "collectorWithState.LatestState.LatestSuccessStart", which is written to a database in UTC format by the GORM library. However, when this timestamp is read from the database and formatted into a string, the timezone information is lost. This can result in a timestamp that is different from the actual time, especially if the user is in a different timezone.

## Solution:
For the first issue, the solution is to add a condition that checks which of the two "updated" fields is later and use that one.

For the second issue, there are several possible solutions, but each has some drawbacks. One solution is to convert the timestamp to the user's local time before formatting it. However, this can lead to delays or data loss if the user's timezone is different from Jira's timezone. Another solution is to subtract a day from the timestamp before formatting it, but this can result in some data being collected twice. In this PR, we take the second one as our final solution.

Closes #4902

### DIFF
--- a/backend/plugins/jira/tasks/issue_collector.go
+++ b/backend/plugins/jira/tasks/issue_collector.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/apache/incubator-devlake/core/errors"
 	"github.com/apache/incubator-devlake/core/plugin"
@@ -66,18 +67,8 @@ func CollectIssues(taskCtx plugin.SubTaskContext) errors.Error {
 	// build jql
 	// IMPORTANT: we have to keep paginated data in a consistence order to avoid data-missing, if we sort issues by
 	//  `updated`, issue will be jumping between pages if it got updated during the collection process
-	jql := "created is not null ORDER BY created ASC"
-
-	// timer filter
-	if data.TimeAfter != nil {
-		jql = fmt.Sprintf("updated >= '%v' AND %v", data.TimeAfter.Format("2006/01/02 15:04"), jql)
-	}
-
-	// diff sync
 	incremental := collectorWithState.IsIncremental()
-	if incremental {
-		jql = fmt.Sprintf("updated >= '%v' AND %v", collectorWithState.LatestState.LatestSuccessStart.Format("2006/01/02 15:04"), jql)
-	}
+	jql := buildJQL(data.TimeAfter, collectorWithState.LatestState.LatestSuccessStart, incremental)
 
 	err = collectorWithState.InitCollector(api.ApiCollectorArgs{
 		ApiClient:   data.ApiClient,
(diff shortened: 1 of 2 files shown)


abeizn (approved):
LGTM

PR #4959, opened 2023-04-19
feat: enrich cicd_pipeline type/env for jenkins

klesh:
### Summary
Previously cicd_pipelines that converted from jenkins builds didn't fill in the type/env fields.
This PR would [do so](https://github.com/apache/incubator-devlake/blob/93bece619450a51ddf50c927706c8ffe7bf5fe78/backend/plugins/jenkins/tasks/build_cicd_convertor.go#L105), along with other small refactors

Related to #4869


abeizn (approved):
LGTM

PR #4976, opened 2023-04-19
4879 nested property extraction with json pointer

### LINKED ISSUE #4879: [Feature][pydevlake] Declarative nested property extraction via JSON pointer
### Search before asking

### Use case

As a python plugin developer, I want to declare how a tool model attribute maps to a possibly nested JSON property in order to not having to write extraction logic.

### Description

We should be able to annotate tool models attributes with a JSON-pointer expression that tell how to populate this attribute value from the input JSON.

### Related issues

_No response_

### Code of Conduct

CamilleTeruel:
### Summary

Add support for declarative property extraction of tool model fields.

```python
from pydevlake import ToolModel, Field

class MyModel(ToolModel):
    id: int = Field(primary_key=True)
    version: str = Field(source='/version/number') # Will extract value found at path /version/number in raw JSON
```

Closes #4879


backend/python/pydevlake/pydevlake/extractor.py
```diff
@@ -15,15 +15,45 @@
...
-                json[key] = value["id"]                    
-    return model_cls(**json)
+    """
+    Automatically extract a tool model from a json object.
+    The tool model class can define fields with a source argument to specify the JSON pointer (RFC 6901) to the value.
+
+    Example:
+        class DummyModel(ToolModel):
+            name: str
+            version: str = Field(source='/version/number')
```
keon94, line 32: how would we specify array paths with this scheme? e.g. "/a/b/c[index]/d/e".

backend/python/plugins/azuredevops/azuredevops/helper/db.py
```diff
@@ -1,28 +0,0 @@
-# Licensed to the Apache Software Foundation (ASF) under one or more
```
keon94, line 1: has this been logic moved elsewhere, or eliminated altogether? I think it's useful to have a helper class for DB querying and joining of models (for converter purposes) that also wraps sqlalchemy-specific syntax

backend/python/pydevlake/pydevlake/extractor.py
keon94: I guess the var itself would have to be an array too: e.g.
`some_var: []str = Field(source=`/a/b/c[]/d/e`)

CamilleTeruel: A JSON pointer points to a single element. You can point to a specific array element by index (e.g. /a/b/c/3/e). 
You can't project a nested property to each array member, like `/a/b/c/*/e` with JSON pointer. See RFC6901 for details. Note that a list attribute of a ToolModel would have to be implemented as a Relation to another table, not as simple column Field.

backend/python/plugins/azuredevops/azuredevops/helper/db.py
CamilleTeruel: I deleted this file because it is not used anymore.

Some converters might indeed need to query the DB.
While it might be beneficial to abstract sqlalchemy API away, this requires quite some work. OTOH, sqlalchemy API is mature and almost standard. So I'm not sure if the cost/benefit ratio is good.

We can think about little helping methods like this one, but I think it should execute the query in the same session than the one used for  conversion.

hezyin (approved):
LGTM

PR #4997, opened 2023-04-21
fix: case inconsistency in sql

### LINKED ISSUE #4987: [Bug][Dashboard] Different versions of mysql are case-sensitive, so the dashboard may not display the data
### Search before asking

### What happened

Different versions of mysql are case-sensitive, so the dashboard may not display the data.

mysql 8.0.33:
image
image

mysql 8.0.26
image
image

**Due to the performance of different versions of mysql, our dashboard display may have problems, so we need to unify the values in the dashboard into the original data style.**

1. issue.type="incident" :

image

2. issue.type="INCIDENT" :
image

Different versions of mysql can display the same data.

### How to reproduce

Use a different mysql version

### Anything else

_No response_

### Version

main

### Code of Conduct

abeizn:
### Summary
fix case inconsistency in sql

Closes #4987 

### Screenshots
image

### Other Information


Startrekzky (approved):
LGTM. I checked all occurrences that matched the pattern `'[a-zA-Z][a-z]*'` in all dashboards.

PR #5028, opened 2023-04-25
fix: deployment_commit_generator produced unwanted records

klesh:
### Summary

Problem: the record in the following screenshot was produced by the `dora.deployment_commit_generator` which is incorrect because it doesn't meet the criteria
image

Why: this is caused by the incorrect SQL condition:
image
The `t.type = p.type` and the `OR` operator are used in conjunction which is incorrect


Startrekzky (approved):
LGTM

PR #5030, opened 2023-04-25
Github workflow run/job doesn't end up in cicd_deployment_commits table with correct Regex

klesh:
### Summary
 
for the `apache/incubator-devlake` repo with the following `transformation settings`
image

we are expecting to see [this github action](https://github.com/apache/incubator-devlake/actions/runs/4782906377) to be appear in `cicd_deployment_commits` table.

However, it didn't happen because the essential information for `cicd_deployment_commits` record to be generated was missing.

1. the `type` and `env` fields were not loaded from the database
2. during the investigation, I found out the `github_graphql` doesn't enrich the `type` / `env` as well, it is not the direct cause of the above problem. but it clearly a bug, might fix it as well.

### Screenshots
With this fix, the `cicd_deployment_commits` can be produced correctly:
image

### DIFF
--- a/backend/plugins/github/tasks/cicd_run_convertor.go
+++ b/backend/plugins/github/tasks/cicd_run_convertor.go
@@ -18,6 +18,9 @@ limitations under the License.
 package tasks
 
 import (
+	"reflect"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


warren830 (approved):
LGTM

Issue #5031, opened 2023-04-25
[Bug][PagerDuty] Data cannot be collected completely


warren830:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

First try, we lost the most recent record. 
After fix the first problem, I found that we cannot collect new data after first collection

I think the problem comes from pagerduty api itself. Previously, when we used the "since" parameter, data would not be returned if the time distance between "since" and the present was over six months. Now, data will not be returned if the time distance is over one month.

### What do you expect to happen

should collect correctly

### How to reproduce

use pagerduty plugin to collect data

### Version

main

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

warren830:
Seems like, the duration cannot exceed 1 month

keon94:
@warren830 please test out this PR: https://github.com/apache/incubator-devlake/pull/5042

PR #5032, opened 2023-04-25
fix(pagerduty): fix offset

warren830:
### Summary
right now, pagerduty use reqData.Pager.Page as offset when sending reqeusts.
This will cause data loss. This pr use reqData.Pager.Skip as offset.

part of #5031 

### Screenshots

### Other Information


mindlesscloud (approved):
LGTM

Issue #4296, opened 2023-02-02
[Bug][UI]gitlab接入时，增加scope时，获取数据group id和 project id相同时，前端异常


joelaugh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

<img width="687" alt="image" src="https://user-images.githubusercontent.com/96901838/216280791-34d2639a-8ce6-4c03-bca2-28425127ce6c.png">

![image](https://user-images.githubusercontent.com/96901838/216280868-71270e1e-56dc-47a3-a7ad-a70761b61caf.png)

![image](https://user-images.githubusercontent.com/96901838/216280897-59c15685-bd97-4c9b-b6ab-e50bcd8b12c6.png)

### What do you expect to happen

不应该发生异常

### How to reproduce

gitlab 创建group与project id相同的数据，且在创建scope里二者都出现

### Version

v0.15.0

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

mappjzc:
thanks for your issue.
This looks like a problem with config-ui.
@mintsweet can you help me have a look.

mintsweet:
it was fixed on https://github.com/apache/incubator-devlake/releases/tag/v0.15.1-beta1

klesh:
@joelaugh Hi, you may update the config-ui image tag to `v0.15.1-beta1`  to see if the problem solved

likyh:
> @joelaugh Hi, you may update the config-ui image tag to `v0.15.1-beta1` to see if the problem solved

That's a new bug on `v0.15.1-beta3` and only found on  gitlab v11/12

PR #5037, opened 2023-04-26
fix: fix config ui do not support same group and project id

mappjzc:
### Summary
Because the request was changed
The fix #4307 on config-ui does not work.
So it needs to be fixed on the backend.
Rename the group id, and add `group:` before the id

related #4296 

### Screenshots
image
image


mintsweet (approved):
LGTM.

PR #5042, opened 2023-04-26
[fix-5031]: Pagerduty created after time fix

### LINKED ISSUE #5031: [Bug][PagerDuty] Data cannot be collected completely

keon94:
Closes #5031

### Screenshots

### Other Information


warren830:
image
Seems like it's working now

warren830 (approved):
LGTM

Startrekzky:
@keon94 Please cherrypick this to release-v0.17 if needed. We'll automate this process in the future.

PR #5043, opened 2023-04-26
fix: fix missing id in tapd transformationRule migration

likyh:
### Summary
fix missing id in tapd transformationRule migration.

### Screenshots
The bug:
img_v2_089ce6c0-8859-46db-9218-680d70e86f3g

After fix, migration can run~
image

After run:
image
Although id is not primary key if the table transformation rule exist, I think it's no matter. Because this script haven't publish to users.


mindlesscloud (approved):
LGTM

PR #5066, opened 2023-04-28
fix: unauthorized return 400 instead of 401

### LINKED ISSUE #5068: [Bug][config-ui] config-ui reset login when TestConnection failed and return 401
### Search before asking

### What happened

Because the TestConnection returns the status code 401,
The Config-ui will reset the login state.

It made the user need to rewrite the username and password to config-ui again and again in each TestConnection 401 code return.
image

This is a very bad experience for users.

The user needn't rewrite there username and password to config-ui anymore after the TestConnection failed.

### How to reproduce

test any connection with wrong token.

### Anything else

_No response_

### Version

main

### Code of Conduct

mappjzc:
### Summary
Because of the 401 error will reset the config-UI login.
So It changed all 401 to 400.

Closes #5068 

### Screenshots
image
image
image
image
image
image
image
image
image

### DIFF
--- a/backend/plugins/teambition/api/connection.go
+++ b/backend/plugins/teambition/api/connection.go
@@ -20,13 +20,14 @@ package api
 import (
 	"context"
 	"fmt"
+	"net/http"
+
 	"github.com/apache/incubator-devlake/core/errors"
 	"github.com/apache/incubator-devlake/core/plugin"
 	"github.com/apache/incubator-devlake/helpers/pluginhelper/api"
 	"github.com/apache/incubator-devlake/plugins/teambition/models"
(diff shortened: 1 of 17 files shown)


abeizn (approved):
LGTM

backend/plugins/teambition/api/connection.go
```diff
@@ -70,7 +74,10 @@ func TestConnection(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput,
 		return nil, err
 	}
 	if resBody.Code != http.StatusOK {
```
likyh, line 76: why do you check `resBody.Code != http.StatusOK twice` ?

PR #5074, opened 2023-05-02
Fix empty tx rules patterns

### LINKED ISSUE #5072: [Bug][Azure DevOps] Records are inserted into cicd_deployment_commits even when regexes don't match
### Search before asking

### What happened

**Version v0.17.0-alpha7**

## Azure DevOps Setup

I imported https://github.com/keon94/cicd-playground into my Azure DevOps account and created a pipeline using the definition at https://github.com/keon94/cicd-playground/blob/main/azure-pipelines.yml. I ran the pipeline twice manually on the same commit:

Screen Shot 2023-04-28 at 9 48 50 PM

## How to reproduce this bug

1. I first collect the Azure repo created above without setting any transformation rule. Then I check cicd_deployment_commits table and it's empty. This aligns with my expectation.
Screen Shot 2023-04-27 at 9 28 22 PM
Screen Shot 2023-04-27 at 9 28 12 PM

2. I then collect the Azure repo again with a transformation rule this time, but I intentionally set the wrong regex ("jooooob") for deployment regex and leave the production regex empty. However, when I check the cicd_deployment_commits table, there're two records this time. Since the deployment regex doesn't match any job names or pipeline names, I expect this table to remain empty after this collection.
Screen Shot 2023-04-28 at 9 56 06 PM

3. When I check the Azure DevOps transformation rules in the database, I notice that the deployment and production regex columns are empty.
Screen Shot 2023-04-28 at 8 51 48 AM

See above

### How to reproduce

See above

### Anything else

_No response_

### Version

main

### Code of Conduct

CamilleTeruel:
### Summary

Fix empty transformation rule fields in db caused by a casing mismatch.
@hezyin 

Closes #5072

### DIFF
--- a/backend/python/plugins/azuredevops/azuredevops/streams/pull_request_commits.py
+++ b/backend/python/plugins/azuredevops/azuredevops/streams/pull_request_commits.py
@@ -30,7 +30,7 @@ class GitPullRequestCommits(Substream):
     def collect(self, state, context, parent: GitPullRequest) -> Iterable[tuple[object, dict]]:
         repo: GitRepository = context.scope
         azuredevops_api = AzureDevOpsAPI(context.connection)
-        response = azuredevops_api.git_repo_pull_request_commits(repo.org_id, repo.project_id, parent.repo_id, parent.id)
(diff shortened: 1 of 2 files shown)


hezyin (approved):
LGTM

Issue #4801, opened 2023-03-28
[Feature][Plugins] Request for GitHub Integration Using GitHub Apps


mtahle:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

As an SRE, I am concerned about the security and flexibility of our project's integration with GitHub. With this feature, we can manage permissions more efficiently, reduce security risks, and enhance the integration's functionality.

### Description

Integrating with GitHub using GitHub Apps instead of personal access tokens for better security and more flexibility.

### Related issues

#4537

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Startrekzky:
Thanks for the proposal, @mtahle

PR #5077, opened 2023-05-03
feat(github plugin): adds authentication via github app

mkmarek:
### ⚠️ Pre Checklist

### Summary
This PR adds a new option to use authentication via GitHub App while creating a new GitHub connection. Then it alters several places in `github` and `github_graphql` plugins to fetch the appropriate token given an installation id, which it can fetch based on the repository name and owner.

Closes https://github.com/apache/incubator-devlake/issues/4801

### Screenshots

GitHub app configuration section:
image

In the repository selection the user/org column will represent the orgs/users where the app is installed and the right side are repositories allowed by that installation:
image

### Other Information
I'm doing this change on behalf of Telia Company (https://github.com/telia-company). I'm not sure if they sent in CCLA, but I'm trying to find out.

I haven't been around this codebase for very long (last two weeks) so any recommendations for changes, what tests to add, or hints on what to do next are very much appreciated.


klesh:
Hi, @mkmarek , Thank you so much for the contribution.
Can you also provide the `permissions` required for the Github App so I can test it locally?

The current implementation uses `if` in many places for the `MultiAuth` to work which is not ideal IMO,
Ideally, the only file that we need to modify would be `backend/plugins/github/models/connection.go` and other dirs/files like `backend/plugins/github/api` and `backend/plugins/github/impl` should be left intact.
 
Would you mind taking a look at `jira` implementation`? https://github.com/apache/incubator-devlake/blob/main/backend/plugins/jira/models/connection.go#L42
The idea is to delegate authentication via the `MultiAuth` struct and isolate the authentication from other places.

To my understanding, the Github App Authentication works pretty much like the `feishu` plugin, I suggest we do it like this:
1. Alias the `helper.AppKey` as ` GithubAppKey` and implement the `GithubAppKey.SetupAuthentication` for each request and/or `GithubConnection.PrepareApiClient` if some kind of global initialization is required. (ps: an `if` is needed here or maybe we could delegate the logic to the  `Authenticator` level as well, it is an open subject that we can discuss)
2. Embed the `helper.MultiAuth` and `GithubAppKey` into `GithubConn` just like how `jira` did it.
3. Revert the changes to the `api` and the `impl` folders.

klesh:
If I understand it correctly:
1. fetch the `installationId` by `owner/repo`
2. authentication `app installation` using the `appId`, `secretKey` and `installationId` which produce the `token`
3. uses the `token` for the rest requests

We can not utilize the `PrepareApiClient` because it works on the `connection` level and the `owner/repo` information is missing from the signature.

The problem is: Github App Installation Authentication works on the `repo` level, is that correct?

mkmarek:
Yeah pretty much. It's possible to fetch the installation ID based on `owner/repo`. It's also possible to fetch all installations and then pick the correct one just based on `owner`. Or if you know if the `owner` is either org or user there are endpoints for fetching the token directly just based on `owner`.

I have made an alternative version of the code. I haven't cleaned it up properly yet. But I can push it now. It removes the if-elses in the impl and api parts and moves everything to `connection.go` using multiauth. What I got left with in the `impl` and `api` files is the minimum that is needed to provide for the connection to work.

I took a bit different approach where it initially fetches tokens for all the installations in the `PrepareApiClient` function and then uses the tokens in `SetupAuthentication` based on either a installationid or a repository. The installationid and the repository have to be provided from the outside.

klesh:
@mkmarek Thanks for the clarification. It is much cleaner now.
I'm wondering if we can do better, the current design of the `MultiAuth` is flawed and fail to support such a scenario.
It would be nice if we could improve it and keep other files intact.
I would need some time to think about it, or we could merge your work first if it is urgent from your side.

klesh:
One more thing, I noticed that Git App has [Authentication on behalf of a user](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/about-authentication-with-a-github-app#authentication-on-behalf-of-a-user), will it be possible to do it that way? @mkmarek

mkmarek:
I don't think it's that urgent from our side. Until this feature gets in, we can build our own docker images to test things out, and when it gets in, we can switch back to your releases. So there's plenty of time to implement this the right way.

About authenticating on behalf of user. I'm not sure that would fit the same use case. It's used more for CLIs or webpages where you can "login with GitHub". The flow is very different there. In our use case Github App replaces machine users we'd need to have.
With auth on behalf of user we'd need to have either both the app and the machine users or only the app and have real users in the organization go into devlake and perform auth there.. and in both of those cases the existing token solution would be probably better.

klesh:
@mkmarek  So, installation is on the owner/organization level? Can I use it to select multiple repos under the same owner/org?

klesh:
I see, LGTM.
@Startrekzky @hezyin What do you think

Startrekzky:
Hi @mkmarek , can you solve the conflicts and check if the commit messages follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)?
image

klesh:
@mkmarek We all believe that the limitation is acceptable, would you like to rebase your PR and we can get it to merge ASAP

mkmarek:
@klesh @Startrekzky I did a rebase, fixed all conflicts, adjusted the commit message, and retested. Now it should be good to go.

klesh (approved):
BE looks good to me

klesh:
Hi, @mkmarek , I'm trying to verify if it is working correctly but am stuck at the very first step 🥲 "Register new Github App", the options are overwhelming, I'm not sure how to set up correctly.
Would you like to write a document to guide us through the whole process? Thank you very much.
I think we can add a section to https://github.com/apache/incubator-devlake-website/blob/main/docs/Configuration/GitHub.md to describe how to setup the Github App and use it for data collection.

PR #5057, opened 2023-04-27
4986 refactor plugin registration

CamilleTeruel:
### Summary

Refactor remote plugin registration to not rely on /plugin/register endpoint.
Open API spec generation is moved from pydevlake to remote plugin framework on go side.

### Does this close any open issues?
Closes #4986 and #5056 

@keon94 @klesh

PR #5078, opened 2023-05-03
fix: Correct Swagger spec template path discovery

keon94:
### Summary
Error seen on server startup after https://github.com/apache/incubator-devlake/pull/5057 is:
img_v2_aaa30e32-08df-4c93-be21-542736fb32eg

Because the swagger template file's location is not discoverable at runtime, in a containerized environment. This PR fixes that problem.


hezyin (approved):
LGTM

PR #5131, opened 2023-05-08
refactor: using `crypto/rand` for password and secret key handling instead of `math/rand`

### LINKED ISSUE #5129: [Refactor][Plugin Utils] Consider using a secure package for password and secret key handling instead of `math/rand`
## What and why to refactor
Currently, the `math/rand` package is used for generating random numbers in our application. However, this package is not intended for handling secure information such as passwords and secret keys.

To ensure the security of our application, we should consider using a package specifically designed for handling such sensitive information. There are several well-known packages available in Go, such as `crypto/rand` and `golang.org/x/crypto/rand`, which are designed to generate cryptographically secure random numbers suitable for use in password and secret key generation. Therefore, I suggest we consider using a secure package for password and secret key handling instead of relying on `math/rand`.

## Related issues
No

## Additional context
No

mindlesscloud:
### Summary
fix #5129 ([Refactor][Plugin Utils] Consider using a secure package for password and secret key handling instead of math/rand)

Closes #5129

### DIFF
--- a/backend/core/plugin/plugin_utils.go
+++ b/backend/core/plugin/plugin_utils.go
@@ -140,6 +140,6 @@ func AesDecrypt(crypted, key []byte) ([]byte, errors.Error) {
 }
 
 // RandomEncKey will return a random string of length 128
-func RandomEncKey() string {
+func RandomEncKey() (string, errors.Error) {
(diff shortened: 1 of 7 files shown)


abeizn (approved):
LGTM

PR #5135, opened 2023-05-09
fix: no pagination is used in the collectionAccounts of gitlab

### LINKED ISSUE #5126: [Bug][Gitlab] No pagination is used in the 'collectAccounts' plugin
### Search before asking

### What happened

As a user I want to see the collected information about the PRs of Gitlab in the 'Engineering Throughput and Cycle Time - Team View' dashboard. To show this data DevLake needs information about all authors of the PRs. But in some special cases this user data is not available. In Gitlab this could happen when a project or group has more than 20 members. In this case only the first 20 members are fetched by DevLake because the `page_size` of pagination is set to 20 by default and the other pages are not fetched. So it could happen that the author information of a PR is missing  because the user data has not been fetched by the `collectAccounts` plugin. Due to the missing user data, the query of the dashboard would then not show the corresponding PRs.

For Gitlab the `members/all` REST API call is used: https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-and-invited-members 

The problematic part of the code for Gitlab can be found here: https://github.com/apache/incubator-devlake/blob/dfeec2ea732b90fa90f4cf09e612c756cf8aa679/backend/plugins/gitlab/tasks/account_collector.go#L42

If a Gitlab project or group has more than 20 members, the `collectAccounts` plugin should retrieve all members using the pagination API.

### How to reproduce

Add more than 20 members to a Gitlab project or group and let DevLake collect them all using the `collectAccounts` plugin.

### Anything else

_No response_

### Version

v0.16.0

### Code of Conduct

klesh:
### Summary
Cherrypick #5134 to v0.17

Closes #5126

### DIFF
--- a/backend/plugins/gitlab/tasks/account_collector.go
+++ b/backend/plugins/gitlab/tasks/account_collector.go
@@ -19,6 +19,7 @@ package tasks
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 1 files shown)


abeizn (approved):
lgtm

PR #5138, opened 2023-05-09
5081 fix missing records azuredevops

### LINKED ISSUE #5081: [Bug][Azure DevOps] Missing records in cicd_tasks
### Search before asking

### What happened

**Version: v0.17.0-beta3**

## Azure DevOps Setup

I have two pipeline runs in Azure DevOps, each with two jobs, "job_1" and "job_2".
Screen Shot 2023-05-03 at 6 50 34 PM
Screen Shot 2023-05-03 at 6 50 21 PM

## DevLake Collection Result

I got 2 cicd_pipelines, which was as expected. But I only got cicd_tasks for the second cicd_pipeline, see the pipeline_id column of the cicd_tasks I got. Also, all of their environment was set to "TESTING", which was unexpected.

I also didn't expect to get the "Finalize Build" job, but since it's also displayed as a job in the Azure DevOps pipeline page, so I think it's ok.

Screen Shot 2023-05-03 at 6 49 17 PM
Screen Shot 2023-05-03 at 6 50 03 PM

See above

### How to reproduce

See above

### Anything else

_No response_

### Version

main

### Code of Conduct

CamilleTeruel:
### Summary
Implements migration of tool tables.
Fix a non-unique primary key in azuredevops table

Closes #5081

### DIFF
--- a/backend/python/README.md
+++ b/backend/python/README.md
@@ -232,6 +232,25 @@ Do not forget `table=True`, otherwise no table will be created in the database. 
 
 To facilitate or even eliminate extraction, your tool models should be close to the raw data you collect. Note that if you collect data from a JSON REST API that uses camelCased properties, you can still define snake_cased attributes in your model. The camelCased attributes aliases will be generated, so no special care is needed during extraction.
 
+#### Migration of tool models
+
+Tool models, connection, scope and transformation rule types are stored in the DevLake database.
+When you change the definition of one of those types, you need to migrate the database.
+You should implement the migration logic in the model class by defining a `migrate` class method. This method takes a sqlalchemy session as argument that you can use to
+execute SQL `ALTER TABLE` statements.
+
+```python
+class User(ToolModel, table=True):
+    id: str = Field(primary_key=True)
+    name: str
+    email: str
+    age: int
+
+    @classmethod
+    def migrate(cls, session):
+        session.execute(f"ALTER TABLE {cls.__tablename__} ADD COLUMN age INT")
+```
+
 
 ### Create the stream class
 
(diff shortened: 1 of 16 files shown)


backend/python/README.md
```diff
@@ -232,6 +232,25 @@ Do not forget `table=True`, otherwise no table will be created in the database.
...
+```python
+class User(ToolModel, table=True):
+    id: str = Field(primary_key=True)
+    name: str
+    email: str
+    age: int
+
+    @classmethod
+    def migrate(cls, session):
+        session.execute(f"ALTER TABLE {cls.__tablename__} ADD COLUMN age INT")
```
hezyin, line 251: Hi @CamilleTeruel, some clarifying questions on how tool model migrations work:

1. When are these migrations run? Are they triggered from the Golang side?
2. Our Golang side keeps track of migrations by date so each time only new migrations are run. How do we avoid re-running migrations that have already been run? Would the `migrate` function throw an error if it's run on a later version of DevLake?

CamilleTeruel: > Hi @CamilleTeruel, some clarifying questions on how tool model migrations work:
> 
>     1. When are these migrations run? Are they triggered from the Golang side?

Yes, after loading a remote plugin, go-side calls run-migration.
 
>     2. Our Golang side keeps track of migrations by date so each time only new migrations are run. How do we avoid re-running migrations that have already been run? 

Right now we don't. This is just minimal support for migration to support this change. We must add versioning to migration functions.

> Would the `migrate` function throw an error if it's run on a later version of DevLake?

A migrate function can indeed throw an error if it has already run once, so we should be careful with migration until we have versionning support. But the `Job.migrate` function in particular is fine as it executes idempotent SQL statements.

hezyin: Hi @CamilleTeruel, thanks for the clarification. I'm merging this PR now. Could you please create a new GitHub issue for adding versioning to migration functions? Please also cherry-pick this fix back to release-v0.17 branch.

hezyin (approved):
LGTM

CamilleTeruel:
@keon94  This PR is missing from v0.17-beta6

PR #4906, opened 2023-04-12
[feat-4762]: Support deleting data by scope

keon94:
### Summary
1. New Plugin endpoint added: DeleteScope. It deletes all the data associated with a scope and optionally the scope itself too (by default it does). Returns all the impacted blueprints.
2. Python framework adapted to return table names for plugins back to Go, similar to how the GetTableInfo() API works on Go side.
3. GetScope and ListScopes API modified to support returning the blueprints using the scopes in addition to the rest of the payload.
4. Completely reworked the ScopeHelper to be generic enough to serve both Go plugins and Python (a.k.a remote) plugins (we had two originally roughly identical implementations which had diverged over time). Code maintenance should be much better now. (TxRules API refactor will be considered in another PR). This was done largely out of necessity to get the components of Delete to work.
5. Data deletion of scopes relies on the _raw_data_params field on the models - some plugin subtasks were not writing to that field. Reworked the framework code to automatically write to that field and remove the burden from the subtasks. This ended up touching all the plugins' code.
6. The original partially-implemented-now-obsolete Go Azure devops plugin completely removed as part of fixing broken tests. 

### Does this close any open issues?
Closes #4762 

### Screenshots
Sample new endpoints' swagger docs:
image
image

### Additional Tests
Github Dashboard after blueprint run across 3 repos
image

Github Dashboard after removing one repo
image

### Other Information
Any other information that is important to this PR.

Issue #4762, opened 2023-03-24
[Feature][BE] After users remove a board from a blueprint or connection, the data relates to the board will be deleted


Startrekzky:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

As a scrum manager, I re-organize Jira boards at times. When I first started using DevLake, I collected data from Jira board-1, then I created a new board in Jira called board-2, added board-2, and removed board one from the corresponding DevLake blueprint. 

I was expecting that DevLake would delete the board-1 data, but it did not happen.

### Description

I want DevLake to support the **data deletion of a removed data scope** after editing a blueprint, a connection, or any operation that might delete the data scope.

For example, if I collected b1, and then edited the blueprint to ONLY collect b2. 
- b1: i1, i2, i3, i4
- b2: i2, i3, i5, i6

b1 in table.boards, i1,i4 in table.issues, and all issue_changelogs, issue_labels, issue_worklogs related to i1 and i4 will be deleted.

i2 and i3 in b1 won't be affected as they also belong to other boards. See the pic below, the shadow part should be kept.

![image](https://user-images.githubusercontent.com/14050754/228779609-43dec35c-4d04-4cfd-8967-f83c9c92c542.png)

### To Do

- [ ] Detect if there're boards removed after users save a 'data scope edit'
- [ ] If so, prompt the users if they want to delete the data. The default operation is to delete or not (needs discussion)?
   - [ ] If users choose to delete the data of the removed data scope, only the data of the existing data scope will be kept.
   - [ ] If not, delete the data (belonging to existing boards) automatically when the bp runs the next time

### Related issues

Obsolete PR:
#4906 

Child PRs:
#5153 
#5155 

Child Issue:
#5277 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

keon94:
This sounds conceptually similar to deleting a blueprint. We essentially need to know what data belongs to a given scope which is just like the problem with blueprints: what data belongs to a given blueprint? I think both of these problems have to be solved in a similar way

keon94:
@Startrekzky @yumengwang03 Will it make more sense to implement a feature to delete data based on scopes instead? This will prevent tying the data to the blueprint that created them

[...]

PR #5153, opened 2023-05-10
[feat-4762] part1: Minimal support for deleting scopes

keon94:
### Summary
Adds minimal code to support deleting scopes. Only PagerDuty plugin will be supported for this PR.

Part of #4762 
Reduced from #4906 

### Screenshots
Screenshots found under #4906 

### Other Information


backend/helpers/pluginhelper/api/api_rawdata.go
```diff
@@ -37,16 +38,23 @@ type RawData struct {
...
 type RawDataSubTaskArgs struct {
 	Ctx plugin.SubTaskContext
 
 	//	Table store raw data
 	Table string `comment:"Raw data table name"`
 
-	//	This struct will be JSONEncoded and stored into database along with raw data itself, to identity minimal
-	//	set of data to be process, for example, we process JiraIssues by Board
-	Params interface{} `comment:"To identify a set of records with same UrlTemplate, i.e. {ConnectionId, BoardId} for jira entities"`
+	// Deprecated: Use Options instead
```
klesh, line 52: Please explain the reason behind it, I don't understand.

backend/helpers/pluginhelper/api/scope_helper.go
```diff
@@ -207,24 +243,53 @@ func (c *ScopeApiHelper[Conn, Scope, Tr]) GetScopeList(input *plugin.ApiResource
...
+	if params.loadBlueprints {
+		if len(scopeIdFieldName) == 0 {
+			return nil, errors.Default.New("scope Id field name is not known") //temporary, limited solution until I properly refactor all of this in another PR
+		}
+		scopesById := c.mapByScopeId(apiScopes, scopeIdFieldName[0])
+		var scopeIds []string
+		for id := range scopesById {
+			scopeIds = append(scopeIds, id)
+		}
+		blueprintMap, err := c.bpManager.GetBlueprintsByScopes(scopeIds...)
```
klesh, line 255: Is this correct? I believe the `connectionId` is missing here.

backend/helpers/pluginhelper/api/scope_helper.go
```diff
@@ -245,9 +310,95 @@ func (c *ScopeApiHelper[Conn, Scope, Tr]) GetScope(input *plugin.ApiResourceInpu
...
+			if err != nil {
+				return nil, errors.Default.Wrap(err, fmt.Sprintf("error extracting scope parameter name for scope %s", params.scopeId))
+			}
+		}
+		for _, table := range tables {
+			err = db.Exec(createDeleteQuery(table, rawScopeParamName, scopeParamValue))
+			if err != nil {
+				return nil, errors.Default.Wrap(err, fmt.Sprintf("error deleting data bound to scope %s for plugin %s", params.scopeId, params.plugin))
+			}
+		}
```
klesh, line 357: I think we have to delete related records from the `_devlake_collector_latest_state` table as well

backend/helpers/pluginhelper/api/scope_helper.go
```diff
@@ -245,9 +310,95 @@ func (c *ScopeApiHelper[Conn, Scope, Tr]) GetScope(input *plugin.ApiResourceInpu
...
+		return nil, errors.Default.Wrap(err, fmt.Sprintf("error verifying connection for connection ID %d", params.connectionId))
+	}
+	db := c.db
+	blueprintsMap, err := c.bpManager.GetBlueprintsByScopes(params.scopeId)
+	if err != nil {
+		return nil, errors.Default.Wrap(err, fmt.Sprintf("error retrieving scope with scope ID %s", params.scopeId))
+	}
+	blueprints := blueprintsMap[params.scopeId]
+	// find all tables for this plugin
+	tables, err := getPluginTables(params.plugin)
```
klesh, line 339: the function name is confusing, I suggest to rename it to `getAffectedTables`

klesh (commented):
Nice work, However, my pagerduty account has expired, unable to run it locally.
Please attache a set of screenshots to demonstrate the records in related tables are deleted correctly, including `raw layer` `tool layer` , `domain layer` and the `_devlake_collector_latest_state` tables.

backend/helpers/pluginhelper/api/api_rawdata.go
keon94: The idea here is to remove this boilerplate of setting the params in every single subtask of a plugin. This value is a constant for a given plugin (as far as I'm aware - correct me if I'm wrong), so it makes a lot more sense to have it be set in a single place and be enforced.  Doing this using an interface made the most sense to me.

backend/helpers/pluginhelper/api/scope_helper.go
keon94: Should it filter by bother connectionId and scopeId? Right now it only does based on scopeId

backend/helpers/pluginhelper/api/scope_helper.go
```diff
@@ -245,9 +310,95 @@ func (c *ScopeApiHelper[Conn, Scope, Tr]) GetScope(input *plugin.ApiResourceInpu
...
+	if err != nil {
+		return nil, errors.Default.Wrap(err, fmt.Sprintf("error getting database tables managed by plugin %s", params.plugin))
+	}
+	// delete all the plugin records referencing this scope
+	if rawScopeParamName != "" {
```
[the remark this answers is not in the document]
keon94: I don't think webhook even uses this API. It probably needs a separate solution

backend/helpers/pluginhelper/api/api_rawdata.go
klesh: I see, can we reuse the variable instead of creating a new one?

backend/helpers/pluginhelper/api/scope_helper.go
klesh: Pardon?  Did you mean both?
`ScopeId` itself is not sufficient to identify a scope, it should be used with `connectionId`, shouldn't it?
image

backend/helpers/pluginhelper/api/scope_helper.go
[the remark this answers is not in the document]
klesh: You are right, remember to deal with it.

backend/helpers/pluginhelper/api/api_rawdata.go
keon94: That will require widespread change across all the plugins which you wanted to avoid here. That can be deferred to another refactor PR.

klesh: I see, you are right. the `Params` is moved to the `TaskOptions`, It is the right way to go.

keon94:
> Nice work, However, my pagerduty account has expired, unable to run it locally. Please attache a set of screenshots to demonstrate the records in related tables are deleted correctly, including `raw layer` `tool layer` , `domain layer` and the `_devlake_collector_latest_state` tables.

@klesh 
Before Delete:
image
image
image
image


After Delete:
image
image
image
image

klesh (approved):
LGTM

PR #5160, opened 2023-05-11
fix: missing board for tapd

### LINKED ISSUE #5156: [Bug][TAPD] Board data not extracted to domain layer boards entity when collecting TAPD data in Apache DevLake
### Search before asking

### What happened

When using Apache DevLake to collect development data from third-party platforms, specifically TAPD, we encountered a problem where board data is not being extracted to the domain layer boards entity as expected. Even though the data certainly exists at the tool layer, it's not being properly transformed and transferred to the domain layer schema.

We expect that board data from TAPD should be successfully extracted and populated into the domain layer boards entity when collecting TAPD data using Apache DevLake.

### How to reproduce

Set up Apache DevLake to collect data from TAPD
Verify that there are boards with data at the tool layer in TAPD
Check if the board data is properly transformed and transferred to the domain layer schema in Apache DevLake.

### Anything else

d21283dc-ae41-4dfa-b6e6-484bc5d363f1
a8d89331-1732-4797-86f0-226c650391fe

### Version

main

### Code of Conduct

mindlesscloud:
### Summary
Fix #5156
The routine of converting the TAPD workspace to the board was moved from the subtask list to MakeDataSourcePipelinePlanV200, which would not be executed in the `ADVANCE` mode. In this PR, we fix it by adding a new subtask `convertWorkspace` to the subtask list, it will be called in any case, whether it is `NORMAL` or `ADVANCE`.

Closes #5156 

### Screenshots
image

### DIFF
--- a/backend/plugins/tapd/impl/impl.go
+++ b/backend/plugins/tapd/impl/impl.go
@@ -95,6 +95,7 @@ func (p Tapd) Description() string {
 
 func (p Tapd) SubTaskMetas() []plugin.SubTaskMeta {
 	return []plugin.SubTaskMeta{
+		tasks.ConvertWorkspaceMeta,
 		tasks.CollectWorkitemTypesMeta,
 		tasks.ExtractWorkitemTypesMeta,
 		tasks.CollectStoryCustomFieldsMeta,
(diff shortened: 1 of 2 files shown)


abeizn (approved):
LGTM

PR #5174, opened 2023-05-12
Refresh token

klesh:
### Summary
Refresh token

### Screenshots
refresh-token

### DIFF
--- a/config-ui/src/utils/request.ts
+++ b/config-ui/src/utils/request.ts
@@ -49,18 +49,19 @@ instance.interceptors.response.use(
             data: {
               refreshToken: refreshToken,
             },
-          })
-            .then((resp) => {
+          }).then(
+            (resp) => {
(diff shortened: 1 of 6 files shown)


config-ui/src/utils/request.ts
```diff
@@ -37,6 +39,38 @@ instance.interceptors.response.use(
...
+      if (refreshToken) {
+        refreshingToken =
+          refreshingToken ||
+          request('/login/refreshtoken', {
+            method: 'POST',
+            data: {
+              refreshToken: refreshToken,
+            },
+          })
+            .then((resp) => {
```
likyh, line 53: I think err handle should move to `.then` from `.catch(……`.  `.catch` will catch both request error and the error trigger by `.then`.

likyh: .then((resp) => {……}, (err) => {})

likyh (approved):

PR #5258, opened 2023-05-23
feat: github multiple labels match the same regex

### LINKED ISSUE #5079: [Question][GitHub] What's the right issue type when multiple labels match the same regex?
The following screenshot was taken on stable v0.16.0, but the issue also exists on main and v0.17.

Take issue https://github.com/apache/incubator-devlake/issues/4861 as an example, it has two `type/xxx` labels: `type/docs` and `type/bug`.

However, based on this [code](https://github.com/apache/incubator-devlake/blob/main/backend/plugins/github_graphql/tasks/issue_collector.go#L273), this issue's `original_type` column can only have one value (the exact value would depend on the execution order of the labels in the for loop).

This is what the issue looks like in our database:

Screen Shot 2023-05-03 at 9 47 28 AM

Questions to discuss:

1. Does an issue's `original_type` column support multiple original types?
2. Does an issue's `type` column support multiple types?

abeizn:
### Summary
github multiple labels match the same regex

Closes #5079 

### Screenshots
image

### Other Information


mindlesscloud (approved):
LGTM

PR #5286, opened 2023-05-25
5250 define migration script on python side

### LINKED ISSUE #5250: [Feature][pydevlake] Better migrations
### Search before asking

### Use case

Right now python plugins handle DB migration of tool models in a low level-manner, executing raw SQL statements.
This makes compatibility with both mysql and postgre difficult.
Moreover this is not integrated with the migration flow of go plugins.

### Description

It would be better for python plugin to return migration scripts that would be executed alongside migrations of go plugins.

### Related issues

_No response_

### Code of Conduct

CamilleTeruel:
### Summary
Let python plugin define migration scripts that are transmitted to go side like regular `MigrationScript`s.
That is to say, remote plugins now implement `PluginMigration`

Closes #5250


backend/python/pydevlake/pydevlake/migration.py
```diff
@@ -0,0 +1,113 @@
...
+        self.operations.append(DropColumn(table=table, column=column))
+
+    def drop_table(self, table: str):
+        """
+        Drops a table.
+        """
+        self.operations.append(DropTable(table=table))
+
+
+def migration(version: int):
```
keon94, line 84: could we add a second optional param for the script name? We can default it to fn.__name__ like in L99 if not provided.

backend/server/services/remote/models/conversion_test.go
```diff
@@ -0,0 +1,153 @@
...
+		"primaryKey": true,
+	}
+	tag := getGormTag(schema, stringType)
+	assert.Equal(t, "gorm:\"primaryKey;type:varchar(255)\"", tag)
+}
+
+func TestGetGormTagEncDec(t *testing.T) {
+	schema := map[string]interface{}{
+		"type":   "string",
+		"format": "password",
```
keon94, line 149: It's only semantics, but would calling the format "encrypted" or "secret" be more general than "password"?

keon94: oh, nevermind. this is a pydantic thing. I reckon it can't be changed

backend/core/utils/json.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func GetItem[T any](array JsonArray, index int) (T, errors.Error) {
+	if index < 0 || index >= len(array) {
+		return *new(T), errors.Default.New(fmt.Sprintf("Index %d out of range", index))
+	}
+	return convert[T](array[index])
+}
+
+func convert[T any](value any) (T, errors.Error) {
```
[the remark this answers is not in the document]
CamilleTeruel: Done.

backend/python/pydevlake/pydevlake/migration.py
CamilleTeruel: Done.

backend/server/services/remote/models/conversion_test.go
CamilleTeruel: What we can do is to provide a new type `EncStr` that has json schema prop `"encrypted": true` instead of `"format": "password"`

hezyin (dismissed):
LGTM

hezyin (approved):
LGTM

hezyin (approved):
LGTM

Issue #4866, opened 2023-04-06
[Feature][Framework/Config-UI] Allow enrichment to be run independently without running the entire blueprint


yumengwang03:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

If a user edits a Transformation, they will most likely want to see the updated dashboards for the new enrichments. But right now the collection and enrichment are both in the Blueprint and cannot be run separately, and therefore DevLake should allow enrichment to be run without running the entire Blueprint.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

yumengwang03:
<img width="1470" alt="Screen Shot 2023-05-11 at 5 04 18 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/e1cf43e3-f61d-4ba3-9b0f-897ce6fce9cf">
<img width="1378" alt="Screen Shot 2023-05-11 at 5 04 31 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/29adc789-c512-4301-b946-a3981bfe61fd">
Contact me if you need figma links or need a walkthrough.

yumengwang03:
<img width="1180" alt="Screen Shot 2023-05-22 at 2 42 56 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/9a87fc47-2745-4db5-8a40-2ccadbeece60">
<img width="1424" alt="Screen Shot 2023-05-22 at 2 43 22 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/794bdcfe-695f-4d0e-b693-e17912b3249a">
<img width="1318" alt="Screen Shot 2023-05-22 at 2 43 34 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/a3e35bd6-50f1-4a65-be8a-75513937ad42">

PR #5294, opened 2023-05-26
feat: allow enrichment to be run independently without running the en…

abeizn:
### Summary
allow enrichment to be run independently without running the entire blueprint.
It only takes effect once and will not be executed with crontab

part of #4866  

### Screenshots
image

### Other Information


mindlesscloud (approved):
LGTM

PR #5308, opened 2023-05-29
Skip empty blueprint

### LINKED ISSUE #5307: [Refactor][Blueprint] skip empty blueprints
## What and why to refactor
All blueprints will be scheduled even if there are no plans to be executed. it's a reasonable choice to skip and not run this kind of  blueprints

- Not run the blueprints which do not contain any plugin other than `org`, `dora` and `refdiff`
- Not run the blueprints if all of the plans are empty
- Allow creating empty blueprints in advanced mode

mindlesscloud:
### Summary
- Not run the blueprints which do not contain any plugin other than `org`, `dora` and `refdiff`
- Not run the blueprints if all of the plans are empty
- Allow creating empty blueprints in advanced mode

Closes #5307

### DIFF
--- a/backend/core/plugin/plugin_blueprint.go
+++ b/backend/core/plugin/plugin_blueprint.go
@@ -38,6 +38,19 @@ type PipelineStage []*PipelineTask
 // PipelinePlan consist of multiple PipelineStages, they will be executed in sequential order
 type PipelinePlan []PipelineStage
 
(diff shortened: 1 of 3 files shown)


abeizn (commented):
What about the advanced mode？

mindlesscloud:
@abeizn As long as the blueprint meets the first two conditions, it will be skipped regardless the mode

PR #5325, opened 2023-05-31
fix: github create connection failed with InstallationId required error when auth method is AccessToken

klesh:
### Summary
multi-auth

Installation ID should not be required when AuthMetho is AccessToken


mintsweet (approved):
LGTM.

PR #5335, opened 2023-05-31
cherrypick(v0.17.0-beta9): #5311

### LINKED ISSUE #5311: [Bug][PagerDuty] Plugin isn't pulling all incidents from PagerDuty to _raw_pagerduty_incidents table
### Search before asking

### What happened

I created a Blueprint with 1 Data Scope from GitHub and 2 from PagerDuty:
image

The Data Entity Issue Tracking is collected only by the PagerDuty Data Scopes.

Both PagerDuty Data Scopes (Services) have, in total, 217 incidents in the last 6 months.
image
image

But, when I Sync the Project and open the DORA Dashboard Validation, I see that it pulled only 9 incidents instead of 217:
image

Probably, this is because the incidents were just collected from the Service with 9 incidents. So I checked the table `board_issues` and can confirm that it has only incidents related to one of the two Services:
image

The PagerDuty plugin should pull all incidents related to the Data Scopes (Services) configured, not just one of them.

### How to reproduce

Create a blueprint with two or more Services and Sync.

The total number of incidents is going to be the number of incidents of one Service instead of all of them added together.

### Anything else

_No response_

### Version

v0.17.0-beta8

### Code of Conduct

keon94:
### Summary
PagerDuty API params was missing the service ID reference. This is already fixed in main/v0.18, but wasn't back-ported to v0.17.

Closes #5311

### DIFF
--- a/backend/plugins/pagerduty/tasks/incidents_collector.go
+++ b/backend/plugins/pagerduty/tasks/incidents_collector.go
@@ -64,6 +64,7 @@ func CollectIncidents(taskCtx plugin.SubTaskContext) errors.Error {
 		Ctx: taskCtx,
 		Params: PagerDutyParams{
 			ConnectionId: data.Options.ConnectionId,
+			ServiceId:    data.Options.ServiceId,
 		},
 		Table: RAW_INCIDENTS_TABLE,
 	}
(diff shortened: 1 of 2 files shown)


hezyin (approved):
LGTM

PR #5356, opened 2023-06-02
ListScopes returns "null" instead of "[]"

CamilleTeruel:
### Summary

The ListScopes endpoint was responsing with a "null" body instead of "[]" when there was no scope. It was not causing issues before, but it does with the new config UI.


hezyin (approved):
LGTM

PR #5376, opened 2023-06-06
Kw 5371 scope with config

### LINKED ISSUE #5371: [Refactor][pluginhelper] Load the whole scopeConfig instead of scopeConfigName for scopeHelper
## What and why to refactor
Previously, we would load the `scopeConfigName` for `list scopes` and `get scope` APIs for all our plugins, so that config-ui could be able to display the related `scopeConfig` without calling another API. However, there are a couple of complication:

1. The new prototype requires `DataEntities` to be displayed when listing scopes or viewing a specific scope.
2. with #5159, the DataEntities option was moved into ScopeConfig, which lead to an extra API request.

Most of us if not all had came to an agreement that those APIs should provide `ScopeConfig` entity for simplicity's sake.

1. Remove `scopeConfigName` from the API response
2. Add `scopeConfig` which essentially is the related scope config entity if `scope_config_id` is available

## Related issues
#5159 

@keon94 @CamilleTeruel python plugins need to adopt as well.

klesh:
### Summary
1. upgrade `swaggo` to the latest version for better Generic support
2. fixed the #5371 
3. rename all `scope_config` to `scope-config` for all swagger doc

Closes #5371 

### Screenshots
image
image


mindlesscloud (approved):
LGTM

PR #5380, opened 2023-06-06
fix(Jira): Fix Jira scope reflection field name: BoardID -> BoardId

keon94:
### Summary
Fixes the call to `/plugins/jira/connections/{connectionId}/scopes/?blueprints=true` erroring out with:
image

It errors out because the reflection param value name for Boards doesn't exactly match the name on the struct:
https://github.com/apache/incubator-devlake/blob/f3e23cebde39cea18b618c7c72546eeff938099f/backend/plugins/jira/models/board.go#L29

Closes n/a

### Screenshots

### Other Information


mindlesscloud (approved):
LGTM

PR #5415, opened 2023-06-09
fix(config-ui): the offline page error.

mintsweet:
### Summary
1. refactor(config-ui): move global store to base layout.
2. fix(config-ui): the style for buttons component.
3. fix(config-ui): the offline page error.


abeizn (approved):
lgtm

PR #5468, opened 2023-06-13
fix: fix jira remote api

### LINKED ISSUE #5467: [Bug][Jira] remote api get only 50 results boards but not all
### Search before asking

### What happened

image

can use remote API to get all results.

### How to reproduce

1. set a connection with boards of more than 50.
2. try to scroll the scope list to visit all boards of it.
3. check board count and real board count.

### Anything else

_No response_

### Version

2b190844de47d343d84dcf351c6ac57aed826298

### Code of Conduct

mappjzc:
### Summary
Add analyzingQuery to get real PerPage and Page.
Because some Jira servers limit the maxResults.

Closes #5467


backend/plugins/jira/api/remote.go
```diff
@@ -76,3 +83,8 @@ func initialQuery(queryData *api.RemoteQueryData) url.Values {
 	query.Set("startAt", fmt.Sprintf("%v", (queryData.Page-1)*queryData.PerPage))
 	return query
 }
+
+func analyzingQuery(maxResults int, startAt int, queryData *api.RemoteQueryData) {
+	queryData.PerPage = maxResults
+	queryData.Page = startAt/maxResults + 1
```
mindlesscloud, line 89: could the variable `maxResults` be zero?

backend/plugins/jira/api/remote.go
```diff
@@ -58,14 +58,21 @@ func RemoteScopes(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, er
...
+				StartAt    int                 `json:"startAt"`
+				Values     []apiv2models.Board `json:"values"`
 			}{}
 
 			err = api.UnmarshalResponse(res, &resBody)
 			if err != nil {
 				return nil, err
 			}
 
+			if (queryData.PerPage != resBody.MaxResults) ||
```
mintsweet, line 71: why not use `isLast` as the condition?

image

backend/plugins/jira/api/remote.go
mappjzc: I am not sure. I think it can not happen.

backend/plugins/jira/api/remote.go
mappjzc: IsLast can not be used for this checking, because it only checks if maxResult is reset by remote.

PR #5470, opened 2023-06-13
[fix-5453](pydevlake): Make ScopeConfig optional

### LINKED ISSUE #5453: [Bug][Pydevlake] Python should allow no ScopeConfig configuration
### Search before asking

### What happened

Current PyDevlake enforces that a ScopeConfig needs to be set per Scope. This should not be mandatory per @Startrekzky.
I have fixed the Go side of the code in #5449.

Python will throw an error like this if ScopeConfig is not set:

image

This code should be made flexible to allow this possiblity.

### How to reproduce

See above

### Anything else

_No response_

### Version

main

### Code of Conduct

CamilleTeruel:
### Summary
Make ScopeConfig optional in pydevlake

Closes #5453

### DIFF
--- a/backend/server/services/remote/plugin/plugin_impl.go
+++ b/backend/server/services/remote/plugin/plugin_impl.go
@@ -19,6 +19,7 @@ package plugin
 
 import (
 	"fmt"
+
 	"github.com/apache/incubator-devlake/core/dal"
 	"github.com/apache/incubator-devlake/core/errors"
 	coreModels "github.com/apache/incubator-devlake/core/models"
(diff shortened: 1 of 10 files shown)


backend/server/services/remote/plugin/plugin_impl.go
```diff
@@ -176,8 +177,10 @@ func (p *remotePluginImpl) getScopeAndConfig(db dal.Dal, connectionId uint64, sc
 		if err != nil {
 			return nil, nil, err
 		}
+		return wrappedScope.Unwrap(), wrappedScopeConfig.Unwrap(), nil
+	} else {
+		return wrappedScope.Unwrap(), nil, nil
```
keon94, line 182: Was ```wrappedScopeConfig.Unwrap()``` not just resolving to nil here?

keon94: If you want to break the code up like this, I suggest initializing `wrappedScopeConfig` inside the if statement to keep its scope limited to there.

CamilleTeruel: > Was `wrappedScopeConfig.Unwrap()` not just resolving to nil here?

No this resolves to a zeroed scope config.

CamilleTeruel:
One thing I considered was to force scope configs to only have fields with defaults. When a config is not set we set a scope config with only defaults. Like that plugin devs don't have to check whether its nil everytime.

klesh:
> One thing I considered was to force scope configs to only have fields with defaults. When a config is not set we set a scope config with only defaults. Like that plugin devs don't have to check whether its nil everytime.

That is how Golang plugin handles it, it is much easy, especially when the scope config is missing and we have to check the DataEntities. Note that we assuming users wants all DataEntities for missing scope config or `entities` is empty or null

CamilleTeruel:
> That is how Golang plugin handles it, it is much easy, especially when the scope config is missing and we have to check the DataEntities. Note that we assuming users wants all DataEntities for missing scope config or `entities` is empty or null

Alright, I changed the PR to do the same thing.

keon94 (approved):

PR #5497, opened 2023-06-15
fix: Assign proper value to the column `type`  of table `boards`

### LINKED ISSUE #5157: [Bug][Tapd] Tapd board type not set to scrum in domain layer's boards table
### Search before asking

### What happened

In the boards table, which belongs to the domain layer, there is a missing designation of the Type column for the Tapd board. As a result, the value for this column remains null.

The Type column for the Tapd board should be properly set and assigned the value of "scrum", since tapd is only support scrum board.

### How to reproduce

Setup devlake to collect tapd data.  
Check the type value of boards table, it kept null.

### Anything else

image

### Version

v0.16.0

### Code of Conduct

mindlesscloud:
### Summary
Fix #5157 ([Bug][Tapd] Tapd board type not set to scrum in domain layer's boards table)

Closes #5157

### DIFF
--- a/backend/plugins/customize/service/service.go
+++ b/backend/plugins/customize/service/service.go
@@ -167,7 +167,7 @@ func (s *Service) SaveBoard(boardId, boardName string) errors.Error {
 			Id: boardId,
 		},
 		Name: boardName,
-		Type: "kanban",
+		Type: "csv",
 	})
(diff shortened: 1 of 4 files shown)


abeizn (approved):
lgtm

PR #5506, opened 2023-06-16
[feat-5503]: Deletable scopes and connections

### LINKED ISSUE #5503: [Feature][framework] Check deletable before deleting connection/scope
### Definition:
> A scope is considered "deletable" if it is not referenced by any blueprint. Similarly, a connection is "deletable" if all scopes associated with it are deletable.

To ensure data integrity, we need to perform a deletable check before deleting scopes or connections. However, if the user simply wants to clear the data, we won't perform this check.

### Delete Scope
When attempting to delete a scope, the following steps should be followed:

If the scope is not deletable, we should not delete it. Instead, return a status code 409 (conflict) along with the following JSON body:
```json
{
  "projects": [
    "projectName1",
    "projectName2"
  ],
  "blueprints": [
    "blueprintName1",
    "blueprintName2"
  ]
}
```
### Delete Connection
When attempting to delete a connection, the process should be as follows:

If the connection is deletable, remove all scopes associated with it and then delete the connection itself.
If the connection is not deletable, we should not delete it. Instead, return a status code 409 (conflict) along with the following JSON body:
```json
{
  "projects": [
    "projectName1",
    "projectName2"
  ],
  "blueprints": [
    "blueprintName1",
    "blueprintName2"
  ]
}
```
By following these guidelines, we can ensure that only deletable scopes and connections are removed, and conflicts are properly handled by returning the appropriate status codes and JSON bodies.

### Related issues

_No response_

### Code of Conduct

keon94:
### Summary
What does this PR do?
- Deleting connections and scopes now requires that they are not being referenced by any blueprints, otherwise the endpoints will return error 409 with the list of those blueprints and their projects in the payload.
- Deleting a project now also deletes its associated blueprint (to prevent having orphaned blueprints)
- Test cases added for these scenarios

Closes #5503 

### Screenshots
Swagger examples:
image
image

Deleting a connection used by blueprints (status 409):
image
Deleting a scope used by blueprints (status 409):
image

### Other Information


keon94:
See this comment: https://github.com/apache/incubator-devlake/issues/5503#issuecomment-1594156478
I have not implemented "auto delete scopes upon deleting a deletable connection" for the reason stated there.

backend/helpers/pluginhelper/api/connection_helper.go
```diff
@@ -101,8 +104,16 @@ func (c *ConnectionApiHelper) List(connections interface{}) errors.Error {
...
+func (c *ConnectionApiHelper) Delete(plugin string, connection interface{}) (*services.BlueprintProjectPairs, errors.Error) {
+	connectionId := reflectField(connection, "ID").Uint()
+	referencingBps, err := c.bpManager.GetBlueprintsByConnection(plugin, connectionId)
+	if err != nil {
+		return nil, err
+	}
+	if len(referencingBps) > 0 {
+		return services.NewBlueprintProjectPairs(referencingBps), errors.Conflict.New("Found one or more references to this connection")
+	}
+	return nil, CallDB(c.db.Delete, connection)
```
klesh, line 116: also need to check if there are scopes referencing the connection

backend/helpers/pluginhelper/api/scope_generic_helper.go
```diff
@@ -288,43 +290,49 @@ func (gs *GenericScopeApiHelper[Conn, Scope, Tr]) GetScope(input *plugin.ApiReso
...
 	if err != nil {
-		return err
+		return nil, err
 	}
 	err = gs.dbHelper.VerifyConnection(params.connectionId)
 	if err != nil {
-		return errors.Default.Wrap(err, fmt.Sprintf("error verifying connection for connection ID %d", params.connectionId))
+		return nil, errors.Default.Wrap(err, fmt.Sprintf("error verifying connection for connection ID %d", params.connectionId))
+	}
+	if refs, err := gs.getScopeReferences(input.GetPlugin(), params.connectionId, params.scopeId); err != nil || refs != nil {
```
klesh, line 302: We should allow users to clean the scope even if it is referenced

backend/helpers/pluginhelper/api/connection_helper.go
```diff
@@ -101,8 +104,16 @@ func (c *ConnectionApiHelper) List(connections interface{}) errors.Error {
 }
 
 // Delete connection
-func (c *ConnectionApiHelper) Delete(connection interface{}) errors.Error {
-	return CallDB(c.db.Delete, connection)
+func (c *ConnectionApiHelper) Delete(plugin string, connection interface{}) (*services.BlueprintProjectPairs, errors.Error) {
```
klesh, line 107: Let us move the `plugin` to the `ConnectionHelper` as a field. We should refactor it as a Generic type in the future.

keon94: We can't do that. The name of the plugin is only available at runtime, on the ApiResourceInput. See https://github.com/apache/incubator-devlake/blob/354652597537b4dec82049bff784e0e682e4967c/backend/core/plugin/plugin_api.go#LL35C2-L35C2.
The Scope Helpers also handle the plugin name in the same way.

backend/helpers/pluginhelper/api/scope_generic_helper.go
keon94: But disallowing that is part of the ticket's requirements:
https://github.com/apache/incubator-devlake/issues/5503#issuecomment-1593356778
image

Do you mean force the option via a flag?

backend/helpers/pluginhelper/api/connection_helper.go
keon94: See question 5: https://github.com/apache/incubator-devlake/issues/5503#issuecomment-1593356778
The front-end has a references to the scopes and will delete this after they've been removed.
@Startrekzky

Startrekzky: > also need to check if there are scopes referencing the connection

@klesh I think we should allow users to delete a connection that has data scope(s), as long as the connection and its data scopes are not referenced by any project or bp.

In this case, when deleting a connection, the frontend can loop the data scope(s) and delete both the data scopes under this connection and the connection itself.

backend/helpers/pluginhelper/api/scope_generic_helper.go
Startrekzky: @keon94  I agree with @klesh , the screenshot shows the criteria of 'deletable or not', but cleaning/reseting a data scope under a connection is not a 'delete' operation. Hence, cleaning/reseting a data scope is allowed when a data scope is referenced.

keon94: @Startrekzky, So we should only clean the data (as if delete_data_only=true), but still return status 409?

keon94: Done

backend/helpers/pluginhelper/api/connection_helper.go
klesh: https://github.com/apache/incubator-devlake/blob/main/backend/plugins/github/api/init.go#L42
Can we pass the `plugin` name as a Constant here? We are in the github plugin, right?

backend/helpers/pluginhelper/api/connection_helper.go
keon94: I agree the BE should behave in a self-contained way. But technically speaking, the whole lack of support for cascade-deletion that we initially decided on falls into this category, as well as the need for users to rerun their blueprints and manually fix DORA integrity problems. I think these have to be addressed with proper design and time in mind, and right now we're too tight on time to include in this release, so let's please settle for the simpler solution and have an issue/epic to capture the long term solution.

klesh: @keon94 I'm not asking for the cascade deletion, I'm asking you to check if there are any scopes left before deleting the connection, that is all.

backend/helpers/pluginhelper/api/connection_helper.go
keon94: It can be done, we just have to go and hardcode constants (the name) for each plugin. Right now, the plugins are registered by the names derived from their compiled *.so files. So as long as there is never a discrepancy between those names and these constants, it'll be ok.

backend/helpers/pluginhelper/api/connection_helper.go
keon94: What I can try doing is to do a check and return a 409 if there are scopes tied to this connection, with a list of scopes in the payload. The user will then have to manually go and delete them before being able to delete the connection. I'll make a separate PR for that to keep this one from blowing up any further.

keon94: https://github.com/apache/incubator-devlake/issues/5513

hezyin (approved):
LGTM

PR #5524, opened 2023-06-19
feat: add deep search for project

### LINKED ISSUE #5523: [Feature][zentao] use project on zentao and deep search bug and story by project.
### Search before asking

### Use case

support to only use project scope to collect bug, story, and task data.

### Description

Because the config ui only supports one type of scope, we need to compose the project and product.
Now we only keep the project and remove the product.

project need deep search the data to collect bug, story, and task. 

### Related issues

_No response_

### Code of Conduct

mappjzc:
### Summary
For the plan v0.18
The user could only set the project scope but not the product scope.
And we need the user to collect the story and bug data simultaneously.

Add deep search on the bug for the project scope.
Add deep search on the story for the project scope.
Add GetScopes for GetProjectScopes.

Remove product scopes.

The product scope will be supported on v0.19.
Closes #5523 

### Screenshots
image
image
image
image
image
image
image


backend/plugins/zentao/e2e/bug_test.go
```diff
@@ -47,6 +46,14 @@ func TestZentaoBugDataFlow(t *testing.T) {
 				},
 			},
 		},
+		ProductList: map[int64]string{
+			3: "",
+		},
+		FromBugList: map[int]bool{
```
likyh, line 52: what's `FromBugList`?

likyh (approved):
Let's try. It's seems right but I think we should test more.

PR #5526, opened 2023-06-20
fix: generate TAPD task parent Issue identifier by Story model

### LINKED ISSUE #5525: [Bug][tapd] no record of TAPD task's parent issue
### Search before asking

### What happened

In TAPD's task convertor, when process the `parent_issue_id` which is considered to a `story` of the `task`, the function `taskIdGen` is called instead of `storyIdGen` ([ref](https://github.com/apache/incubator-devlake/blob/v0.16.0/backend/plugins/tapd/tasks/task_converter.go#L72)).

Then the identifier of a task's parent issue starts with `tapd:TapdTask` instead of `tapd:TapdStory` which is assumed the correct one, so that cannot get the parent issue in domain layer.

The `parent_issue_id` of a `task` should start with `tapd:TapdStory`.

### How to reproduce

Collect TAPD data into domain layer, and execute the following query in database. There would be always no records for parent issues.

```
SELECT
    tbl.issue_id AS issue_id,
    parent_issues.id AS parent_issue_id
FROM (
    SELECT
        issues.id AS issue_id,
        issues.parent_issue_id AS parent_issue_id
    FROM
        issues AS issues
    WHERE
        issues.id LIKE 'tapd:TapdTask%' AND
        issues.parent_issue_id IS NOT NULL
    LIMIT 10
) AS tbl
INNER JOIN
    issues AS parent_issues
ON
    tbl.parent_issue_id = parent_issues.id
```

### Anything else

_No response_

### Version

v0.17.0-beta2

### Code of Conduct

usharerose:
### Summary
use `TapdStory` for task's parent Issue identifier generator instead of `TapdTask`

Closes #5525 

### Screenshots
N/A

### Other Information
N/A


klesh (approved):
LGTM

PR #5541, opened 2023-06-20
[fix-5535]: raw_data_params structural and query change for safer deletion

### LINKED ISSUE #5520: [Feature][PyDevlake] Use CamelCase field name in raw_data_params
### Search before asking

### Use case

Python is using snake_case in raw_data_params. It's not breaking anything, but it goes against the convention of Go plugins. For the sake of consistency, use CamelCase there. And if we ever have to introduce a standardized field in there, we won't be able to write a simple query for it (Currently we have ConnectionId  (Go) vs connection_id (Python) as an example, but we're not querying for it). Edit: Actually in #5541 we do start querying it.

### Description

_No response_

### Related issues

Part of #5541 

### Code of Conduct

### LINKED ISSUE #5535: [Bug][pluginhelper] Data Deletion may cause collateral damage
### Search before asking

### What happened

During investigation of #5534 , I notice the following logic
https://github.com/apache/incubator-devlake/blob/8f8edfc2a02b114160470fe68fad6f1dee3490fb/backend/helpers/pluginhelper/api/scope_generic_helper.go#L627

Seems like we are deleting records from all tables by examining the `raw_data_params` only with the `LIKE` clause.

1. `raw_data_params` itself doesn't contain the **plugin name** and **connection id**, which means it may delete records that belongs to other plugins or even connection id
2. `LIKE` clause is fuzzy, it is likely to match records belongs to another scope

connectionId and pluginName should be included in the SQL, and we should use the exact match `=` instead of `LIKE`

### How to reproduce

...

### Anything else

_No response_

### Version

b40efa574abbe0530d3cf623cbddb5a3e34bcc7f

### Code of Conduct

keon94:
### Summary
What does this PR do?
- Address the requirments of #5535
- Updates the e2e test CSV files to include the new "plugin" entry in _raw_data_params; hence the large number of file changes.

Closes #5535 
Closes #5520

### Screenshots
image

### Other Information


klesh (commented):
Hold it for further discussion.
1. We have to migrate all `_raw_data_params` columns for all tables or it will not be backward compatible.
2. We have to migrate the `_devlake_collector_latest_state` or it will break
3. Some of the members tend to use a new `column` instead of adding the `field` to existing column

I'm not saying it is off the table, but it is a massive work and we will have to re-test everything.
Let's go with a hacky solution for now and we may have time to discuss and think

keon94:
Closing in favor of #5551

PR #5550, opened 2023-06-21
Proxy fixes for github_graphql and gitextractor

### LINKED ISSUE #5502: [Bug][github] Graphql and git connections does not use proxy configuration
### Search before asking

### What happened

Devlake github connections for git and graphql data doesn't reach github when run behind a corporate http proxy. Instead the connection times out since direct connections are blocked by firewall rules:

```
time="2023-06-13 16:20:24" level=error msg=" [pipeline service] [pipeline #6] [task #48] run task failed with panic
	caused by: run task failed with panic (github.com/apache/incubator-devlake/helpers/pluginhelper/api.CreateAsyncGraphqlClient:71)
	Wraps: (2) Post "https://api.github.com/graphql": dial tcp 140.82.121.6:443: i/o timeout
	Wraps: (3) Post "https://api.github.com/graphql"
	Wraps: (4) dial tcp 140.82.121.6:443
	Wraps: (5) i/o timeout
	Error types: (1) *hintdetail.withDetail (2) *hintdetail.withDetail (3) *url.Error (4) *net.OpError (5) *net.timeoutError"
```
The plugin fails when it tries to get it's remaining graphql quota:
https://github.com/apache/incubator-devlake/blob/main/backend/plugins/github_graphql/impl/impl.go#L172
Which uses the oauth2 client created here:
https://github.com/apache/incubator-devlake/blob/main/backend/plugins/github_graphql/impl/impl.go#L163

I tried to fix this by changing the contexts `http.Client` as explained here:
https://go.googlesource.com/oauth2/+/refs/heads/master/oauth2.go#332
However my experience with go is limited so I could never get it to work.

As a workaround we set `HTTP_PROXY` environment variables to configure both git and graphql for now.

Both github graphql and git should be able to reach github throught the proxy.

### How to reproduce

Block direct traffic to github and force the application to go through an http_proxy and the graphql and git collector fail.

### Anything else

_No response_

### Version

v0.17.0-beta5

### Code of Conduct

albinvass:
### ⚠️ Pre Checklist

### Summary
Fix for github_graphql and gitextractor when running devlake behind a corporate proxy. Currently both connections ignores the proxy configuration option and times out when trying to connect to github.

#### gitextractor
The proxy option `git.ProxyTypeAuto` in libgit2 seems to only use proxy when http proxy environment variables are set, instead `git.ProxyTypeSpecified`should be used to correctly specify proxy url.

#### github_graphql
To set proxy config on oauth2 one has to pass a custom `http.Client` object to the library through the context when creating the oauth2 client:
https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.9.0:oauth2.go;l=334

Closes #5502

### DIFF
--- a/backend/plugins/github_graphql/impl/impl.go
+++ b/backend/plugins/github_graphql/impl/impl.go
@@ -189,7 +189,6 @@ func (p GithubGraphql) PrepareTaskData(taskCtx plugin.TaskContext, options map[s
 		}
 	}
 
-
 	httpClient := oauth2.NewClient(oauthContext, src)
 	endpoint, err := errors.Convert01(url.JoinPath(connection.Endpoint, `graphql`))
 	if err != nil {
--- a/backend/plugins/github_graphql/impl/impl.go
+++ b/backend/plugins/github_graphql/impl/impl.go
@@ -20,6 +20,7 @@ package impl
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/url"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


backend/plugins/github_graphql/impl/impl.go
```diff
@@ -166,7 +167,30 @@ func (p GithubGraphql) PrepareTaskData(taskCtx plugin.TaskContext, options map[s
 	src := oauth2.StaticTokenSource(
 		&oauth2.Token{AccessToken: tokens[0]},
 	)
-	httpClient := oauth2.NewClient(taskCtx.GetContext(), src)
+	oauthContext := taskCtx.GetContext()
```
keon94, line 170: I suggest refactoring all of this logic into a NewOauth2Client method inside `api_client.go`, and reusing it across these plugins.

likyh:
I think it's OK. Have you test them?

backend/plugins/github_graphql/impl/impl.go
klesh: @albinvass Nice work, LGTM
@likyh , please take a look if the fix is legit
@keon94 Agreed, but it is out of the scope of this PR. Would you like to create a issue and we can do it later?

albinvass:
I've tested it locally with some iptables configuration, but I'll deploy it in our test environment as a sanity check.

albinvass:
Works in our environment :)

klesh:
hi, @albinvass , The PR looks good, would you mind fixing the linting?

klesh (approved):
LGTM

PR #5552, opened 2023-06-21
fix: Jira _raw_data_params order

### LINKED ISSUE #5534: [Bug][Data Deletion] Not all associated entities are deleted after deleting a data scope
### Search before asking

### What happened

When deleting a Jira board, the data in the following entities are not deleted:
- [x] `boards` @keon94 
- [x] `issue_assignees`. Fixed in PR #5537 
- [x] `_tool_jira_boards` This is caused by the _raw_data_params value in this entity
image
- [x] `_tool_jira_issue_comments`
- [x] ...

When deleting a TAPD board, the data in the following entities are not deleted:
- [x] `boards` @keon94 
- [x] `_tool_tapd_story_statuses`
- [x] `_raw_tapd_api_bug_statuses`
- [x] and more 

When deleting a Jenkins board, the data in the following entities are not deleted:
- [x] `cicd_scopes` @keon94 

Delete all associated entities after the data deletion

### How to reproduce

N/A

### Anything else

_No response_

### Version

v0.18.0-alpha7

### Code of Conduct

mindlesscloud:
### Summary
Fix the `_raw_data_params` is not consistent for Jira. We introduced a  hook function `RawParamEncoder`, so users could customize the serialization of raw params by setting this function

Closes #5534 

### Screenshots
_tool_jira_boards
img_v2_fe381245-8905-4f7e-8d9d-69b3e8ea312g
boards
img_v2_91bb26b5-0689-49b2-8a92-5c0593a154dg

### DIFF
--- a/backend/helpers/pluginhelper/api/scope_generic_helper.go
+++ b/backend/helpers/pluginhelper/api/scope_generic_helper.go
@@ -70,6 +70,8 @@ type (
 		ScopeIdColumnName string `validate:"required"`
 		// This corresponds to the scope field on the ApiParams struct of a plugin.
 		RawScopeParamName string `validate:"required"`
+		// This function customizes how the RawDataParams are serialized
+		RawParamEncoder func(connectionId uint64, scopeId any) (string, errors.Error)
 	}
 	ScopeHelperOptions struct {
 		// Define this if the raw params doesn't store the ScopeId but a different attribute of the Scope (e.g. Name)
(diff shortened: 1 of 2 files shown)


keon94:
I don't think we need this, and it adds complexity. The sorting of the JSON keys is alphabetical, which is why for Jira BoardId comes before ConnectionId. So there is consistency; and once we take care of #5535 we shouldn't have problems.

PR #5558, opened 2023-06-23
[feat-5517]: Add Delete API for Scope configs

### LINKED ISSUE #5517: [Feature][BE+UI] Delete ScopeConfig support
### Search before asking

### Use case

We don't have a Delete ScopeConfig API. This ticket adds it to backend and front-end. Additionally:
- Deleting connections must check for any child scope-configs (backend's responsiblity), and cascade-delete them.
- Deleting a scope config must set to NULL all scope_config_id fields on all scopes that reference it.

### Description

_No response_

### Related issues

#5513 

### Code of Conduct

keon94:
### Summary
What does this PR do?

Closes #5517 

### Screenshots

### Other Information
New API added for all plugins supporting scope-configs:
`DELETE /plugins/{plugin-name}/connections/{connection-id}/scope-configs/{config-scope-id}` cc FYI @mintsweet (For UI)


backend/helpers/pluginhelper/api/connection_helper.go
```diff
@@ -117,13 +117,23 @@ func (c *ConnectionApiHelper) Delete(connection interface{}) (*services.Blueprin
...
 		return services.NewBlueprintProjectPairs(referencingBps), errors.Conflict.New("Found one or more blueprint/project references to this connection")
 	}
-	scopeModel := c.getScopeModel()
-	count, err := c.db.Count(dal.From(scopeModel.TableName()), dal.Where("connection_id = ?", connectionId))
+	src, err := c.getPluginSource()
 	if err != nil {
-		return nil, errors.Default.Wrap(err, fmt.Sprintf("error deleting scopes for plugin %s using connection %d", c.pluginName, connectionId))
+		return nil, err
+	}
+	if scopeModel := src.Scope(); scopeModel != nil {
```
klesh, line 124: No, we agreed that connection with any scopes shall NOT be deleted

keon94: @klesh Ok, I've made the change

backend/helpers/pluginhelper/api/scope_generic_helper.go
```diff
@@ -306,23 +306,18 @@ func (gs *GenericScopeApiHelper[Conn, Scope, ScopeConfig]) DeleteScope(input *pl
...
-	}
 	// now we can as scope to state its `Params` for data bloodline identification
 	if refs, err := gs.getScopeReferences(params.connectionId, params.scopeId); err != nil || refs != nil {
 		if err != nil {
 			return nil, err
 		}
 		return refs, errors.Conflict.New("Found one or more references to this scope")
 	}
-	if err = gs.deleteScopeData(toolScope); err != nil {
+	if err = gs.deleteScopeData(*scope); err != nil {
```
klesh, line 320: Why dereference?

keon94: Because Scope is of type *ToolLayerScope here, which is basically a *interface{}

backend/helpers/pluginhelper/api/scope_config_helper.go
```diff
@@ -104,29 +108,62 @@ func (t ScopeConfigHelper[Tr]) Update(input *plugin.ApiResourceInput) (*plugin.A
...
+	p, _ := plugin.GetPlugin(t.pluginName)
+	pluginSrc, ok := p.(plugin.PluginSource)
+	if !ok {
+		return errors.Default.New("plugin doesn't implement PluginSource")
+	}
+	scopeModel := pluginSrc.Scope()
+	if scopeModel == nil {
+		return nil
+	}
+	return t.db.Exec(fmt.Sprintf("UPDATE %s SET scope_config_id = NULL WHERE scope_config_id = %d", scopeModel.TableName(), scopeConfigId))
```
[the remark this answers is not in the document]
keon94: The problem is with Dynamic structs that we use for Python. Gorm doesn't play nicely with them, especially with update operations.

backend/helpers/pluginhelper/api/scope_generic_helper.go
klesh: Aah, I see.
I think need to update the `scope_db_helper` since `Scope` was a `struct` (so `pointer` made sense at that point) and has been changed to `ToolLayerScope` which should be a pointer already.

backend/helpers/pluginhelper/api/scope_config_helper.go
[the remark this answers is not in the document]
klesh: Sorry for the typo, it should be `db.Update`.
Alright then, let's add a comment to it to avoid mis-edit

backend/helpers/pluginhelper/api/scope_generic_helper.go
keon94: The problem is, at runtime the `Scope` generic is going to be a `Struct`, but at compile time it is the `ToolLayerScope` interface. So it makes sense to have `*Scope` because at runtime that resolves to, for example, `*GithubRepo`, which is what we want. But at compile them it gets treated as this awkward double-pointer. See for instance [this](https://github.com/apache/incubator-devlake/blob/9b9290e7a328d7c14d9e3ef06144d6e1f86c7d0e/backend/helpers/pluginhelper/api/scope_db_helper.go#L88). That's a case where we'd want the method to return a pointer.
Unfortunately, afaik, there's no syntax that would have the compiler treat the `Scope` generic as a struct so that `*Scope` would be treated like a normal pointer.

klesh: Ok

hezyin (approved):
LGTM

PR #5561, opened 2023-06-25
fix: index out of range error of BatchSave

### LINKED ISSUE #5560: [Bug][gitextractor] Panic error due to index out of range during commit collection
### Search before asking

### What happened

I encountered a panic error while attempting to collect commits in my project. The error message suggests that the index is out of range. 
image
image

There should be no error

### How to reproduce

A race condition causes the error, it only happens occasionally

### Anything else

_No response_

### Version

main

### Code of Conduct

mindlesscloud:
### Summary
A race condition causes the error.
A mutex is used to create a critical section to ensure only one goroutine at a time can execute that code section.

Closes #5560

### DIFF
--- a/backend/helpers/pluginhelper/api/batch_save.go
+++ b/backend/helpers/pluginhelper/api/batch_save.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


klesh (approved):
LGTM

PR #5576, opened 2023-06-26
fix: scope selection is not working

klesh:
### Summary
Changing the scopes of a connection would see the following error

image

**Cause**: the `useRowSelection` does not support changing the type of the `rowKey`
**Fix**: added an optional `getRowKey` for callers to be able to customize the behavior.


mindlesscloud (approved):
lgtm

PR #5577, opened 2023-06-26
fix: fix gitlab commit sha

mappjzc:
### Summary
Cherry pick of #5566 and #5501

PR #5591, opened 2023-06-28
Fix the migration for the introduction of the squash and diff columns

dantleech:
As noted [here](https://github.com/apache/incubator-devlake/pull/5577/files#r1244735425) the fix for `gitlab commit sha` introduced an erroneous migration which this PR should fix.

- Should this PR be on `release-17` or `master`?
- Should this be a new migration? (as i guess the previous migration hasn't been tagged yet)


klesh (approved):
LGTM

klesh:
@dantleech Thanks for the fix, nice work.

PR #5606, opened 2023-06-30
[feat-5598]: v0.18 Migrations for _raw_data_table and _raw_data_params

### LINKED ISSUE #5598: [Feature][Migrations] Write DB migrations for v0.17->v0.18 scopes to be backward-compatible
### Search before asking

### Use case

Any tool or domain scope saved to the DB prior to v0.18 will be missing the _raw_params_table column data which is critical for Deletion to work.

Write migration scripts to populate this field for these scopes.

### Description

_No response_

### Related issues

Encapsulates #5587 

### Code of Conduct

keon94:
### Summary
What does this PR do?
Migrations to populate _raw_data_table and _raw_data_params for all tool and domain scopes models.

Closes #5598 #5644 

### Screenshots
Before upgrade
mig01
mig02

After upgrade
mig03
mig04

### Other Information
Validated against a few plugins: azure, gitlab, pagerduty.


klesh (commented):
This approach would work temporarily, but it is not forward-compatible. 

Migration script should be self-contained and consistent, meaning it should not be using any models outside the module of itself because they could be changed over time, which changes the behavior of the script. That is why we copying the original `models` as `migrationscripts/archived`.

Another problem is that we should not do it in the Framework level: what if we compile only jira and boot up the system, now the migration script gets applied, and it won't get executed again if I do it for github next time.

I would suggest that we do it in one plugin one script manner without using the interface(it might change in the future as well). copy the struct if it is needed by the script.

keon94:
> This approach would work temporarily, but it is not forward-compatible.
> 
> Migration script should be self-contained and consistent, meaning it should not be using any models outside the module of itself because they could be changed over time, which changes the behavior of the script. That is why we copying the original `models` as `migrationscripts/archived`.
> 
> Another problem is that we should not do it in the Framework level: what if we compile only jira and boot up the system, now the migration script gets applied, and it won't get executed again if I do it for github next time.
> 
> I would suggest that we do it in one plugin one script manner without using the interface(it might change in the future as well). copy the struct if it is needed by the script.

Actually, as far as compilations, the Name() of the script contains the plugin name, meaning if we boot up the system with a new plugin it should detect the missing migration for that plugin based on the name. If you look at the migration history table, it is adding one record per plugin to it.
Perhaps this logic can be moved to a different package and not live in core (e.g. in helper/), but I highly prefer having something like this to reduce boilerplate and redundancies in our codebase.

keon94:
@klesh 
Actually, as far as compilations, the Name() of the script contains the plugin name, meaning if we boot up the system with a new plugin it should detect the missing migration for that plugin based on the name. If you look at the migration history table, it is adding one record per plugin to it. Also, perhaps this logic could be moved to a different package and not live in core (e.g. in helper/).
The only problem I am seeing here now is that for future plugins these sorts of migration end up getting executed as well, so yeah, sadly I think I might have to resort back to duplicated-code approach. ugh

backend/core/plugin/migrationscripts/register.go
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+package migrationscripts
+
+import "github.com/apache/incubator-devlake/core/plugin"
+
+// All this is intended for migrations that are common across plugins (generalized plugin migrations)
+func All(p plugin.PluginMeta) []plugin.MigrationScript {
+	return allScripts(
+		newAddRawParamTableForScopes(p),
```
keon94, line 25: I suppose we could technically add a list of explicit plugin names to this constructor to make this only applicable to those intended plugins... e.g. `newAddRawParamTableForScopes("azuredevops", "github", "github_graphql", "gitlab", "jenkins", "pagerduty", "tapd")`...

keon94: but this begins to kind of defeat the purpose "generality"

keon94:
Code's been reworked to just use regular migrations for plugin tool tables.

hezyin (dismissed):
LGTM. @klesh Could you please give this PR another pass?

backend/core/models/migrationscripts/20230630_add_raw_param_table_for_domain_scopes.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}()
+	for _, domainModel := range domaininfo.GetDomainTablesInfo() {
+		if scopeModel, ok := domainModel.(plugin.Scope); ok {
+			scopes := toSlice(scopeModel)
+			err := db.All(scopes)
+			if err != nil {
+				return err
+			}
+			sliceSize, err := forEach(scopes, func(scope plugin.Scope) errors.Error {
```
klesh, line 53: Should we divide this logic into plugins as well? The `plugin.Scope` might be changed in the future, or the returned value might vary from time to time.  The key is that migration scripts should not depend on things that would change over time.

keon94: No, I definitely wouldn't want to do that. But it may be possible to write this to use `map[string]any` instead of the Scope interface. Right now this logic is not even strongly relying on it.

keon94: Instead of checking if the domain model is of Scope type, we could check of it has an empty raw table field.

keon94: If you want not rely on domainInfo models either, we have to switch to use `map` and a hardcoded list of table names.

keon94: I was initially opposed to this but have rethought it. Since we have a _small_ number of Scope domain models we can just have a migration for them listed out explicitly. But in general doing this wouldn't be a smart solution.

backend/core/models/migrationscripts/20230630_add_raw_param_table_for_domain_scopes.go
```diff
@@ -0,0 +1,95 @@
...
+	db := basicRes.GetDal().Begin()
+	defer func() {
+		if r := recover(); r != nil {
+			err := db.Rollback()
+			if err != nil {
+				basicRes.GetLogger().Error(err, "error rolling back transaction")
+			}
+		}
+	}()
+	err := script.updateDomainScopeTables(db,
```
keon94, line 49: @klesh I checked and I believe these are the only domain-scopes. Can you verify?

klesh: LGTM

backend/plugins/bamboo/models/migrationscripts/20230630_add_raw_param_table_for_scopes.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/apache/incubator-devlake/core/plugin"
+	"github.com/apache/incubator-devlake/plugins/bamboo/models/migrationscripts/archived"
+)
+
+var _ plugin.MigrationScript = (*addRawParamTableForScope)(nil)
+
+type addRawParamTableForScope struct{}
+
+func (script *addRawParamTableForScope) Up(basicRes context.BasicRes) errors.Error {
+	return basicRes.GetDal().UpdateColumn(archived.BambooProject{}.TableName(), "_raw_data_table", "_raw_bamboo_scopes",
```
klesh, line 33: Shouldn't we update the `_raw_data_params` as well? I believe they are missing in previous versions.

klesh: As to other plugins

keon94: I don't think so. Plugins have been writing connection_id+scope_id to _raw_data_params since well before the Delete feature.

keon94: But I assume there was never a migration done for any of them when "scopes" were introduced...

keon94: Should we do that in another PR?

klesh: I don't think so, the `scopehelper` already introduced in v0.17 which didn't write the `_raw_data_params` at all, we have to fix the existing data scope for deleting to function properly

keon94: I've logged this: https://github.com/apache/incubator-devlake/issues/5644
That'll be a tough one to do, but we should be good to close this PR out

backend/plugins/jenkins/models/migrationscripts/20230630_add_raw_param_table_for_scopes.go
```diff
@@ -0,0 +1,81 @@
...
+			src.RawDataParams = string(errors.Must1(json.Marshal(&params20230630{
+				ConnectionId: src.ConnectionId,
+				FullName:     src.FullName,
+			})))
+			updateSet := []dal.DalSet{
+				{ColumnName: "_raw_data_table", Value: src.RawDataTable},
+				{ColumnName: "_raw_data_params", Value: src.RawDataParams},
+			}
+			where := dal.Where("id = ?", fmt.Sprintf("jenkins:JenkinsJob:%v:%v", src.ConnectionId, src.FullName))
+			db.UpdateColumns("repos", updateSet, where)
```
abeizn, line 67: Maybe need checking the returned error or choosing to ignore it.

abeizn (approved):
lgtm

Issue #4864, opened 2023-04-06
[Feature][Framework] Support for Project, Blueprint and Scope deletion


keon94:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

Deleting projects,  blueprints and scopes is currently unsupported. This ticket will capture the R&D necessary to support this feature.

### Description

There are technical challenges involved with this. The main question is, how can we purge all the domain-level data associated with a blueprint once we delete that blueprint? We need a way to perform that mapping so we know what to delete from the database.

### Child issues

Scope deletion (PagerDuty + AzureDevops): #4762 
UI work: #4388
Blueprint deletion PR: #5227 
Project deletion PR: #5289 
Scope deletion (Other plugins): #5277 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

keon94:
I'm assigning this to myself. cc @Startrekzky

yumengwang03:
Hi @keon94 here is some additional info: in v0.18, we will (most likely) move data scope and transformation to be managed at the data connection level, and Blueprints will only retain its association with the data scope, meaning that adding and deleting data scope need to be done in each data connection, not from the Blueprint anymore. Deleting data scope from a Blueprint doesn't delete the data scope, but delete its association with the Blueprint.

keon94:
@yumengwang03 @Startrekzky Will it make more sense to repurpose this ticket to delete data based on scopes instead? Doing so will prevent data from getting tied the blueprint that created them

yumengwang03:
Adding the prototypes here:
<img width="1412" alt="Screen Shot 2023-05-11 at 5 10 02 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/96615030-6af3-462a-86c8-fe36e7f1e20d">
<img width="1603" alt="Screen Shot 2023-05-11 at 5 10 34 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/b95fef22-2e95-4b42-b21f-11e76e65c84c">

PR #5609, opened 2023-06-30
feat: pipeline/scope-deletion mutual exclusive

klesh:
### Summary

Due to #4864 had been introduced, there are a couple of side effects that must be taken care of:
1. What if a scope gets deleted while related pipelines are still running(users may remove the scope from the project while there is a pipeline running, so it is possible for the user to delete it on the Connection page): some unwanted data would be produced and they are hard to be identified and removed
2. What if a new pipeline gets created while deletion is in progress: this is unlikely to happen when the deletion is quick, but it might when the number of records is huge, especially after we support  Cascade Connection Deletion.

This PR makes pipeline-execution and scope-deletion mutually exclusive to prevent the above things from happening.

**NOTE: We should reconsider most of our operation and consider wrap them inside Transaction due to Deletion Support**

### Screenshots

Deleting scope while any pipeline is running would return error:
image
The Config-UI need to adopt the new error response format to show message correctly: @mintsweet the `Data` is renamed to `data` 
image
Snipaste_2023-06-29_17-43-00

Triggering pipeline while any deletion is taking place would be blocked until deletion finished
image


backend/helpers/dbhelper/txhelper.go
```diff
@@ -0,0 +1,96 @@
...
+	return l.tx
+}
+
+// LockTablesTimeout locks tables with timeout
+func (l *TxHelper[E]) LockTablesTimeout(timeout time.Duration, tables map[string]bool) errors.Error {
+	println("timeout", timeout)
+	c := make(chan errors.Error, 1)
+	go func() {
+		c <- l.tx.LockTables(tables)
+	}()
```
mindlesscloud, line 52: It's better to close the channel here

backend/core/errors/util.go
```diff
@@ -28,3 +28,17 @@ func Is(err, target error) bool {
 func As(err error, target any) bool {
 	return errors.As(err, &target)
 }
+
+func Must(err error) {
+	if err != nil {
+		print("napiaskj")
```
keon94, line 34: remove the print

backend/impls/dalgorm/dalgorm_transaction.go
```diff
@@ -48,6 +50,56 @@ func (t *DalgormTransaction) Commit() errors.Error {
 	return nil
 }
 
+func (t *DalgormTransaction) LockTables(tables map[string]bool) errors.Error {
```
keon94, line 53: For better readbility and future-proofing, I suggest you change the input of this function from map to []struct where the struct has these fields 
```
TableName string
LockExclusivity bool
```
or better (to have more compile time safety)
```
Tabler dal.Tabler
LockExclusivity bool
```

backend/helpers/pluginhelper/api/scope_generic_helper.go
```diff
@@ -297,7 +300,24 @@ func (gs *GenericScopeApiHelper[Conn, Scope, ScopeConfig]) GetScope(input *plugi
 	return scopeRes, nil
 }
 
-func (gs *GenericScopeApiHelper[Conn, Scope, ScopeConfig]) DeleteScope(input *plugin.ApiResourceInput) (*serviceHelper.BlueprintProjectPairs, errors.Error) {
+func (gs *GenericScopeApiHelper[Conn, Scope, ScopeConfig]) DeleteScope(input *plugin.ApiResourceInput) (refs *serviceHelper.BlueprintProjectPairs, err errors.Error) {
+	txHelper := dbhelper.NewTxHelper(gs.basicRes, &err)
+	defer txHelper.End()
+	tx := txHelper.Begin()
+	err = txHelper.LockTablesTimeout(2*time.Second, map[string]bool{"_devlake_pipelines": false})
```
keon94, line 307: the pipeline locking and check should be its own function `LockPipelines(...)`. Perhaps it can move to the helper package next to BlueprintManager?

klesh: Not really, the tables to be locked depends on contexts.
Sometimes we just need the `_devlake_pipelines`, sometimes we need the `_devlake_pipeline_labels`, and others as well.

backend/impls/dalgorm/dalgorm_transaction.go
klesh: This is a good idea, can you log it as an issue? I will do it later in another PR

backend/core/errors/util.go
```diff
@@ -28,3 +28,17 @@ func Is(err, target error) bool {
 func As(err error, target any) bool {
 	return errors.As(err, &target)
 }
+
+func Must(err error) {
```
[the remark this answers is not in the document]
keon94: Ok. Although for multi-wrapping, there is no harm as that doesn't mess with the status_code of the error. If you wrap a 400-type error with a Default error, for example, the new error just inherits that 400 status. Just wanted to clarify.

backend/impls/dalgorm/dalgorm_transaction.go
keon94: https://github.com/apache/incubator-devlake/issues/5622

mindlesscloud (approved):
LGTM

PR #5611, opened 2023-06-30
Kw upgrade go1.20.4 for better generic support

klesh:
### Summary

Same source code would fail on golang v0.19, but not v1.20

https://github.com/apache/incubator-devlake/actions/runs/5421209399/jobs/9856360137

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -111,7 +111,7 @@ RUN cd /usr/local/deps/target/lib && \
     done
 
 
-FROM python:3.9.9-slim-bullseye as base
+FROM python:3.9-slim-bullseye as base
 
 ENV PYTHONUNBUFFERED=1
 
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -31,7 +31,7 @@ FROM --platform=linux/arm64 debian:bullseye as debian-arm64
 RUN apt-get update
 RUN apt-get install -y libssh2-1-dev libssl-dev zlib1g-dev
 
-FROM --platform=$BUILDPLATFORM golang:1.19-bullseye as builder
+FROM --platform=$BUILDPLATFORM golang:1.20-bullseye as builder
 
 # docker build --build-arg GOPROXY=https://goproxy.io,direct -t mericodev/lake .
 ARG GOPROXY=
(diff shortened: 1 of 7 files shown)


devops/docker/lake-builder/Dockerfile
```diff
@@ -55,7 +55,7 @@ COPY --from=builder /tmp/deps/include/ /usr/include/
 ENV PKG_CONFIG_PATH=/usr/lib/x86_64-linux-gnu/pkgconfig
 
 # Install Golang
-RUN curl -L https://git.io/vQhTU | bash -s -- --version 1.19.7
+RUN curl -L https://git.io/vQhTU | bash -s -- --version 1.20.4
```
keon94, line 58: lets might as well bump the python version from 3.9.9 to 3.9.16 on line 44

backend/Dockerfile
```diff
@@ -31,7 +31,7 @@ FROM --platform=linux/arm64 debian:bullseye as debian-arm64
 RUN apt-get update
 RUN apt-get install -y libssh2-1-dev libssl-dev zlib1g-dev
 
-FROM --platform=$BUILDPLATFORM golang:1.19-bullseye as builder
+FROM --platform=$BUILDPLATFORM golang:1.20-bullseye as builder
```
keon94, line 34: here too. L144, 3.9.9 -> 3.9.16

klesh: in this case, we should 3.9? I believe it refers to the latest minor version

devops/docker/lake-builder/Dockerfile
klesh: same as above

klesh:
@keon94 done, please take another look

backend/Dockerfile
keon94, line 34: Maybe for golang lets use an explicit Go version (e.g. 1.20.4), declared as a variable in the Dockerfile, so we can also use it on L58 to keep the compiled version and runtime versions consistent

keon94 (approved):

klesh:
DONE

keon94 (approved):
LGTM

mindlesscloud (approved):
lgtm

PR #5625, opened 2023-07-05
fix: connectionHelper.Delete returns proper error response

### LINKED ISSUE #5624: [Bug][helper] ConnectionHelper.Delete should be returning the same format as ScopeHelper.Delete
### Search before asking

### What happened

In #5609, we updated the error response format of the `ScopeHelper` to:
```
{
  "success": bool,
  "message": "errormessage",
  "data": {
    "blueprints": [ referencing blueprints ],
    "projects:" [ referencing projects ]
  }
}
```
So Config-UI can distinguish situations of **referencing**(need to show a list of resources that are referencing the scope) and **other errors**(showing the error message only) and act accordingly.

The `ConnectionHelper.Delete` has a similar situation and it is returning null for **other errors** situation.

Update `ConnectionHelper` to output the same format as `ScopeHelper`

### How to reproduce

no need

### Anything else

_No response_

### Version

d21bf45

### Code of Conduct

klesh:
### Summary
Returning error message alongside `refs`

Closes #5624

### Screenshots
Self Tested on Jenkins, the other plugins should work because they are all using the same underlying code.

### DIFF
--- a/backend/helpers/pluginhelper/api/connection_helper.go
+++ b/backend/helpers/pluginhelper/api/connection_helper.go
@@ -19,8 +19,10 @@ package api
 
 import (
 	"fmt"
-	"github.com/apache/incubator-devlake/helpers/pluginhelper/services"
 	"strconv"
+
+	"github.com/apache/incubator-devlake/helpers/pluginhelper/services"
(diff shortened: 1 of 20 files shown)


mindlesscloud (approved):
lgtm

PR #5648, opened 2023-07-07
feat: add  migration script linter

klesh:
### Summary
Writing database migration scripts is hard because it must be forward-compatible meaning no matter codebase changes, the migration script behavior must not be affected, importing structs or relying on interfaces to return some values to migrate is most likely broken. e.g. a field that existed before might be deleted in the future causing the migration script producing different database structure.

However, it is easier said than done. Human beings are NOT to be trusted with such a subtle rule. We need aid, so, here we are. A linter tries to help us mitigate the problem.

###  How it works
The `linter` would scan the given Golang source files and detect if the `import`ed packages are allowed:
1. Any 3rd packages / std libs are considered OK
2. Any sub packages of the currently package of the file are considered OK
3. Whitelisted packages by using the `-a` option

#### Github action

An github action has been added, errors/warnings should be able to show on the PR like the following
Snipaste_2023-07-10_14-09-53

#### Local mode

Run the following command in the `backend` folder:
```sh
[klesh@klesh-laptop] ~/P/m/l/backend (kw-migration-script-linter|*28+2)> make migration-script-lint                                                                          ✗ 0.266s
go run core/migration/linter/main.go -- $(find . -path '**/migrationscripts/**.go')                                                                                                  
warning: ./core/models/migrationscripts/20220927_add_snapshot.go imports forbidden package "github.com/apache/incubator-devlake/core/models/common"                                  
        ./core/models/migrationscripts/20220927_add_snapshot.go:23:2                                                                                                                 
warning: ./plugins/sonarqube/models/migrationscripts/20230321_expend_project_key.go imports forbidden package "github.com/apache/incubator-devlake/core/models/common"               
        ./plugins/sonarqube/models/migrationscripts/20230321_expend_project_key.go:23:2                                                                                              
exit status 1                                                                                                                                                                        
make: *** [Makefile:173: migration-script-lint] Error 1
```


mindlesscloud (approved):
lgtm

PR #5655, opened 2023-07-10
[feat-5360]: Normalize Blueprints across multiple tables

### LINKED ISSUE #5360: [Refactor][Framework] Break up Blueprint config in the database
## What and why to refactor
Blueprints are currently stored in a monolithic fashion in the database, with the plan and settings stored as encrypted json blobs within them. The blob structure makes searching for blueprints by some blob attribute impossible except by querying for all blueprints and brute-force searching through them to filter for the one(s) we're looking for. This is bad for performance obviously.

Break the plan and settings up into other tables and link them to the BP table via IDs. We'll have to decide which columns of these tables should be encrypted. At the moment, the main use case will be having at least a blueprint_scopes table, which will be referenced by blueprint_connections, referenced by blueprint_settings. Having these tables will be the minimum requirement of this ticket.  

## Related issues
#5153 - introduced brute-force searching for blueprints by scopes. This ticket should optimize that.

## Additional context
n/a

keon94:
### Summary
What does this PR do?

Closes #5360 

### Screenshots

### Other Information

### DIFF
--- a/backend/core/models/blueprint.go
+++ b/backend/core/models/blueprint.go
@@ -31,65 +31,46 @@ const (
 	BLUEPRINT_MODE_ADVANCED = "ADVANCED"
 )
 
-// @Description CronConfig
-type Blueprint struct {
-	Name        string          `json:"name" validate:"required"`
-	ProjectName string          `json:"projectName" gorm:"type:varchar(255)"`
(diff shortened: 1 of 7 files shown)


klesh:
Hi, @keon94 , thanks for the work, are you still working on it? I would like to take it over since it is part of my refactor plan, is it ok for you?

PR #5658, opened 2023-07-11
fix: can not add connection to bp

mindlesscloud:
### Summary
Fix the Zentao connections can not be added to the blueprint

related to #5653


abeizn (approved):
lgtm

Issue #5203, opened 2023-05-16
[Feature][Config UI] Allow Data Scope search when creating a Blueprint in the user flow


yumengwang03:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

In the user flow where the management of Data Scope has been moved to Data Connections, users need to be able to search Data Scope when they create Blueprints for Projects.

### Related issues

https://github.com/apache/incubator-devlake/issues/4871

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

yumengwang03:
<img width="911" alt="Screen Shot 2023-05-23 at 4 58 15 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/2943181c-fe68-4a66-b451-4e3a8ae6dd38">
<img width="804" alt="Screen Shot 2023-05-23 at 4 58 25 PM" src="https://github.com/apache/incubator-devlake/assets/9535430/09cbf89b-8bc1-4167-bdea-586c1d00d30b">

PR #5699, opened 2023-07-19
fix: support gitlab data scope search

abeizn:
### Summary
fix: support gitlab data scope search
update "ProjectId" to "Name"

img_v2_b207ed02-c9cb-4184-b582-8b3b885319fg

reletad to #5203 

### Screenshots

### Other Information

### DIFF
--- a/backend/helpers/pluginhelper/api/scope_generic_helper.go
+++ b/backend/helpers/pluginhelper/api/scope_generic_helper.go
@@ -74,6 +74,7 @@ type (
 		ScopeIdColumnName string `validate:"required"`
 		// This corresponds to the scope field on the ApiParams struct of a plugin.
 		RawScopeParamName    string `validate:"required"`
+		// This corresponds to the scope field for allowing data scope search.
 		SearchScopeParamName string
 	}
 	ScopeHelperOptions struct {
--- a/backend/helpers/pluginhelper/api/scope_generic_helper.go
+++ b/backend/helpers/pluginhelper/api/scope_generic_helper.go
@@ -73,7 +73,8 @@ type (
 		// This corresponds to the database column name of the scope struct's ID (typically primary key) field
 		ScopeIdColumnName string `validate:"required"`
 		// This corresponds to the scope field on the ApiParams struct of a plugin.
-		RawScopeParamName string `validate:"required"`
+		RawScopeParamName    string `validate:"required"`
+		SearchScopeParamName string
 	}
 	ScopeHelperOptions struct {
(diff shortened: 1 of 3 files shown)


backend/plugins/gitlab/api/init.go
```diff
@@ -44,7 +44,7 @@ func Init(br context.BasicRes, p plugin.PluginMeta) {
 	params := &api.ReflectionParameters{
 		ScopeIdFieldName:  "GitlabId",
 		ScopeIdColumnName: "gitlab_id",
-		RawScopeParamName: "ProjectId",
+		RawScopeParamName: "name",
```
keon94, line 47: RawScopeParamName corresponds to the scopeId field on the ApiParams struct that Gitlab uses. Is that called "name"? And it would have to be "Name" since the fields are capitalized

abeizn: @keon94 Yes, updated.

keon94: But we're using ProjectId: https://github.com/apache/incubator-devlake/blob/6ea437954f0e1064e2fab331aa096c002ebab70a/backend/plugins/gitlab/tasks/shared.go#L141

keon94: I don't think it's the same problem, but Github uses the "Name" field in its ApiParams (Gitlab doesn't seem to though), and the way to handle that is to "derive" the Name from the ScopeId, and for that you need to implement a special function. See how it's done here: https://github.com/apache/incubator-devlake/blob/166e0b7a36e1279cdc15df96b659518377c838b7/backend/plugins/github/api/init.go#L60

backend/helpers/pluginhelper/api/scope_generic_helper.go
```diff
@@ -73,7 +73,8 @@ type (
 		// This corresponds to the database column name of the scope struct's ID (typically primary key) field
 		ScopeIdColumnName string `validate:"required"`
 		// This corresponds to the scope field on the ApiParams struct of a plugin.
-		RawScopeParamName string `validate:"required"`
+		RawScopeParamName    string `validate:"required"`
+		SearchScopeParamName string
```
keon94, line 77: can you add a comment to this field to explain what it is?

keon94 (dismissed):
LGTM

backend/helpers/pluginhelper/api/scope_generic_helper.go
abeizn: added.

mindlesscloud (approved):

PR #5701, opened 2023-07-19
fix: azuredevops automigrate failed on newer gorm

klesh:
### Summary

dropping the "NOT NULL" constraint for Primary Key is not allowed in PG

### Does this close any open issues?
Closes #5687

### Screenshots
Snipaste_2023-07-19_17-05-26

PR #5705, opened 2023-07-20
fix: connection id should be uint64

klesh:
### Summary
fix a typo for #5701 

img_v2_cfdb68c9-7e8a-4332-b710-343e8f56f11g

### DIFF
--- a/backend/server/services/remote/models/models.go
+++ b/backend/server/services/remote/models/models.go
@@ -140,7 +140,7 @@ type PipelineData struct {
 
 type ToolModel struct {
 	common.NoPKModel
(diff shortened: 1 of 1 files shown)


backend/server/services/remote/models/models.go
```diff
@@ -140,7 +140,7 @@ type PipelineData struct {
 
 type ToolModel struct {
 	common.NoPKModel
-	ConnectionId string `json:"connectionId" gorm:"column:connection_id;not null"`
+	ConnectionId int `json:"connectionId" gorm:"column:connection_id;not null"`
```
keon94, line 143: I think it's better if we made it uint64 so it's like our other models

keon94 (approved):

PR #5702, opened 2023-07-19
fix: no pipelines data when running pipelines on postgres

abeizn:
### Summary
fix: no pipelines data when running pipelines
add
```
		if r != nil {
			msg = fmt.Sprintf("%v", r)
		}
```
to judge the normal situation

### Does this close any open issues?
Closes #5703 

### Screenshots
image

image

### Other Information
Any other information that is important to this PR.

PR #5711, opened 2023-07-20
fix: pipeline dequeue deadlock

klesh:
### Summary

Before:
image
After:
image

Bug introduced by #5702


abeizn (approved):
lgtm

PR #5726, opened 2023-07-24
fix: unify scopeConfig field for task_data

klesh:
### Summary
Most of the plugins use the singular form for `scopeConfig` in their `TaskOption`
image
However, there are some blacksheeps:
img_v2_63e1c175-615b-4c3c-8bb3-85e2eebefbcg


mindlesscloud (approved):
lgtm

PR #5736, opened 2023-07-25
fix: apicollector might report table exists when running in parallel

### LINKED ISSUE #5733: [Bug][helper] ApiCollector fails on concurrent tasks
### Search before asking

### What happened

When multiple subtasks belonging to the same plugin are running in parallel, they might try to create the same `raw table` at the same time, which of cause would cause problems:

Snipaste_2023-07-25_09-08-09

raw tables should be created successfully

### How to reproduce

1. create project
2. add multiple connections from the same plugin
3. add one scope for each connection
4. trigger project and see error

### Anything else

_No response_

### Version

v0.18.0-beta2

### Code of Conduct

klesh:
### Summary
Add a mutex lock to avoid table creation conflict

Closes #5733 

### Screenshots
now they can be executed in parallel without error
Snipaste_2023-07-25_14-45-38

### DIFF
--- a/backend/helpers/pluginhelper/api/api_collector.go
+++ b/backend/helpers/pluginhelper/api/api_collector.go
@@ -139,6 +139,13 @@ func NewApiCollector(args ApiCollectorArgs) (*ApiCollector, errors.Error) {
 
 var rawTableAutoMigrateLock sync.Mutex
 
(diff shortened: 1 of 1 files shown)


mindlesscloud (approved):
LGTM

PR #5745, opened 2023-07-25
feat(config-ui): add healthcheck endpoint to nginx conf

peakwinter:
### Summary
This PR adds a healthcheck endpoint to the nginx config for the config-ui. It is particularly helpful when using basic authentication to protect access to the config UI, but when deploying to services like Kubernetes on GKE when some healthchecks cannot be configured to use basic auth (or are difficult to configure that way).

N/A

### Screenshots
N/A

### Other Information
N/A

### DIFF
--- a/config-ui/nginx.conf
+++ b/config-ui/nginx.conf
@@ -2,6 +2,7 @@ server {
   listen 4000;
   server_name localhost;
   absolute_redirect off;
+
 ${SERVER_CONF}
 
   location / {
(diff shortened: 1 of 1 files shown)


klesh (commented):
LGTM

klesh (approved):

klesh:
@peakwinter Thanks for your contribution, it seems good to me, but notice it works only for the `config-ui` container which is the frontend, NOT the `devlake` backend.

PR #5751, opened 2023-07-26
refactor: simplify jira scope config CRUD

### LINKED ISSUE #5315: [Refactor][jira] Simplify scopeConfig(transformationRule) CRUD logic
## What and why to refactor
Currently, the jira scopeConfig(transformationRule) CRUD logic is quite complex because some of its fields are using complex-type in order to store structural data.

The code below stores those complex-typed fields in `json` by using a ProxyStruct which works but is cumbersome.

https://github.com/apache/incubator-devlake/blob/e68ba0cfc0b61eb44dbe8a01f8f701ee8f207e97/backend/plugins/jira/tasks/task_data.go#L43-L106

During the development of #5159, I found out we can simply use the `gorm json serializer` to make thing much easier:
https://github.com/apache/incubator-devlake/blob/289008f544dc22723772937ca5cd9c44a2c9fb6f/backend/core/models/common/base.go#L33

With this approach, we can use ComplexType for any field we need without manually handling the save/load detail.

We should refactor our code to adopt such a method to reduce the complexity.

klesh:
### Summary
Use json data type to simplify the jira scope config CRUD logic

Closes #5315

### Screenshots
image

### DIFF
--- a/backend/helpers/pluginhelper/api/apihelperabstract/validate.go
+++ b/backend/helpers/pluginhelper/api/apihelperabstract/validate.go
@@ -1,0 +1,24 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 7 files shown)


mindlesscloud (approved):
LGTM

Issue #3729, opened 2022-11-14
[Refactor] Field types for json are not unified 


warren830:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

In our project, some of fields' type in migrationscripts is datatypes.JSON, and related fields' type in models is json.RawMessage.

### What do you expect to happen

We need to keep them the same

### How to reproduce

none

### Version

main

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #5779, opened 2023-07-31
Refactor: Database uses json type but not Golang code

klesh:
### Summary
Currently, we tend to use `json` in the golang sourcecode while `longtext` or `longblob` in the database
image
image
which introduces many hardcoded logic for conversion
image
image

**IT IS UGLY**

By reversing the pattern, we can simplify our code base and make it easier to understand:

```golang
type Task struct {
	...
	Subtasks       []string               `json:"subtasks" gorm:"type:json;serializer:json"`  // use json in database
	Options        map[string]interface{} `json:"options" gorm:"serializer:encdec"`  // encrypted field stay as text type
	...
}
```
image
This brings the following **merits**:

1. Our Golang is more cleaner, easy to understand and maintain
2. We can use `json` query function provided by the dbs for debugging if needed to

This is the PoC and very beginning of the refactor, please feel free to comment and share your thoughts.

@keon94 @mindlesscloud @abeizn @d4x1 @likyh @chenggui53 

Thanks.

Part of #3729 

### Screenshots
mysql
Snipaste_2023-07-28_15-35-37
**postgres**
before migration
Snipaste_2023-08-02_11-11-43
after migration
Snipaste_2023-08-02_11-12-52
Snipaste_2023-08-02_11-13-13

### DIFF
--- a/backend/core/models/migrationscripts/20230728_tasks_use_json.go
+++ b/backend/core/models/migrationscripts/20230728_tasks_use_json.go
@@ -54,7 +54,6 @@ func (script *tasksUsesJSON) Up(basicRes context.BasicRes) errors.Error {
 			if len(src.Subtasks) == 0 {
 				return nil, nil
 			}
-			println("src.Subtask", string(src.Subtasks))
 			errors.Must(json.Unmarshal(src.Subtasks, &dst.Subtasks))
 			return dst, nil
 		},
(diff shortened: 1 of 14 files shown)


backend/core/models/migrationscripts/20230728_tasks_use_json.go
```diff
@@ -0,0 +1,70 @@
...
+		"_devlake_tasks",
+		[]string{"subtasks"},
+		func(src *srcTaskSubtaskJSON20230731) (*dstTaskSubtaskJSON20230731, errors.Error) {
+			dst := &dstTaskSubtaskJSON20230731{
+				Model: src.Model,
+			}
+			if len(src.Subtasks) == 0 {
+				return nil, nil
+			}
+			println("src.Subtask", string(src.Subtasks))
```
mindlesscloud, line 57: It looks like this is a line of debug code

mindlesscloud (approved):
LGTM

PR #5787, opened 2023-08-02
fix tooling versions for v0.18 (beta4)

klesh:
### Summary
1. upgraded tooling versions for `make go-dep`
2. remove `dlv` from the docker image


mindlesscloud (approved):
LGTM

PR #5803, opened 2023-08-03
refactor: move iso8601time and csttime to core/models/common

klesh:
### Summary
Currently, we have 2 `iso8601time` structs defined in `helpers/pluginhelper/api` and `helpers/utils` which is certainly a mistake. Another problem is both `iso8601time` and `csttime` were put into `helpers/api` which is not the right place for such commonly used structs. This PR fixes these problems.

1. delete `iso8601time` in the `helpers/utils`
2. move `iso8601time` and `csttime` from `helpers/pluginhelper/api` to `core/models/common` package
3. all **archived** models use plain `time.Time` instead of the `iso8601time` or `csttime` because they must be immutable and they don't need the features offered by these structs.


backend/core/models/common/iso8601time.go
```diff
@@ -77,7 +77,7 @@ func init() {
 
 // Iso8601Time is type time.Time
 type Iso8601Time struct {
-	time   time.Time
+	Time   time.Time
```
d4x1, line 80: If `Time` field is not private, then method `Iso8601Time.ToTime` will be duplicated. 
Can we delete it ?

backend/impls/dalgorm/db_mapper.go
```diff
@@ -33,7 +34,7 @@ func ToDatabaseMap(tableName string, m map[string]any) map[string]any {
 			continue
 		}
 		if str, ok := v.(string); ok {
-			t, err := api.ConvertStringToTime(str)
+			t, err := common.ConvertStringToTime(str)
```
d4x1, line 37: `common` is under `models` package, `ConvertStringToTime` is a function, can we put it to `pluginhelper`?

d4x1 (commented):
BTW, why don't we use time.RFC3339 directly ?

abeizn (approved):
lgtm

PR #5825, opened 2023-08-08
refactor: replace datatypes.JSON with json.RawMessage for migrationscripts

klesh:
### Summary

Purpose:
1. Isolates `gorm` from our system
2. Unifying model definition style across all migration scripts so contributors would get the right idea when referencing the existing scripts

Part of #3729


abeizn (approved):
lgtm

Issue #4852, opened 2023-04-04
[Bug][Swagger] The api parameters of refdiff are not displayed clearly


mintsweet:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

The api parameters of refdiff are not displayed clearly.

![screenshot-20230404-103154](https://user-images.githubusercontent.com/37237996/229671827-491f59bf-a497-4fff-8086-00039c44d1b0.png)

![screenshot-20230404-103210](https://user-images.githubusercontent.com/37237996/229671831-ada8ff14-0fe4-4426-94fd-d6a03fb441b1.png)

### What do you expect to happen

Detailed display of api parameters about refdiff.

### How to reproduce

1. make dev in root dir.
2. visit `http://localhost:8080/swagger/index.html`.

### Version

main

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #5830, opened 2023-08-08
refactor: move RefdiffOptions to refdiff/models package

klesh:
### Summary

In order to calculate `commits` between releases, github/gitlab would generate a `refdiff` task during `make plan` phase. In the current implementation, we rely on passing the `options map[string]interface{}` to control `refdiff` behavior. It poses a couple of drawbacks:

1. The swagger doc won't be able to display the `ScopeConfig.Refdiff` correctly, it would be displayed as `{}` which is useless
2. `map[string]interface{}` is **Untyped**, we wont be able to detect errors if `refdiff` were to change its field-naming

This PR moves RefdiffOptions to refdiff/models package

Step 1 of #4852


abeizn (approved):
lgtm

PR #5877, opened 2023-08-15
fix: scopes would not be listed over 50

### LINKED ISSUE #5853: [Bug][Config-ui] Adding data scope succeeds but the scope is never listed
### Search before asking

### What happened

Ia have a gitlab connection which works for many projects using it. I am creating a new devlake project but I can't manage to add a new data scope to it. Each time I go through the wizard to add the data scope, it terminates successfully (message under) but when I get back to the connection at `https://<devlake-ui-host>/connections/gitlab/2` it does not list the new data scope, which implicitly means that I can't configure this new project as I can't assign it a data scope, which is a blocker for me.

Screenshot 2023-08-10 at 16 11 45

I am expecting to find the new data scope in the list of scopes so I can select and use it.

### How to reproduce

1. Create a gitlab connection
3. Add datascope to it
4. Verify that you get the message confirming their creation
5. Verify that you find the created data scopes and are able to use/select them

Make sure to do 3-5 more than 50 times, I suspect there could be a limit somewhere, see next section..

I could reproduce this behavior with several repositories on my gitlab connection, so it does not seem to be related to the repositories themselves.

### Anything else

I notice that I have 50 datascopes listed in this connection. Could there be a limitation in your code which only returns 50 datascope objects? I find it peculiar that it stops at such a round number.

I find no suspicious message in config-ui's logs. 

### Version

0.18.0beta4

### Code of Conduct

klesh:
### Summary
Remove pagination for scopes list api

Closes #5853

### DIFF
--- a/backend/helpers/pluginhelper/api/scope_db_helper.go
+++ b/backend/helpers/pluginhelper/api/scope_db_helper.go
@@ -113,9 +113,8 @@ func (s *ScopeDatabaseHelperImpl[Conn, Scope, Tr]) GetScopeAndConfig(connectionI
 }
 
 func (s *ScopeDatabaseHelperImpl[Conn, Scope, Tr]) ListScopes(input *plugin.ApiResourceInput, connectionId uint64) ([]*Scope, errors.Error) {
-	limit, offset := GetLimitOffset(input.Query, "pageSize", "page")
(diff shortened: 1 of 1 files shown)


abeizn (approved):
lgtm

PR #5908, opened 2023-08-18
fix: converting pipelines to deployments should ignore skipped items

### LINKED ISSUE #5885: [Question][GitHub] How to avoid skipped jobs counting as successful production deployment?
## Question
Our CI/CD is based on `release-please` and for every PR merged we deploy to our staging environment but for production we only deploy when a release pr was merged, otherwise we skip this job, as screen shots bellow.

We have setup our Data Transformation to identify all deploys matching `deploy(.*)` and deploys matching `deploy-prod` as production deployments.

But when we made only staging deployment, devlake mark the workflow as a production deployment in our metrics. 

There is a way we can setup devlake to avoid this skipped jobs or the best approach is to divide our deployments in different workflows?

Thanks.
## Screenshots
Case when deploy-prod was skipped and shouldn't be counted as a production deployment
image

Case when deploy-prod runs and should be counted as a production deployment
image

## Additional context
Add any other context here.

klesh:
### Summary
fix: converting pipelines to deployments should ignore skipped items

Closes #5885

### Screenshots
e2e test case is updated and passed
image


abeizn (approved):
lgtm

PR #5912, opened 2023-08-18
fix: devlake timezone issue

### LINKED ISSUE #426: Data type for storing datetime values and timezone handling
## Description

Now we have two data-types for `DATETIME` storage:

1. `time.Time` / `DATETIME` like `jira plugin`
2. `string` / `longtext` like `gitlab plugin`

Both have their pros and cons.
For `DATETIME` type we have to deal with the timezone problem, because mysql doesn't store timezone information for `DATETIME` type.
For `longtext`, it's hard for user to query on Grafana Dashboard, they have to manually convert string to `DATETIME` type, do the filtering then convert it back, makes query tedious. And there is performance issue for indexing.

For the sake of user friendly, I suggest we opt for `DATETIME`, and address the timezone problem.
The solution is to store/load them with a consistence timezone, and this timezone must be what user prefers, and could be modified when needed.

### Some information I found:

> In MySQL 8.0.19 and later, you can specify a time zone offset when inserting a TIMESTAMP or DATETIME value into a table. See Section 9.1.3, “Date and Time Literals”, for more information and examples.
> https://dev.mysql.com/doc/refman/8.0/en/datetime.html

If I understand it correctly, mysql will convert date-string like `2038-01-19 08:00:00+08:00` to a specific timezone, for example, if  that specific timezone was UTC, then it became `2038-01-19 00:00:00` in database.  The specific timezone can be configured through connection string. 

> gorm配置的时区为使用本地时区，通过查询数据库表可以观察得到，生成的时间为本地当前时间。
> https://teroy.github.io/2018/12/01/gorm-loc/

According to above article,  by setting up `loc` we specify the timezone that mysql uses to store `DATETIME` values.

### How `time.Time` value changes during store/load process

I've created an Demo to show how `DATETIME` got handled by `golang`, `gorm` and `mysql`

[klesh/gorm-mysql-timezone-demo](https://github.com/klesh/gorm-mysql-timezone-demo)

### Solution

1. Add `loc` setting to `DB_URI` to specify what timezone should mysql use to store `DATETIME` values.
2. `loc` is read from `TZ` Environment Variable, because this variable sets the default timezone for current golang process( dictate the timezone of `time.Now()`).
3. If `TZ` was omitted, `loc` should fall back to `Local` or it would be set to `UTC` by `gorm`, which is not so friendly when users try to browse the database directly.
4. Change all data type of `DATETIME` fields from string to `DATETIME`

## Q&A

- What if the server had a default timezone which was not what I wanted?
  Set the `TZ` Environment Variable before launching devlake
- Do I need to set `TZ` on my local machine?
  Normally, No! The timezone on your local machine is what your preferred, NORMALLY.
- My database is having all my date-time values wrong, like couple hours off.
  Delete database, Set the `TZ` Environment Variable and launch devlake to re-pull all the data
- Can I move the whole deployment to another server?
  Yes, as long as your TZ settings are consistence.

abeizn:
### Summary
fix: devlake timezone issue

Closes #426 

### Screenshots

### Other Information

### DIFF
--- a/devops/releases/lake-v0.18.0/docker-compose.yml
+++ b/devops/releases/lake-v0.18.0/docker-compose.yml
@@ -44,6 +44,7 @@ services:
       MYSQL_DATABASE: lake
       MYSQL_USER: merico
       MYSQL_PASSWORD: merico
+      TZ: UTC
     restart: always
     depends_on:
       - mysql
@@ -59,6 +60,7 @@ services:
       - ./.env
     environment:
       LOGGING_DIR: /app/logs
+      TZ: UTC
     depends_on:
       - mysql
 
@@ -80,4 +82,3 @@ volumes:
   mysql-storage:
   grafana-storage:
   devlake-log:
-
--- a/devops/releases/lake-v0.18.0/docker-compose.yml
+++ b/devops/releases/lake-v0.18.0/docker-compose.yml
@@ -27,6 +27,7 @@ services:
       MYSQL_DATABASE: lake
       MYSQL_USER: merico
       MYSQL_PASSWORD: merico
+      TZ: UTC
     command: --character-set-server=utf8mb4
       --collation-server=utf8mb4_bin
       --skip-log-bin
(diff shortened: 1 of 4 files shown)


ZhangNing10 (commented):
mysql and config-ui also need to be unified with the same TZ

ZhangNing10 (approved):
LGTM

klesh (approved):

PR #5957, opened 2023-08-25
Fix azure devops migration from 0.17 to 0.18

### LINKED ISSUE #5802: [Bug][Azure devops] Can't associate scope config
### Search before asking

### What happened

Using v0.18.0-beta3 (upgraded from 0.17)

I have a connection to Azure devops, test say 'Connected'.

In this project I also have a 'Data scope' associated with this project.

But 'Scope config' for this 'Data scope' says 'N/A'.

When I want to associate a scope config (an existing one) I just get a red square on the top saying 'Operation failed.'.

When I want to edit that scope config (the pencil icon) I get 'TypeError: Cannot read properties of null (reading 'map')'

Any thoughts?

Scope config should be attached to data scope.

### How to reproduce

Install 0.17, create connection, data scope, scope config, upgrade to 0.18

### Anything else

Sadly nothing in development console or in pod logs

### Version

v0.18.0-beta3

### Code of Conduct

CamilleTeruel:
### Summary
Fix various migration issues from 0.17 to 0.18 been introduced in recent 0.18 versions.
I still have a problem with git extractor but it might be unrelated.

Closes  #5802

### DIFF
--- a/backend/server/services/remote/models/migration.go
+++ b/backend/server/services/remote/models/migration.go
@@ -101,6 +101,28 @@ func (o DropTableOperation) Execute(basicRes context.BasicRes) errors.Error {
 
 var _ Operation = (*DropTableOperation)(nil)
 
+type RenameColumnOperation struct {
+	Table   string `json:"table"`
+	OldName string `json:"old_name"`
+	NewName string `json:"new_name"`
+}
+
+func (o RenameColumnOperation) Execute(basicRes context.BasicRes) errors.Error {
+	db := basicRes.GetDal()
+	if !db.HasColumn(o.Table, o.OldName) {
+		return nil
+	}
+	if db.HasColumn(o.Table, o.NewName) {
+		err := db.DropColumns(o.Table, o.NewName)
+		if err != nil {
+			return err
+		}
+	}
+	return db.RenameColumn(o.Table, o.OldName, o.NewName)
+}
+
+var _ Operation = (*RenameTableOperation)(nil)
+
 type RenameTableOperation struct {
 	OldName string `json:"old_name"`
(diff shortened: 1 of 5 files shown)


backend/server/services/remote/models/migration.go
```diff
@@ -101,6 +101,28 @@ func (o DropTableOperation) Execute(basicRes context.BasicRes) errors.Error {
...
+	NewName string `json:"new_name"`
+}
+
+func (o RenameColumnOperation) Execute(basicRes context.BasicRes) errors.Error {
+	db := basicRes.GetDal()
+	if !db.HasColumn(o.Table, o.OldName) {
+		return nil
+	}
+	if db.HasColumn(o.Table, o.NewName) {
+		err := db.DropColumns(o.Table, o.NewName)
```
keon94, line 116: why should we drop the column if it exists? That'll cause data loss. Isn't just doing RenameColumn() enough?

keon94: in this scenario specificially, if the column with the correct (new) name already exists we should just immediately return nil, no?

klesh:
Did you test it on a new deployment of v0.18 without upgrading?

klesh:
@CamilleTeruel Another potential scenario is :
1. Users may deploy the devlake `v0.18.0-beta1`: so the column name is like `_tool_azuredevops_jobs.start_time`
2. Upgrade to the latest with your fix
Will it work?

CamilleTeruel:
@klesh That migration from `startTime` to `start_time` is not from this PR, it is just refactored to use a `rename_column` operation instead of raw SQL.

klesh (approved):

PR #5977, opened 2023-08-30
feat: allow users to configure the sync modes for plugins that suppor…

### LINKED ISSUE #5841: [Feature][Data collection] Allow users to configure the sync modes for plugins that support incremental sync
### Search before asking

### Use case

Some users found out that after an issue was deleted in Jira, the issue would not be deleted in DevLake after re-syncing the data, as Jira used the `incremental sync` by default to improve the speed. 

However, those users want to get rid of the deleted issues in DevLake.

### Description

A solution to this is to: 

- [x] Provide the configuration for users to choose which sync modes to use. According to the [supported data sources](https://devlake.apache.org/docs/next/Overview/SupportedDataSources), plugins that support the incremental sync should be supported:
   - [x] GitHub
   - [x] GitLab
   - [x] Jira
   - [x] TAPD
   - [x] PagerDuty
- [ ] Allow users to trigger an `empty and re-collect data` operation for a 'incremental sync' mode blueprint. Users, especially those with a large data size don't need to fully refresh data every time as it takes time, they just need to remove the dirty data at a certain point, for example, the time before monthly/annual reporting.

### Related issues

_No response_

### Code of Conduct

abeizn:
### Summary
Allow users to configure the sync modes for plugins that support incremental sync.

**### don't merge before v0.19.0-beta1**

Closes #5841

### Screenshots
**github graphql:**

_pre:_
6c0f4b8f-cb15-4a46-9f04-385987bdd487
img_v2_fefeeeef-79bb-4c88-aa4b-291f4386b13g
_now:_
img_v2_4fb3176d-4331-4504-a3c6-3efd53d9b4ag
d876520b-6199-413b-a0e5-7714c502ec1a

**jira:**
_pre:_
img_v2_413f18df-8644-40ff-a640-a42e897a0ffg
img_v2_dea8c77d-191b-402a-b2a8-0626fe6c794g
_now:_
31c2cfa3-0af3-4666-a050-ec8c6f9b2e7e
270da104-c0d0-42a7-8ac7-940432258d44

### Other Information


backend/helpers/pluginhelper/api/api_collector_with_state.go
```diff
@@ -153,10 +156,12 @@ func NewStatefulApiCollectorForFinalizableEntity(args FinalizableApiCollectorArg
...
 		return nil, err
 	}
-
 	// // prepare the basic variables
-	var isIncremental = manager.IsIncremental()
 	var createdAfter *time.Time
+	var isIncremental = manager.IsIncremental()
+	if manager.Ctx.TaskContext().FullSync() {
+		isIncremental = false
+	}
```
klesh, line 164: I think it is better to put it into `manager.IsIncremental`

klesh (commented):
I feel like the `IsFullSync` should be a part of [SyncPolicy](https://github.com/apache/incubator-devlake/blob/f81f96ce8591ce74ab30409b763eb662ed747c65/backend/core/models/blueprint.go#L86) (The `Blueprint` prefix should be removed )

backend/helpers/pluginhelper/api/api_collector_with_state.go
abeizn: fixed

abeizn:
> I feel like the IsFullSync should be a part of [SyncPolicy](https://github.com/apache/incubator-devlake/blob/f81f96ce8591ce74ab30409b763eb662ed747c65/backend/core/models/blueprint.go#L86) (The Blueprint prefix should be removed )

I don’t think it’s necessary. If it’s not packaged with struct, it will only add one more layer when passing parameters. Others seem useless at the moment.

PR #5983, opened 2023-08-30
fix: azure devops db connection fail

### LINKED ISSUE #5978: [Bug]Unable to make database connection
### Search before asking

### What happened

When I run the extractor it doesn't work, I get an error regarding database connection.

image

Since the 'lake' pod only starts if the db connection works:

```
  initContainers:
    - name: waiting-database-ready
      image: alpine:3.16
      command:
        - sh
        - '-c'
        - |
          until nc -z -w 2 devlake-mysql 3306 ; do
            echo wait for database ready ...
            sleep 2
          done
          echo database is ready
```

I cross checked and it seems that everything is fine (DB_URL).

I upgraded to 0.18-beta6 and it worked for two to three runs and then...this.

Any ideas what else to check?

The run should be executed successfully.

### How to reproduce

Upgrade from 0.17 to 0.18-beta8, establish a project and rerun.

### Anything else

_No response_

### Version

v0.18-beta8

### Code of Conduct

abeizn:
### Summary
fix: azure devops db connection fail

Closes #5978 
Relates to https://github.com/apache/incubator-devlake-helm-chart/pull/175/

### Screenshots
image

### Other Information

### DIFF
--- a/backend/python/pydevlake/pydevlake/ipc.py
+++ b/backend/python/pydevlake/pydevlake/ipc.py
@@ -129,6 +129,8 @@ def create_db_engine(db_url) -> Engine:
     connect_args = dict(parse_qsl(urlparse(db_url).query))
     if 'parseTime' in connect_args:
         del connect_args['parseTime']
+    if 'loc' in connect_args:
+        del connect_args['loc']
     try:
         engine = create_engine(base_url, connect_args=connect_args)
         tables = SubtaskRun.metadata.tables


backend/python/pydevlake/pydevlake/ipc.py
```diff
@@ -129,6 +129,8 @@ def create_db_engine(db_url) -> Engine:
     connect_args = dict(parse_qsl(urlparse(db_url).query))
     if 'parseTime' in connect_args:
         del connect_args['parseTime']
+    if 'loc' in connect_args:
+        del connect_args['loc']
```
klesh, line 133: just delete? shouldnt we convert ？

klesh (approved):

PR #6040, opened 2023-09-08
fix: check both update time and record count when judge skippable

### LINKED ISSUE #6039: [Bug][plugin] starrocks plugin doesn't handle deleted records when incremental updating
### Search before asking

### What happened

StarRocks plugin, which exports source data to StarRocks, supports incremental mode to sync the table which has updated records.

Currently, timestamp of updating is the label. It could only cover the scenario of UPDATE and INSERT. If records were deleted, the plugin would still skip this table and StarRocks would contain dirty data compared to the source.

Would trigger the process when the changing of table is record deletion.

### How to reproduce

1. trigger a pipeline with StarRocks plugin plan which set `update_column`, making target StarRocks database has the source records
2. delete a record from source data
3. trigger the above pipeline again

### Anything else

_No response_

### Version

37065158d80a2d67c66f77699c53feb9cdaf64f1

### Code of Conduct

usharerose:
### Summary
add an additional condition when judge exporting to StarRocks skippable
* whether the record of source and destination is equal or not

Closes #6039

### Screenshots
N/A

### Other Information
N/A

### DIFF
--- a/backend/plugins/starrocks/tasks/tasks.go
+++ b/backend/plugins/starrocks/tasks/tasks.go
@@ -180,7 +180,19 @@ func createTmpTableInStarrocks(dc *DataConfigParams) (map[string]string, string,
 				}
 			} else {
 				if updatedFrom.Equal(updatedTo) {
-					return nil, "", true, nil
+					sourceCount, err := db.Count(dal.From(table))
+					if err != nil {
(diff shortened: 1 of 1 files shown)


abeizn (approved):
lgtm

PR #6050, opened 2023-09-11
feat(webhook): plugin webhook's detail and list api can get api key info now

d4x1:
### ⚠️ Pre Checklist

> Please complete _ALL_ items in this checklist, and remove before submitting

- [x] I have read through the [Contributing Documentation](https://devlake.apache.org/community/).
- [x] I have added relevant tests.
- [x] I have added relevant documentation.
- [x] I will add labels to the PR, such as `pr-type/bug-fix`, `pr-type/feature-development`, etc.

### Summary
What does this PR do?
1. add ApiKey field for plugin webhook's list and detail APIs, in order to make it possible to regenerabe webhook's api key with the api `HTTP PUT/api-keys/:apiKeyId/`.
2. Simplify the response struct to make code less.
3. add fields `postIssuesEndpoint` `closeIssuesEndpoint` `postPipelineTaskEndpoint` `postPipelineDeployTaskEndpoint` `closePipelineEndpoint`  to help to simplify the front end codes.

### Does this close any open issues?
Closes xx

### Screenshots
Include any relevant screenshots here.
1.
image
image

3. 
image

### Other Information
Any other information that is important to this PR.

PR #6054, opened 2023-09-11
cherry-pick feat(webhook): plugin webhook's detail and list api can get api key info now

d4x1:
cherry-pick #6050 feat(webhook): plugin webhook's detail and list api can get api key info now


mintsweet:
This PR will closed.
#6050 doesn't need to cherry-pick.

PR #6064, opened 2023-09-12
refactor: syncPolicy timeAfter

### LINKED ISSUE #5841: [Feature][Data collection] Allow users to configure the sync modes for plugins that support incremental sync

abeizn:
### Summary
adjust syncPolicy timeAfter

Closes #5841 

### Screenshots

### Other Information


backend/core/runner/run_task.go
```diff
@@ -54,8 +54,16 @@ func RunTask(
 		return err
 	}
 	syncPolicy := &models.SyncPolicy{}
-	syncPolicy.SkipOnFail = dbPipeline.SkipOnFail
-	syncPolicy.FullSync = dbPipeline.FullSync
+	blueprint := &models.Blueprint{}
```
klesh, line 57: This doesn't seem right, Why do we need to access the bp when running a Task? Shouldn't we access the `dbPipeline.SyncPolicy` instead?

backend/plugins/ae/ae.go
```diff
@@ -30,12 +30,13 @@ func main() {
 	aeCmd := &cobra.Command{Use: "ae"}
 	connectionId := aeCmd.Flags().Uint64P("Connection-id", "c", 0, "ae connection id")
 	projectId := aeCmd.Flags().IntP("project-id", "p", 0, "ae project id")
+	timeAfter := aeCmd.Flags().StringP("timeAfter", "a", "", "collect data that are created after specified time, ie 2006-01-02T15:04:05Z")
```
klesh, line 33: Should be `time-after` instead of `timeAfter` because the previous line is using `project-id`

backend/plugins/bamboo/bamboo.go
```diff
@@ -31,6 +31,7 @@ func main() {
 	bambooCmd := &cobra.Command{Use: "bamboo"}
 	connectionId := bambooCmd.Flags().Uint64P("Connection-id", "c", 0, "bamboo connection id")
 	projectKey := bambooCmd.Flags().StringP("project-key", "p", "", "bamboo project key")
+	timeAfter := bambooCmd.Flags().StringP("timeAfter", "a", "", "collect data that are created after specified time, ie 2006-01-02T15:04:05Z")
```
klesh, line 34: Same as the above

backend/plugins/bitbucket/tasks/api_common.go
```diff
@@ -100,11 +101,12 @@ func GetQueryCreatedAndUpdated(fields string, collectorWithState *api.ApiCollect
 		}
 		query.Set("fields", fields)
 		query.Set("sort", "created_on")
+		syncPolicy := collectorWithState.Ctx.TaskContext().SyncPolicy()
 		if collectorWithState.IsIncremental() {
 			latestSuccessStart := collectorWithState.LatestState.LatestSuccessStart.Format(time.RFC3339)
 			query.Set("q", fmt.Sprintf("updated_on>=%s", latestSuccessStart))
-		} else if collectorWithState.TimeAfter != nil {
-			timeAfter := collectorWithState.TimeAfter.Format(time.RFC3339)
+		} else if syncPolicy != nil && syncPolicy.TimeAfter != nil {
```
klesh, line 108: Doesn't seem correct.
`collectorWithState.TimeAfter` may be later than `syncPolicy.TimeAfter` because of `incremental sync`, the `collectorWithState.TimeAfter` would be updated automatically after each data collection while `syncPolicy.TimeAfter` can only be updated by the user

backend/core/runner/run_task.go
abeizn: fixed

backend/plugins/ae/ae.go
abeizn: fixed

backend/plugins/bamboo/bamboo.go
abeizn: fixed

backend/plugins/bitbucket/tasks/api_common.go
abeizn: fixed

backend/helpers/pluginhelper/api/api_collector_with_state.go
```diff
@@ -154,18 +160,19 @@ func NewStatefulApiCollectorForFinalizableEntity(args FinalizableApiCollectorArg
...
 
 	// // prepare the basic variables
+	syncPolicy := manager.Ctx.TaskContext().SyncPolicy()
 	var isIncremental = manager.IsIncremental()
 	var createdAfter *time.Time
 	if isIncremental {
 		createdAfter = manager.LatestState.LatestSuccessStart
-	} else {
-		createdAfter = manager.TimeAfter
+	} else if syncPolicy != nil && syncPolicy.TimeAfter != nil {
```
klesh, line 174: the change for this file should not be needed since https://github.com/apache/incubator-devlake/pull/6064/files#diff-c0bf74c35ffacb49d0e0b6a70207324ed5ae01fb1f4c841e2ebeb9bf6b0627c2R67 already takes care the `manager.TimeAfter`

abeizn: fixed

klesh (approved):

PR #6116, opened 2023-09-19
Fix 6112

### LINKED ISSUE #6112: [Bug][Pagerduty] Unable to collect incidents with API token
### Search before asking

### What happened

401 error when collecting incidents from pagerduty with a valid token specified

Being able to collect incidents. 
Instead we have 401 returned by this call:
https://api.pagerduty.com/incidents?date_range=all&limit=100&offset=0&service_ids%5B%5D=PKRB4TU&sort_by=created_at%3Adesc&total=true

The issue is the fact that DevLake is not setting the Authentication header correctly: 
In the test endpoint (that works correctly) is being set with 
`"Authorization": fmt.Sprintf("Token token=%s", connection.Token)`
In the "PrepareTaskData" instead is set as follow:
`"Authorization": fmt.Sprintf("Token %s", connection.Token)` 

### How to reproduce

Try to collect incidents from a pagerduty cloud instance with a token

### Anything else

_No response_

### Version

v0.18.0

### Code of Conduct

Marco-De-Stefani:
### Summary
Fixing #6112 and removing the typo in the version of golangci-lint 

Closes #6112

### Screenshots
Before:
image
After:
image


keon94:
Has PagerDuty made this change to their API recently? How come this wasn't an issue before?

Marco-De-Stefani:
@keon94 I think that Pagerduty API has not been changed, by having a quick look at the GH history I believe this has been introduced with the refactor of the Pagerduty plugin.
It looks to me that the first version of the plugin implemented in the revision eef06e96e95ad9e95fd9bc4dde27cc55123e0740 uses the correct token.
In that revision, the connection.go file contains --> 
Token string `mapstructure:"token" env:"PAGERDUTY_AUTH" validate:"required" encrypt:"yes"`

keon94:
@Marco-De-Stefani very odd. I tried running the pagerduty tests under `test/e2e/manual/pagerduty` with and without this change and it works for me, so I don't know what's going on. Either way, there's a better fix for this which I've pointed out in the file changes.

backend/plugins/pagerduty/impl/impl.go
```diff
@@ -109,7 +109,7 @@ func (p PagerDuty) PrepareTaskData(taskCtx plugin.TaskContext, options map[strin
 	}
 
 	client, err := helper.NewApiClient(taskCtx.GetContext(), connection.Endpoint, map[string]string{
```
keon94, line 111: replace all of this with:
```go
client, err := helper.NewApiClientFromConnection(taskCtx.GetContext(), taskCtx, connection)
```

keon94: This will utilize this [code](https://github.com/apache/incubator-devlake/blob/8bcacd8eeefe4c88d0a0e5135bd026feac94a5b3/backend/plugins/pagerduty/models/connection.go#L33) which already correctly sets the token.

Marco-De-Stefani:
> @Marco-De-Stefani very odd. I tried running the pagerduty tests under `test/e2e/manual/pagerduty` with and without this change and it works for me, so I don't know what's going on. Either way, there's a better fix for this which I've pointed out in the file changes.

I noticed that as well, it works cause at line 125 of the pagerduty_test.go file it's calling "PagerDutyAccessToken" which sets the token in the correct way :)

Marco-De-Stefani:
> > > @Marco-De-Stefani very odd. I tried running the pagerduty tests under `test/e2e/manual/pagerduty` with and without this change and it works for me, so I don't know what's going on. Either way, there's a better fix for this which I've pointed out in the file changes.
> > 
> > 
> > I noticed that as well, it works cause at line 125 of the pagerduty_test.go file it's calling "PagerDutyAccessToken" which sets the token in the correct way :)
> 
> That token object doesn't have an influence [here](https://github.com/apache/incubator-devlake/blob/73883730362d16148cee167cb0cd5f13edc28f62/backend/plugins/pagerduty/impl/impl.go#L111) though. The test still passes without "token=". Here's what my debugger captures - connection.Token is just the raw token. image

Weird, I'll try to debug that as well

keon94 (approved):
LGTM

klesh:
Hi, Nice work, thanks for your contribution, would you like to cherry-pick it back to [release v0.19](https://github.com/apache/incubator-devlake/tree/release-v0.19)?

PR #6126, opened 2023-09-20
feat(pagerduty): add pagerduty priority as issue severity

### LINKED ISSUE #6125: [Feature][Pagerduty] Add Pagerduty priority as a severity when parsing incidents
### Search before asking

### Use case

I would like to have the field called "priority" in Pagerduty inside the "severity" one of Devlake

As discussed here: [https://devlake-io.slack.com/archives/C03APJ20VM4/p1695132777495889](url) 

### Description

_No response_

### Related issues

_No response_

### Code of Conduct

Marco-De-Stefani:
### Summary
Adds pagerduty priority to the severity field of the Issue domain object in Devlake

Closes #6125 

### Screenshots
WIP


backend/plugins/pagerduty/models/incident.go
```diff
@@ -41,6 +41,7 @@ type (
 		Summary      string
 		Status       IncidentStatus  //acknowledged, triggered, resolved
 		Urgency      IncidentUrgency //high or low
+		Priority     string
```
keon94, line 44: you will need to add migrations so that the Incident table in the DB has a column for this field

backend/plugins/pagerduty/tasks/incidents_extractor.go
```diff
@@ -64,6 +64,13 @@ func ExtractIncidents(taskCtx plugin.SubTaskContext) errors.Error {
 				incident.ServiceId = service.Id
 				results = append(results, &service)
 			}
+			if incidentRaw.Priority != nil {
```
keon94, line 67: Going by [the API docs] https://developer.pagerduty.com/api-reference/0fa9ad52bf2d2-list-prioritiesm I think it's better to just use the `name` field on the priorities instead of `summary` or `ID`. `name` can't be null per the doc and it'll make the mapping more robust and predictable.

Marco-De-Stefani: solved & updated the fields returned by pagerduty

keon94 (approved):

Issue #5519, opened 2023-06-19
[Refactor][Framework] Unify Helpers into a single one


keon94:
## What and why to refactor
What are you trying to refactor? Why should it be refactored now?
Unify ConnectionHelper, ScopeConfigHelper and ScopeHelper into a single Helper. Have the plugin interface expose a getHelper() function that returns an instance of this helper. This will make life easy. We will have access a plugin's helper from anywhere in the code given its name.

## Describe the solution you'd like
How to refactor?
Mentioned above

## Related issues
Please link any other

klesh:
Nice, I would prefer the new helper to be called `DataSourceHelper`.

keon94:
@klesh looks like this is done? Can we close it?

klesh:
@keon94 Thanks for asking. There are a couple of plugins left to be updated, I will close it when done.

PR #6143, opened 2023-09-25
refactor: extract scope common fields to `common.Scope`

klesh:
### Summary

Previously, each plugin had to define the columns `ConnectionId` and `ScopeConfigId` on its own, which is tedious and error-prone, not to mention the maintainability overhead.

This PR tries to fix the problem by introducing a new base struct for all scopes across data source plugins named `common.Scope`

Part of #5519


backend/core/models/common/base.go
```diff
@@ -85,10 +79,31 @@ func NewNoPKModel() NoPKModel {
...
+
+func (s Scope) ScopeScopeConfigId() uint64 {
+	return s.ScopeConfigId
+}
+
+type ScopeConfig struct {
+	Model
+	Entities     []string `gorm:"type:json;serializer:json" json:"entities" mapstructure:"entities"`
+	ConnectionId uint64   `json:"connectionId" gorm:"index" validate:"required" mapstructure:"connectionId,omitempty"`
+	Name         string   `mapstructure:"name" json:"name" gorm:"type:varchar(255);index:idx_name_github,unique" validate:"required"`
```
abeizn, line 101: Is this（idx_name_github） a suitable name?

abeizn (dismissed):

abeizn (approved):

PR #6146, opened 2023-09-25
refactor: unify connection/scope/scopeConfig helpers

klesh:
### Summary

Previously, the connection/scope/scopeConfig were written at different times by multiple people so the code is kind of messy.

This PR tries to unify them in a more readable and maintainable way:

1. introduce `service helper` and `api helper`, the former is designed to help with business logic, like loading from the database, creating, updating, and validating, while the latter acts as the adapter between Http Requests and Services
2. introduce `model service helper` which implements some basic business logic shared by all models so you may nest the helper within your own service class and use the functionalities in any way you want including exposing them directly or overriding them when needed.
3. connection / scope / scope helpers are organized in a unified way, all are offered both a `service helper` and an `api helper`, the former helper can be used in any package as you see fit, while the latter `api helper` should be used only in the `api/` package.
4. this PR only updates part of the `github` plugins as a PoC, I will work on other after the PR merge

Part of #5519

### Screenshots

Note that there is a API change, @mintsweet 

#### Scope list api
sope list

#### Scope detail api
sope-detail


abeizn:
By the way, have you tried the compatibility under the pg database?

backend/helpers/srvhelper/model_service_helper.go
```diff
@@ -0,0 +1,212 @@
...
+	return err
+}
+
+// Update validates given model and update it into database if validation passed
+func (srv *ModelSrvHelper[M]) Update(model *M) errors.Error {
+	err := srv.ValidateModel(model)
+	if err != nil {
+		return err
+	}
+	if err != nil && srv.db.IsDuplicationError(err) {
```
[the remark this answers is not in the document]
klesh: fixed

klesh:
@abeizn what are referring to? the `e2e-postgres` passed, does it count?

backend/helpers/pluginhelper/api/model_api_helper.go
```diff
@@ -0,0 +1,160 @@
...
+	dalHelper      *srvhelper.ModelSrvHelper[M]
+	log            log.Logger
+	modelName      string
+	pkPathVarNames []string
+}
+
+func NewModelApiHelper[M dal.Tabler](
+	basicRes context.BasicRes,
+	dalHelper *srvhelper.ModelSrvHelper[M],
+	pkPathVarNames []string, // path variable names of primary key
```
d4x1, line 44: 1. Does variables' order matter?
2. `primary key` should be  plural form `primary keys`.

backend/helpers/srvhelper/connection_service_helper.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (connSrv *ConnectionSrvHelper[C, S, SC]) DeleteConnection(connection *C) (refs *DsRefs, err errors.Error) {
+	err = connSrv.ModelSrvHelper.NoRunningPipeline(func(tx dal.Transaction) errors.Error {
+		// make sure no blueprint is using the connection
+		connectionId := (*connection).ConnectionId()
+		refs, err = toDsRefs(connSrv.getAllBlueprinsByConnection(connectionId))
+		if err != nil {
+			return err
+		}
+		scopeCount := errors.Must1(connSrv.db.Count(dal.From(new(S)), dal.Where("connection_id = ?", connectionId)))
```
d4x1, line 57: I saw many codes use `errors.Must1` or `errors.Must`, which means if something goes wrong, the program will panic.  Is it as expected?

backend/helpers/srvhelper/dsrefs.go
```diff
@@ -0,0 +1,46 @@
...
+		for _, bp := range blueprints {
+			blueprintNames = append(blueprintNames, bp.Name)
+			if bp.ProjectName != "" {
+				projectNames = append(projectNames, bp.ProjectName)
+			}
+		}
+		return &DsRefs{
+			Blueprints: blueprintNames,
+			Projects:   projectNames,
+		}, errors.Conflict.New("Cannot delete record because it is referenced by blueprints")
```
d4x1, line 43: 1. From the error message, we cannot get infomation like: which record is referened by which blueprint.
2. This function seems to convert  blueprints to something else,  there is no logical about `delete record`.

Please consider to update the error message, or the funtion's name.

backend/plugins/github/api/scope_api.go
```diff
@@ -106,5 +97,6 @@ func GetScope(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors
 // @Failure 500  {object} shared.ApiBody "Internal Error"
 // @Router /plugins/github/connections/{connectionId}/scopes/{scopeId} [DELETE]
 func DeleteScope(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
-	return scopeHelper.Delete(input)
+	println("hello delete scope")
```
abeizn, line 100: please detele it

backend/plugins/github/api/scope_config_api.go
```diff
@@ -49,8 +51,10 @@ func CreateScopeConfig(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutpu
 // @Failure 400  {object} shared.ApiBody "Bad Request"
 // @Failure 500  {object} shared.ApiBody "Internal Error"
 // @Router /plugins/github/connections/{connectionId}/scope-configs/{id} [PATCH]
-func UpdateScopeConfig(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
-	return scHelper.Update(input)
+func PatchScopeConfig(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
+	println("api hello")
+	fmt.Printf("api patch scope confg %v\n", input.Params)
```
abeizn, line 56: same above

backend/helpers/srvhelper/model_service_helper.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/apache/incubator-devlake/core/dal"
+	"github.com/apache/incubator-devlake/core/errors"
+	"github.com/apache/incubator-devlake/core/log"
+	"github.com/apache/incubator-devlake/core/models"
+	"github.com/apache/incubator-devlake/helpers/dbhelper"
+	"github.com/go-playground/validator/v10"
+)
+
+type CustomValidator interface {
+	CustomValidate(entitiy interface{}, validate *validator.Validate) errors.Error
```
d4x1, line 34: 1. typo `entitiy`.
2. Param entitiy is unnecessary, because  a struct who implement `CustomValidator` will hold itself and can validate itself.
3. Param validate is unnecessary too, a struct who implement `CustomValidator` can validate without any validator.
4. please consider renaming `CustomValidator` to `Validate`.

backend/helpers/srvhelper/model_service_helper.go
```diff
@@ -0,0 +1,215 @@
...
+		return err
+	}
+	return err
+}
+
+// Update validates given model and update it into database if validation passed
+func (srv *ModelSrvHelper[M]) Update(model *M) errors.Error {
+	err := srv.ValidateModel(model)
+	if err != nil {
+		if srv.db.IsDuplicationError(err) {
```
d4x1, line 116: Can update return db's duplicated error?

backend/helpers/srvhelper/model_service_helper.go
```diff
@@ -0,0 +1,215 @@
...
+			return nil, errors.NotFound.Wrap(err, fmt.Sprintf("%s not found", srv.modelName))
+		}
+		return nil, err
+	}
+	return model, nil
+}
+
+// GetAll returns all models from database
+func (srv *ModelSrvHelper[M]) GetAll() ([]*M, errors.Error) {
+	array := make([]*M, 0)
```
d4x1, line 156: `var array []*M` is preffered.

backend/helpers/srvhelper/model_service_helper.go
```diff
@@ -0,0 +1,215 @@
...
+	return array, srv.db.All(&array)
+}
+
+func (srv *ModelSrvHelper[M]) GetPage(pagination *Pagination, query ...dal.Clause) ([]*M, int64, errors.Error) {
+	query = append(query, dal.From(new(M)))
+	// process keyword
+	searchTerm := pagination.SearchTerm
+	if searchTerm != "" && len(srv.searchColumns) > 0 {
```
d4x1, line 164: This will cause sql injection by splicing sql mauanlly, especiall searchterm is from client side.

backend/helpers/srvhelper/scope_config_service_helper.go
```diff
@@ -0,0 +1,66 @@
...
+		dal.Where("connection_id = ?", connectionId),
+		dal.Orderby("id DESC"),
+	)
+	return scopeConfigs, err
+}
+
+func (scopeConfigSrv *ScopeConfigSrvHelper[C, S, SC]) DeleteScopeConfig(scopeConfig *SC) (refs []*S, err errors.Error) {
+	err = scopeConfigSrv.ModelSrvHelper.NoRunningPipeline(func(tx dal.Transaction) errors.Error {
+		// make sure no scope is using the scopeConfig
+		sc := (*scopeConfig)
```
d4x1, line 54: `sc := (*scopeConfig)`, why parentheses?

backend/helpers/pluginhelper/api/connection_auths.go
```diff
@@ -144,6 +144,11 @@ func (ma *MultiAuth) SetupAuthenticationForConnection(connection plugin.ApiConne
 	return apiAuthenticator.SetupAuthentication(req)
 }
 
+func (ma *MultiAuth) CustomValidate(connection interface{}, v *validator.Validate) errors.Error {
+	return ma.ValidateConnection(connection, v)
+}
+
+// TODO: deprecated, rename to CustomValidate instead
```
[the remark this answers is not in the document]
klesh: I would love that if it can work. But, unfortunately, No, it will cause the linting to fail.

backend/helpers/pluginhelper/api/ds_connection_api_helper.go
```diff
@@ -0,0 +1,76 @@
...
+	"strconv"
+
+	"github.com/apache/incubator-devlake/core/context"
+	"github.com/apache/incubator-devlake/core/errors"
+	"github.com/apache/incubator-devlake/core/plugin"
+	"github.com/apache/incubator-devlake/helpers/srvhelper"
+	"github.com/apache/incubator-devlake/server/api/shared"
+)
+
+// DsConnectionApiHelper
```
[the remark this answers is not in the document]
klesh: short for DataSource

backend/helpers/pluginhelper/api/ds_connection_api_helper.go
```diff
@@ -0,0 +1,76 @@
...
+type DsConnectionApiHelper[C plugin.ToolLayerConnection, S plugin.ToolLayerScope, SC plugin.ToolLayerScopeConfig] struct {
```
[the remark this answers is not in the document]
klesh: How do you suggest to address that?

backend/helpers/srvhelper/model_service_helper.go
klesh: 2. No, it is not. For example, the `CustomValidate` may be implemented in a **nested struct** of the `model` and sometimes we need to check the whole `model` instead of the **nested struct** only. https://github.com/apache/incubator-devlake/blob/90cb26bd62d226f5c4ed54c17689dab9bcf30de0/backend/helpers/pluginhelper/api/connection_auths.go#L147
3. Passing the `*validate` can be beneficial in terms of performance because the cached metadata in Validator can be reused.
4. the function is meant for plugins to take over the Validation when needed, if plugins implement this interface, the helper would skip its own implementation and let the plugin do what it wants. I think `CustomValidate` is more specify and reflects its purpose better.

backend/helpers/srvhelper/model_service_helper.go
klesh: Yes, absolutely. One may update the model's unique index (e.g. name) causing conflict with other rows. it is not unusal.

backend/helpers/srvhelper/model_service_helper.go
klesh: Any proof? I thought the `golangci-lint` would complain if this is not a good practice.

backend/helpers/srvhelper/model_service_helper.go
klesh: Are you sure? would you like to take another closer look? The `searchTerm` is prefixed and suffixed with the `%` and then passed down to `gorm` as a SQL parameters, it should be safe from SQL injection IMO.

backend/plugins/github/api/init.go
```diff
@@ -18,69 +18,44 @@ limitations under the License.
...
 	connectionHelper = api.NewConnectionHelper(
 		basicRes,
 		vld,
 		p.Name(),
 	)
-	params := &api.ReflectionParameters{
-		ScopeIdFieldName:     "GithubId",
-		ScopeIdColumnName:    "github_id",
-		RawScopeParamName:    "Name",
-		SearchScopeParamName: "name",
```
abeizn, line 53: If these are removed, how to implement the data scope search function. I don't see any assignment operations to these anywhere else?

image

backend/helpers/pluginhelper/api/ds_helper.go
```diff
@@ -0,0 +1,49 @@
...
+
+package api
+
+import (
+	"github.com/apache/incubator-devlake/core/context"
+	"github.com/apache/incubator-devlake/core/plugin"
+	"github.com/apache/incubator-devlake/helpers/srvhelper"
+)
+
+func NewDataSourceHelpers[
```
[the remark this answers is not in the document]
klesh: It is returning 6 helpers, 3 `serviceHelper` and 3 `apiHelper`, the former ones can be used in anywhere in the plugin like `tasks` package or `api` package while the latter ones are only for `api` package. It makes no sense for `tasks` package to use `apiHelper`, that is why they are nested inside a struct.

backend/helpers/pluginhelper/api/model_api_helper.go
```diff
@@ -0,0 +1,160 @@
...
+		if err != nil {
+			return nil, errors.BadInput.Wrap(err, fmt.Sprintf("failed to save item %d", i))
+		}
+	}
+	return &plugin.ApiResourceOutput{
+		Body: req.Data,
+	}, nil
+}
+
+func ParsePagination[P any](input *plugin.ApiResourceInput, query ...dal.Clause) (*P, errors.Error) {
```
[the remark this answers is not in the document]
klesh: 1. Good catch.
2. It is a function, so? What is the point exactly? Please elaborate.

backend/helpers/pluginhelper/api/ds_scope_api_helper.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+func (connApi *DsScopeApiHelper[C, S, SC]) GetPage(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
+	pagination, err := ParsePagination[srvhelper.ScopePagination](input)
+	if err != nil {
+		return nil, errors.BadInput.Wrap(err, "failed to decode pathvars into pagination")
+	}
+	scopes, count, err := connApi.ScopeSrvHelper.GetScopesPage(pagination)
+	if err != nil {
```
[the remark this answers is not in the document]
klesh: What additional information do you think is necessary?

backend/helpers/pluginhelper/api/ds_scope_api_helper.go
```diff
@@ -0,0 +1,99 @@
...
+	return &plugin.ApiResourceOutput{
+		Body: map[string]interface{}{
+			"count":  count,
+			"scopes": scopes,
+		},
+	}, nil
+}
+
+func (connApi *DsScopeApiHelper[C, S, SC]) PutMultiple(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
+	// fix data[].connectionId
```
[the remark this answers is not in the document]
klesh: Not, it is not extracting, it is fixing the request body by putting the `connectionId` from the request URL path into the body.

backend/helpers/pluginhelper/api/ds_scope_api_helper.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (connApi *DsScopeApiHelper[C, S, SC]) Delete(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
+	var scope *S
+	scope, err := connApi.FindByPk(input)
+	if err != nil {
+		return nil, err
+	}
+	// time.Sleep(1 * time.Minute) # uncomment this line if you were to verify pipelines get blocked while deleting data
+	// check referencing blueprints
+	refs, err := connApi.ScopeSrvHelper.DeleteScope(scope, input.Query.Get("delete_data_only") == "true")
```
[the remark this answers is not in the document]
klesh: I don't understand, what is a magic variable.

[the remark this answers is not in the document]
d4x1: > I don't understand, what is a magic variable.

magic variable is not a official expression, in fact I mean `"delete_data_only"` in ths line.
Your reply make me realise that it is not a variable, and after searching the project, I find `delete_data_only ` is a common parameter in plugin's API. 

So just ignore this comment and forget it.

backend/helpers/pluginhelper/api/model_api_helper.go
klesh: 1. No, it doesn't, it uses Name 
2. No, `primary key` is always singular or it is not **Primary**. A primary key with multiple columns is called a Composite Primary Key (Singular)

backend/helpers/srvhelper/connection_service_helper.go
klesh: Yes, it should panic because it is **Unrecoverable**

backend/helpers/pluginhelper/api/ds_connection_api_helper.go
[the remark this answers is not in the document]
d4x1: I am not good at golang's generics. So my code will be like:
```golang
type DsConnectionApiHelper[Connection plugin.ToolLayerConnection, Scope plugin.ToolLayerScope, ScopeConfig plugin.ToolLayerScopeConfig] struct {
```

backend/helpers/pluginhelper/api/model_api_helper.go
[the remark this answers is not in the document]
d4x1: Maybe the first choice is best and simple.

backend/plugins/github/api/init.go
klesh: It is handled here https://github.com/apache/incubator-devlake/pull/6146/files#diff-48658af240a5af910b7bb5d4e3effb2a0a55e4433b57d874921b76ef9f0ce51eR58

backend/helpers/srvhelper/model_service_helper.go
d4x1: Forget it, it doesn't matter

backend/plugins/github/api/init.go
abeizn: image

Strangely, shouldn't this be ‘name’, not ‘full_name’?

backend/helpers/pluginhelper/api/ds_connection_api_helper.go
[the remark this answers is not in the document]
klesh: Alright, lets keep it short for now.

backend/plugins/github/api/init.go
klesh: Why? `full_name` includes `name` so it is so much better the `name`, isn't it?

backend/helpers/pluginhelper/api/ds_scope_api_helper.go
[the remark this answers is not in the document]
klesh: I can't think of any case these information could possibly help when an error occurs:
1. `page size`: it is already in the request query string, either it is illegal which is handled by the `parsePagination`, or Out-Of-Range which should be handled in the `parsePagination` function. In any case, `page size` is a 400 User Input Error. outputting this information doesn't seem to be helpful.
2. `page No`, same as above
3. `types of generics`:  it is too obvious by looking at the `URL`. 

In general, I don't think outputting any of the above variables would help, would you like to show a concrete example? Thanks.

In the meantime, I would update the `parsePagination` to verify `page size` and `page no`

backend/plugins/github/api/init.go
abeizn: e..., you are right.

backend/helpers/pluginhelper/api/ds_scope_api_helper.go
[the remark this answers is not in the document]
d4x1: You are right. Generally speaking, when logging something, we should log not only the error message but also the context. But as you said, we can get enough context from other place, so that's enough. 

Thanks for your reply.

abeizn (approved):

PR #6188, opened 2023-10-07
refactor(zentao): add custom types OperatedBy and StringFloat64 to simplify codes

d4x1:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
1. The progress field in response of zentao's api is a string or a float, at #6089  we add a progressRes field to receive this field and convert it into progress(float64). 
2. So do `ClosedBy` and `CanceledBy` fields.

Closes None

### Screenshots
The main branch cannot running, so test is on the way.

### Other Information


d4x1:
Here is an example for `closedBy` field:
```json
{
    "closedBy": {
        "id": 1,
	"account": "admin",
	"avatar": "",
	"realname": "admin"
    },
    "others...": ""
}
```

backend/core/models/common/string_float64.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"database/sql/driver"
+	"encoding/json"
+	"fmt"
+	"github.com/spf13/cast"
+)
+
+type StringFloat64 struct {
+	v float64
+	t string
```
klesh, line 29: I did the same for the `iso8061time`, which was never really useful. 
Should we ignore the original type to make it simpler?

d4x1: The original type is useless if we just want codes to run correctly, but if we define a new custom type, the original type is the basic component. I tend to keep type info in order to make the custom `StringFloat64` "unbroken".

klesh: Yup, I totally understand, that is why I did it for the `iso8601time`, like I said, it has never been used.
Seems like the `type` is only used in the  `Marshal` method which is actually not being called anywhere.

But, I'm OK with it for now.

klesh (approved):

PR #6194, opened 2023-10-08
fix: didgen should check pk values based on pk fields

klesh:
### Summary

Currently, the `didgen.Generate` implementation checks the pk values based on the given parameters which is obviously flawed.

This PR fixes the problem by checking them based on the Primary Key from the database against the given parameters.


d4x1 (approved):
Does this change affect the history data? I mean, if someone use the previous version, and upgrade to version that fix `domain_id_generator`, what will happen to data in tool layer or domain layer?

abeizn (approved):

abeizn:
> Does this change affect the history data? I mean, if someone use the previous version, and upgrade to version that fix `domain_id_generator`, what will happen to data in tool layer or domain layer?

domain layer data will be fully converted every time without affecting

PR #6234, opened 2023-10-12
fix(gitlab): fix failed task when collecting deployments

d4x1:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
1. Add `order_by` field with value "updated_at"  according to GitLab API's response.
2. Fix panic when generating primary keys using `didgen.NewDomainIdGenerator`.

Closes N/A.

### Screenshots
image

### Other Information


backend/plugins/gitlab/tasks/deployment_collector.go
```diff
@@ -60,8 +60,11 @@ func CollectDeployment(taskCtx plugin.SubTaskContext) errors.Error {
 			if err != nil {
 				return query, err
 			}
+
 			if collectorWithState.Since != nil {
 				query.Set("updated_after", collectorWithState.Since.Format(time.RFC3339))
+				// https://gitlab.com/gitlab-org/gitlab/-/issues/328500
+				query.Set("order_by", "created_at")
```
klesh, line 67: Should not be nested in the condition consider `collectorWithState.Since` could be `nil` for the very first collection and we still need the records to be sorted.

klesh (approved):

d4x1:
@klesh  it seems that we can only use `updated_at` sort in this case, because when I run with merged filels, I get an error like this: 
```
time="2023-10-16 15:18:33" level=error msg=" [pipeline service] [pipeline #9] [task #59] [api async client] Retry exceeded 3 times calling projects/50751699/deployments. The last error was: Http DoAsync error calling [GET projects/50751699/deployments]. Response: {\"message\":\"400 Bad request - `updated_at` filter requires `updated_at` sort\"} (400)\n\tWraps: (2) Http DoAsync error calling [GET projects/50751699/deployments]. Response: {\"message\":\"400 Bad request - `updated_at` filter requires `updated_at` sort\"} (400)\n\tError types: (1) *hintdetail.withDetail (2) *errors.errorString"
```

What do you suggest?

PR #6312, opened 2023-10-24
fix: bitbucket search remote scopes not functional

klesh:
### Summary
The existing search remote scopes of the Bitbucket plugin require `gid` parameter to work which is not practical.
This PR uses another API to achieve the goal, however, for the API to filter the result, the `role` parameter is required, so it is impossible to search PUBLIC Repos.

To avoid confusion, we should notify users on the config-ui that searching is only for "**repositories to which the user has explicit read access**"

### Screenshots

image


Startrekzky:
Please paraphrase the placeholder of the search input to: "Enter the keywords to search for repositories that you have read access"

abeizn (approved):

PR #6341, opened 2023-10-26
fix(config-ui): extra error when no env file

mintsweet:
### Summary
Fixed extra error when no env file

### Related PR
#6327


Startrekzky (approved):
LGTM

PR #6387, opened 2023-11-02
feat: set grafana home to the Homepage dashboard

klesh:
### Summary
Make the devlake Homepage dashboard as the default home page of Grafana.
So the **newly added users would land on the dashboard when they log into Grafana**

### Screenshots
image


Startrekzky (approved):
LGTM

PR #6406, opened 2023-11-07
fix: update github/gitlab pipeline/tasks result/status field

abeizn:
### Summary
fix: update github/gitlab pipeline/tasks result/status field

Closes na

### Screenshots
image
image

### Other Information


d4x1 (approved):
I think we should add screenshots to show result in GitLab plugin.

PR #6494, opened 2023-11-21
Complete #5643 Not returning the tokens or passwords on APIs

### LINKED ISSUE #5643: [Refactor][Auth] Not returning the tokens or passwords on APIs
## What and why to refactor

On the Connections Admin form, we can add tokens to access the tools, but those Tokens are returning in the API openly with no necessity and compromising the security of it.

The reason is in the companies where DevLake will be used, not all DevLake Admin are able to have full read-only access to those tools, but as we return it from the API side, it is available for them.

How to refactor?

Not returning the tokens or passwords on APIs, it should be treated as one way only, as soon it is saved on the DB just going to be updated if required by the interface.

## Related issues
Please link any other

d4x1:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
1. Not returning the tokens or passwords on APIs.
2. Add new API `HTTP POST /plugins/{pluginName}/{connectionId}/test`, help to make test connection more security.
3. Make sure PATCH connection request update the fields only in the payload (Config UI can just put changed fields to payload).
4. Fix some typos.

Closes #5643

### Screenshots

#### Sensitive fields are empty now
image

#### New API works
image

### Other Information


backend/plugins/ae/api/connection.go
```diff
@@ -32,6 +32,33 @@ type ApiMeResponse struct {
...
+		return &plugin.ApiResourceOutput{Body: true, Status: 200}, nil
+	case 401: // error secretKey or nonceStr
+		return &plugin.ApiResourceOutput{Body: false, Status: http.StatusBadRequest}, nil
+	default: // unknow what happen , back to user
+		return &plugin.ApiResourceOutput{Body: res.Body, Status: res.StatusCode}, nil
+	}
+}
+
+// TestConnection test ae connection
+// Deprecated
```
klesh, line 61: I think this method should be kept because users may want to test the connection before actually creating it.

backend/plugins/ae/impl/impl.go
```diff
@@ -139,6 +139,9 @@ func (p AE) ApiResources() map[string]map[string]plugin.ApiResourceHandler {
 			"PATCH":  api.PatchConnection,
 			"DELETE": api.DeleteConnection,
 		},
+		"connections/:connectionId/test": {
+			"POST": api.TestConnectionV2,
```
klesh, line 143: How about naming it as `TestExistingConnection` or `TestEncryptedConnection` since it can not replace the previous `TestConnection` function and they must co-exist in the future?

backend/plugins/ae/models/connection.go
```diff
@@ -65,6 +65,11 @@ func (AeConnection) TableName() string {
 	return "_tool_ae_connections"
 }
 
+func (connection AeConnection) CleanUp() AeConnection {
```
klesh, line 68: Will it be better to define a function named `SanatizeEncFields` that will iterate all fields with tag `gorm:"serializer:encdec"` including those inside nested structs and set them to empty strings?

klesh (commented):
I reviewed

backend/plugins/ae/api/connection.go
d4x1: Agreed. This comment should be removed.

backend/plugins/ae/impl/impl.go
d4x1: Agreed! V2 is not a good name.I'll update them later.

backend/plugins/ae/models/connection.go
d4x1: `SanatizeEncFields` is a good name, I agree with you.
As to "iterate all fields with tag gorm:"serializer:encdec" including those inside nested structs and set them to empty strings" , I don't know what to do. what are the functions input and output ?

klesh (dismissed):

klesh (approved):

PR #6530, opened 2023-11-28
refactor: new remote api helpers

klesh:
### Summary
1. added new remote api helpers
2. github uses new remote API helpers

Part of #5519


abeizn (approved):

backend/helpers/pluginhelper/api/ds_remote_api_proxy_api.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body, err := errors.Convert01(io.ReadAll(resp.Body))
+	if err != nil {
+		return nil, err
+	}
+	// verify response body is json
```
d4x1, line 104: If it is just a `Proxy`, maybe passing all remote transparently is a better choice.

backend/helpers/pluginhelper/api/ds_remote_api_scope_list_helper.go
```diff
@@ -0,0 +1,94 @@
...
+	listRemoteScopes DsListRemoteScopes[C, S, P]
+}
+
+// NewDsRemoteApiScopeListHelper creates a new DsRemoteApiScopeListHelper
+func NewDsRemoteApiScopeListHelper[
+	C plugin.ToolLayerApiConnection,
+	S plugin.ToolLayerScope,
+	P any,
+](
+	rap *DsRemoteApiProxyHelper[C],
```
d4x1, line 51: I think `rap`  can be renamed with something like `dsRemoteApiProxyHelper` or `remoteApiProxyHelper`, because the other param's name is `listRemoteScopes` not `lrs`.
Keeping them the same is better in my option.
Just advice.

backend/helpers/pluginhelper/api/ds_remote_api_scope_search_helper.go
```diff
@@ -0,0 +1,80 @@
...
+type DsRemoteApiScopeSearchHelper[C plugin.ToolLayerApiConnection, S plugin.ToolLayerScope] struct {
+	*DsRemoteApiProxyHelper[C]
+	searchRemoteScopes DsSearchRemoteScopes[C, S]
+}
+
+func NewDsRemoteApiScopeSearchHelper[
+	C plugin.ToolLayerApiConnection,
+	S plugin.ToolLayerScope,
+](
+	rap *DsRemoteApiProxyHelper[C],
```
d4x1, line 40: same as above.

backend/helpers/pluginhelper/api/ds_remote_api_scope_search_helper.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, err
+	}
+	params := &models.DsRemoteApiScopeSearchParams{
+		Page:     1,
+		PageSize: 50,
+	}
+	if e := mapstructure.Decode(input.Query, params); e != nil {
+		return nil, errors.BadInput.Wrap(e, "invalid params")
+	}
+	// err = utils.DecodeMapStruct(input.Query, params, true)
```
d4x1, line 62: should we keep these codes?

backend/helpers/pluginhelper/api/models/ds_remote_api_scope_search_params.go
```diff
@@ -0,0 +1,24 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package models
+
+type DsRemoteApiScopeSearchParams struct {
+	Search   string `json:"search" mapstructure:"search" validate:"required"`
```
d4x1, line 21: I know `Search` is used in many places. But `Keyword` may  be a better name. 

Just advice.

backend/helpers/pluginhelper/api/ds_remote_api_proxy_api.go
klesh: No, I don't think so, it is too risky, we should keep it minimal as long as it satisfies what we need.

backend/helpers/pluginhelper/api/models/ds_remote_api_scope_search_params.go
klesh: The naming came from the existing API definition.
We should not change the interface at will unless we can provide it is the right thing to do.

PR #6561, opened 2023-12-05
fix: github token expired may cause 500 error

klesh:
### Summary
fix: github token expired may cause 500 error

### Screenshots
Before
img_v3_025r_1586d741-70ce-44a1-a1da-0263d04e517g

After
image


mintsweet (approved):
LGTM.

PR #6628, opened 2023-12-14
Feat 6520

### LINKED ISSUE #6520: [Feature][Api] API to Provide the Most Recent Synchronization Time for a Data Scope
### Search before asking

### Use case

As a user of Apache DevLake, I want to have an API that can provide the most recent synchronization time for a data scope. This is important because a data scope can exist in multiple blueprints (BPs), and using the most recent synchronization time of a BP cannot accurately describe the most recent synchronization time of a data scope.

### Description

Currently, Apache DevLake allows us to synchronize data from various DevOps tools. However, there is a need for an API that can provide the most recent synchronization time for a data scope. This is because a data scope, which is the smallest collection of data that can be synchronized by DevLake (like a Jira board or a Jenkins job), can exist in multiple BPs. The synchronization time of a BP cannot accurately represent the synchronization time of a data scope, hence the need for this feature.

### Related issues

_No response_

### Code of Conduct

d4x1:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
1. Add new api that can fetch scope's latest sync state

Closes #6520 

### Screenshots
#### AzureDevops
Unsupported

#### Bamboo
image

#### BitBucket
image

#### GitHub
image

#### GitLab
image

#### Jenkins
image

#### JIRA
image

#### PagerDuty
image

#### Sonar Qube
image

#### TAPD
image

##### No History Pipeline
image

##### Has History Pipeline
image

### Other Information


backend/helpers/srvhelper/scope_service_helper.go
```diff
@@ -93,6 +93,44 @@ func (scopeSrv *ScopeSrvHelper[C, S, SC]) GetScopeDetail(includeBlueprints bool,
...
+func (scopeSrv *ScopeSrvHelper[C, S, SC]) GetScopeLatestSyncState(pkv ...interface{}) ([]*models.LatestSyncState, errors.Error) {
+	scope, err := scopeSrv.ModelSrvHelper.FindByPk(pkv...)
+	if err != nil {
+		return nil, err
+	}
+	s := *scope
+	params := plugin.MarshalScopeParams(s.ScopeParams())
+	scopeSrv.log.Debug("scope: %#+v, params: %+v", s, params)
+	rows, err := scopeSrv.db.Cursor(
+		dal.Select("raw_data_table, MAX(latest_success_start) as latest_success_start"),
```
klesh, line 105: I don't think group by `raw_data_table` and take the max of `latest_success_start` is needed because `raw_data_table`s  are unique when `raw_data_params` is specified.
I think simply select the needed field would be sufficient.

backend/helpers/srvhelper/scope_service_helper.go
```diff
@@ -93,6 +93,44 @@ func (scopeSrv *ScopeSrvHelper[C, S, SC]) GetScopeDetail(includeBlueprints bool,
...
+		dal.From("_devlake_collector_latest_state"),
+		dal.Where("raw_data_params = ?", params),
+		dal.Groupby("raw_data_table"),
+	)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	scopeSyncStates := []*models.LatestSyncState{}
```
klesh, line 114: Seems like we are returning the whole list anyway, I suggest that we use the `All` method instead of using `Cursor` to make it simpler and readable

d4x1: > Seems like we are returning the whole list anyway, I suggest that we use the `All` method instead of using `Cursor` to make it simpler and readable

Got it. I'll fix it later.

backend/helpers/srvhelper/scope_service_helper.go
d4x1: > I don't think group by `raw_data_table` and take the max of `latest_success_start` is needed because `raw_data_table`s are unique when `raw_data_params` is specified. I think simply select the needed field would be sufficient.

Got it. I'll fix it later.

backend/plugins/bitbucket/api/scope.go
```diff
@@ -82,6 +82,25 @@ func GetScopeList(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, er
...
+// @Summary process BitBucket repos get request
+// @Description process BitBucket repos get request
+// @Tags plugins/bitbucket
+// @Param connectionId path int true "connection ID"
+// @Param scopeId path string true "repo ID"
+// @Success 200  {object} ScopeRes
+// @Failure 400  {object} shared.ApiBody "Bad Request"
+// @Failure 500  {object} shared.ApiBody "Internal Error"
+// @Router /plugins/bitbucket/connections/{connectionId}/scopes/{scopeId} [GET]
+func GetScopeDispatcher(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
```
klesh, line 95: This seems hacky, why not add a new route and we can describe the response body correctly?

d4x1: Let me explain it in detail.
`ScopeID` in plugin bitbucket is a string with format `{string_a}/{string_b}`, eg `thenicetgp/ptest`, then the gin path will be `/plugins/bitbucket/connections/{connectionId}/scopes/{string_a}/{string_b}`, but `latest-sync-state` api should be `/plugins/bitbucket/connections/{connectionId}/scopes/{scope_id}/latest-sync-state`. Gin cannot handle it with params in path, more detail: https://github.com/gin-gonic/gin/issues/1446. 
So we have to handle it specially.

klesh: I see, I think we should chose either one of the following:
1. encode the `{string_a}/{string_b}` before sending it to the backend, cc @mintsweet 
2. define a new resource for that, e.g. `/plugins/bitbucket/sync_state?connectionId=&scopeId=&table=`

d4x1: @klesh 
Do you mean we change all `latest-sync-state` to `/plugins/{plugin_name}/sync_state` in all plugins ? or just bitbucket ?

klesh: all of course.. we must unified them

abeizn (dismissed):

backend/plugins/jenkins/impl/impl.go
```diff
@@ -189,11 +189,14 @@ func (p Jenkins) ApiResources() map[string]map[string]plugin.ApiResourceHandler
 		// "connections/:connectionId/search-remote-scopes": {
 		// 	"GET": api.SearchRemoteScopes,
 		// },
-		"connections/:connectionId/scopes/*scopeId": {
+		"connections/:connectionId/scopes/:scopeId": {
```
abeizn, line 192: This may also need to be handled like bitbucket

d4x1: Got it.

abeizn (approved):

PR #6649, opened 2023-12-18
fix(zentao): fix panic when checking db connections

d4x1:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
Revert #6633， Fix panic in another way.

Closes N/A

### Screenshots

### Other Information

### DIFF
--- a/backend/core/runner/db.go
+++ b/backend/core/runner/db.go
@@ -167,7 +167,7 @@ func CheckDbConnection(dbUrl string, d time.Duration) errors.Error {
 	go func() {
 		defer func() {
 			if err := recover(); err != nil {
-				result <- errors.Default.New("panic when checking database connections")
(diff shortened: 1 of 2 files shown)


backend/core/runner/db.go
```diff
@@ -165,6 +165,11 @@ func CheckDbConnection(dbUrl string, d time.Duration) errors.Error {
 	result := make(chan errors.Error, 1)
 	done := make(chan struct{}, 1)
 	go func() {
+		defer func() {
+			if err := recover(); err != nil {
+				result <- errors.Default.New("panic when checking database connections")
```
klesh, line 170: please pass down the err or the information will be lost

d4x1: the error will be something like this:
image

It contains some stack info, do you think we should return it to caller?

abeizn (approved):

PR #6685, opened 2023-12-21
feat: forbid starting tx via the dalgorm.Exec

klesh:
### Summary

While the `Begin` method in the `dal.Dal` interface is the recommended way to initiate a transaction, some may mistakenly use `dalgorm.Exec('start transaction')`. This approach, however, carries a high risk of deadlocks due to the underlying connection pool mechanism.

To avoid deadlocks and ensure consistent data operations, always use the Begin method for initializing transactions. This will leverage the connection pool effectively and prevent potential issues.

Note: Diagnosing deadlocks caused by improper transaction initialization can be extremely challenging. This detection is implemented to proactively prevent such occurrences and safeguard data integrity.

### DIFF
--- a/backend/impls/dalgorm/dalgorm.go
+++ b/backend/impls/dalgorm/dalgorm.go
@@ -21,6 +21,7 @@ import (
 	"database/sql"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


abeizn (approved):
lgtm

PR #6686, opened 2023-12-21
Fix zentao and opsgenis API about sensitive fields

### LINKED ISSUE #5643: [Refactor][Auth] Not returning the tokens or passwords on APIs

d4x1:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
1. fix a typo
#### Opsgenie
1. add new test connection API
2. hide sensitive tokens' content in all existing APIs
#### Zentao
1. add custom patcher when updating connections(works with new connection helper)
2. update path connection api, support secret db url field.

Closes #5643 

### Screenshots
#### Opsgenie using new test API
image

### Other Information

### DIFF
--- a/backend/plugins/zentao/models/connection.go
+++ b/backend/plugins/zentao/models/connection.go
@@ -117,6 +117,8 @@ func (connection ZentaoConnection) Merge(existed, modified *ZentaoConnection) er
 			existed.DbUrl = modified.DbUrl
 		} else {
 			// there is no change with db url field.
+			// existedDBUrl = origin(modified.DbUrl)
+			return nil
 		}
 	} else {
 		existed.DbUrl = modified.DbUrl
--- a/backend/plugins/zentao/models/connection.go
+++ b/backend/plugins/zentao/models/connection.go
@@ -107,6 +107,7 @@ func (connection ZentaoConnection) Sanitize() ZentaoConnection {
 	return connection
 }
 
+// Merge works with the new connection helper.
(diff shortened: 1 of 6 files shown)


backend/plugins/zentao/models/connection.go
```diff
@@ -106,6 +107,31 @@ func (connection ZentaoConnection) Sanitize() ZentaoConnection {
 	return connection
 }
 
+// Merge works with the new connection helper.
+func (connection ZentaoConnection) Merge(existed, modified *ZentaoConnection) error {
+	existedDBUrl := existed.DbUrl
```
abeizn, line 112: maybe 
`
existed.DbUrl = modified.DbUrl
return nil
`
is ok?

d4x1: @abeizn 
I am afraid not. If modified db url is something like `username:*****@ip:port/db_name`, it means there is no update with db url field, and we should use the value in database.

abeizn (commented):
lgtm

abeizn (approved):

PR #6701, opened 2023-12-28
feat #6615, replace libgit2 with go-git.

### LINKED ISSUE #6615: [Refactor][gitextractor] replace libgit2/git2go with go-git/go-git
## What and why to refactor
[go-git/go-git](https://github.com/go-git/go-git)
[libgit2/git2go](https://github.com/libgit2/git2go)
1. go-git/go-git is more popular and more active than libgit2/git2go.
2. go-git/go-git  uses [Apache-2.0 license](https://github.com/go-git/go-git#Apache-2.0-1-ov-file) , libgit2/git2go uses  MIT license.
3. DevLake use libgit2/git2go/v33, and it's based on libgit2 version 1.3, it a very old version, the latest version of libgit2 is v1.7.1 https://github.com/libgit2/libgit2/releases. This causes some problems:
     - It's hard to install libgit2 v1.3 manually
     - On my local machine, some softwares need higher verison of libgit2. I cannot keep old version for long time.
4. libgit2/git2go is based on ligbit2,  ligbit2 is a C liberary, while go-git/go-git is writen in pure Golang.

Abrove all, go-git/go-git is the better choice.

How to refactor?
1. Replace the package, and update to new apis.
2. Write some tests to make sure the result are the same.

## Related issues
#4188

d4x1:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
In order to replace libgit2 with go-git, thir PR provides a new config `USE_GO_GIT_IN_GIT_EXTRACTOR` in env.sample, if `USE_GO_GIT_IN_GIT_EXTRACTOR` is set  `1`,  in plugin GitExtractor, DevLake will use go-git to collect repo's data.

Closes #6615 

### Screenshots

### Other Information
In gitextractor, libgit2 will produce a table named `commit_line_change`, while go-git cannot fetch such information.
`commit_line_change` is not used in any dashboards or processes, so go-git just ignore this tables.


keon94:
My thoughts:

- I think `GitRepo` should be turned into an interface with two impls: GoGitRepo and LibGitRepo. Refactor the appropriate methods.
- Instead of writing CSV files for testing, I'd create a test suite where each test case calls the functions on the GitRepo interface for the two implementations mentioned above. Compare the results of the function calls and make sure they are identical.

The above approach should lead to cleaner and more abstract code that is not so tied to a library's implementation. And it'll allow us to seemlessly support both libraries until we decide to remove libgit later.

keon94:
My thoughts:

- I think `GitRepo` should be turned into an interface with two impls: GoGitRepo and LibGitRepo. Refactor the appropriate methods. This creates better abstraction and less coupling in the code.
- Write a temporary main function that uses LibGitRepo to grab data from a repo and perform the extraction functions. Capture the output in CSV files (like you're doing)
- Write a test suite that uses GoGitRepo to run the same functions on the same repo as above. Test the same functions against the output CSV files from above (i.e. compare results).
- Once all is good, get rid of the LibGitRepo impl and the temporary main function.

d4x1:
> My thoughts:
> 
>     * I think `GitRepo` should be turned into an interface with two impls: GoGitRepo and LibGitRepo. Refactor the appropriate methods. This creates better abstraction and less coupling in the code.
> 
>     * Write a temporary main function that uses LibGitRepo to grab data from a repo and perform the extraction functions. Capture the output in CSV files (like you're doing)
> 
>     * Write a test suite that uses GoGitRepo to run the same functions on the same repo as above. Test the same functions against the output CSV files from above (i.e. compare results).
> 
>     * Once all is good, get rid of the LibGitRepo impl and the temporary main function.

Turning `GItRepo` into an interface is a good idea! I'll follow your advice.

backend/plugins/gitextractor/parser/repo_creator.go
```diff
@@ -48,14 +52,25 @@ func (l *GitRepoCreator) LocalRepo(repoPath, repoId string) (*GitRepo, errors.Er
 		l.logger.Error(err, "OpenRepository")
 		return nil, errors.Convert(err)
 	}
-	return l.newGitRepo(repoId, repo), nil
+	var goGitRepo *gogit.Repository
+	if EnableGoGit {
```
keon94, line 56: is the goal of this ticket to add go-git support and keep libgit2 around for a while, or do we want to drop libgit2 support altogether in this one?

d4x1: I want to drop libgit2.

d4x1:
@keon94 @klesh  Please review this PR. I have updated the main part, and won't add new features.

mindlesscloud:
I‘m glad the libgit2 dependency will be eliminated at last. It has caused too much inconvenience to the developers.

d4x1:
> I‘m glad the libgit2 dependency will be eliminated at last. It has caused too much inconvenience to the developers.

Yes, go-git doesn't equal with libgit2(commit_line_change cannot be collected with go-git). Hope go-git will satisfy DevLake's requirements.

klesh:
@mindlesscloud Would you like to take a look at the PR when you find time?😊

env.example
```diff
@@ -62,3 +62,8 @@ ENCRYPTION_SECRET=
 # Set if skip verify and connect with out trusted certificate when use https
 ##########################
 IN_SECURE_SKIP_VERIFY=
+
+##########################
+# In plugin gitextractor, use go-git to collector repo's data
+##########################
+USE_GO_GIT_IN_GIT_EXTRACTOR=1
```
keon94, line 69: wouldn't using true/false be more conventional?

d4x1: No problem, I'll fix it.

backend/plugins/gitextractor/parser/repo_test.go
```diff
@@ -0,0 +1,293 @@
...
+var (
+	enableRepoTest       = false
+	repoId               = "test-repo-id"
+	runInLocal           = true
+	ctx                  = context.Background()
+	subTaskCtx           = &testSubTaskContext{}
+	devlakeRepoRemoteURL = "https://github.com/apache/incubator-devlake"
+	output               = "./output"
+
+	repoMericoLake        = "/Users/houlinwei/Code/go/src/github.com/merico-dev/lake"
```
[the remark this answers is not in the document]
d4x1: Good advice, it has been updated.

d4x1:
@mindlesscloud  If there is no addtional comment, please approve this PR. Thx.

klesh (approved):

PR #6715, opened 2024-01-02
fix: jira data scope not trans

abeizn:
### Summary
fix: jira data scope not trans

Closes na

### Screenshots
image

image

### Other Information


backend/plugins/jira/impl/impl.go
```diff
@@ -208,6 +208,10 @@ func (p Jira) PrepareTaskData(taskCtx plugin.TaskContext, options map[string]int
 		}
 		if err != nil {
 			return nil, errors.Default.Wrap(err, fmt.Sprintf("fail to find board: %d", op.BoardId))
+		} else {
```
klesh, line 211: `else` is not necessary because of the `return`  on the line 210.

klesh (dismissed):

klesh (approved):

PR #6754, opened 2024-01-05
fix: gitlab server extract deployment may fail

klesh:
### Summary
Related fields might not available on Gitlab Server, this PR simply make them `nullable` because they are not used in anywhere at this point.

No need to migrate db since they are already `Nullable` in `mysql` and `postgres` by default

image
image

### Screenshots
self tested OK
image

### Other Information


abeizn (approved):
lgtm

PR #6876, opened 2024-01-26
fix: github run api page<35 limit

abeizn:
### Summary
fix: github run api page<35 limit.

github api bug, only collect page < 35 data, otherwise will get nil response
example: https://api.github.com/repos/apache/incubator-devlake/actions/runs?per_page=30&page=35&status=completed

Closes na

### Screenshots
image

### Other Information


backend/plugins/github/tasks/cicd_run_collector.go
```diff
@@ -80,7 +75,9 @@ func CollectRuns(taskCtx plugin.SubTaskContext) errors.Error {
 				UrlTemplate: "repos/{{ .Params.Name }}/actions/runs",
 				Query: func(reqData *helper.RequestData, createdAfter *time.Time) (url.Values, errors.Error) {
 					query := url.Values{}
-					query.Set("status", "completed")
+					// github api bug, only collect page < 35 data, otherwise will get nil response
```
klesh, line 78: GitHub API returns only the first 34 pages (with a size of 30) when specifying `status=compleleted`, try the following API request to verify the problem.

klesh (approved):

PR #6911, opened 2024-02-02
[fix-6818]: AzDo connection names are now unique

### LINKED ISSUE #6818: [Bug][AzureDevopos] When creating a new connection, allowing the same name to exist does not meet the expectations.
### Search before asking

### What happened

Select the Azure DevOps plugin and create a new connection, naming this connection 'Azure'.

Create another new connection, also naming this connection 'Azure'.

Click 'save connection'. The creation is successful. After creating three times, the effect is as shown in the picture: 
image

The problem with this situation is: Based on the current circumstances, it is not easy to select when associating connections with projects.

Expectation: Similar to the handling with the Jira plugin, when the name is the same as an existing connection name, the interface/api/plugins/azuredevops/connections should give an error message. Below is an example from Jira:
image

### How to reproduce

See above

### Anything else

_No response_

### Version

main

### Code of Conduct

keon94:
Closes #6818 

### Screenshots
image
image

### Other Information

### DIFF
--- a/backend/python/pydevlake/pydevlake/model.py
+++ b/backend/python/pydevlake/pydevlake/model.py
@@ -65,7 +65,7 @@ class ToolTable(SQLModel):
 
 
 class Connection(ToolTable, Model):
-    name: str
+    name: str = Field(unique=True)
     proxy: Optional[AnyUrl]
 
     @validator('proxy', pre=True)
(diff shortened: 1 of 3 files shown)


d4x1 (approved):
LGTM.

PR #7031, opened 2024-02-27
fix: jira test connection error message for wrong user/pass not working

klesh:
### Summary
Thanks @leric  for pointing out the problem at https://github.com/apache/incubator-devlake/pull/7016#discussion_r1502370197

It was wrong to modify the status code on the `api_client` level which deprive the change for caller to obtain the original status code.

However, we still need to avoid returning 401/403 status code since it would break the Basic Auth mechanism when users chose to enable it https://github.com/apache/incubator-devlake/blob/main/devops/releases/lake-v0.20.0/docker-compose.yml#L78

This fix moves the status code wrapping logic to the `api.UnmarshalResponse` function so plugins could have a chance to handle the status code on its own.

Related to: #7601, #6561

### Screenshots
jira wrong endpoint
image
jira wrong user/pass
image
github remote scope list with expired token 
image

### Other Information

### DIFF
--- a/backend/helpers/pluginhelper/api/api_client.go
+++ b/backend/helpers/pluginhelper/api/api_client.go
@@ -91,17 +91,6 @@ func NewApiClientFromConnection(
 		})
 	}
 
-	apiClient.SetAfterFunction(func(res *http.Response) errors.Error {
-		if res.StatusCode >= 400 {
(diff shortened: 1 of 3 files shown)


abeizn (approved):
lgtm

PR #7032, opened 2024-02-27
 fix `updated_at` field in cicd_tasks and cicd_pipelines

d4x1:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
1. Fix `updated_at` field in cicd_tasks and cicd_pipelines,  make sure it has a default value when updating.
2. Remove domain layer tables when collecting data fully.

Closes N/A

### Screenshots
image

### Other Information

### DIFF
--- a/backend/python/plugins/azuredevops/azuredevops/streams/jobs.py
+++ b/backend/python/plugins/azuredevops/azuredevops/streams/jobs.py
@@ -28,6 +28,7 @@ class Jobs(Substream):
     tool_model = Job
     domain_types = [DomainType.CICD]
     parent_stream = Builds
+    domain_tables = [devops.CICDTask]
 
     def collect(self, state, context, parent: Build) -> Iterable[tuple[object, dict]]:
         repo: GitRepository = context.scope
--- a/backend/python/plugins/azuredevops/azuredevops/streams/jobs.py
+++ b/backend/python/plugins/azuredevops/azuredevops/streams/jobs.py
@@ -28,7 +28,7 @@ class Jobs(Substream):
     tool_model = Job
     domain_types = [DomainType.CICD]
     parent_stream = Builds
-    domain_tables = [devops.CICDTask]
+    domain_models = [devops.CICDTask]
 
     def collect(self, state, context, parent: Build) -> Iterable[tuple[object, dict]]:
         repo: GitRepository = context.scope
--- a/backend/python/pydevlake/pydevlake/model.py
+++ b/backend/python/pydevlake/pydevlake/model.py
@@ -128,7 +128,7 @@ class RawDataOrigin(SQLModel):
 
 
 class NoPKModel(RawDataOrigin):
-    created_at:  [datetime] = Field(
+    created_at:  Optional[datetime] = Field(
         sa_column=Column(DateTime(), default=datetime.utcnow)
     )
     updated_at: Optional[datetime] = Field(
--- a/backend/python/pydevlake/pydevlake/model.py
+++ b/backend/python/pydevlake/pydevlake/model.py
@@ -29,7 +29,6 @@ from sqlalchemy.inspection import inspect
 from sqlmodel import SQLModel
 from pydevlake import Field
 
-
 inflect_engine = inflect.engine()
 
 
@@ -41,6 +40,10 @@ class Model(SQLModel):
     updated_at: Optional[datetime] = Field(
         sa_column=Column(DateTime(), default=datetime.utcnow, onupdate=datetime.utcnow)
     )
+
+    def set_updated_at(self):
+        if self.updated_at is None:
+            self.updated_at = datetime.utcnow()
 
 
 class ToolTable(SQLModel):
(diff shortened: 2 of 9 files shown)


d4x1:
I want to change `created_at` field, remove the `Optional` from it's type hint, but it will cause an error like:
```
/***/lib/python3.9/site-packages/sqlmodel/main.py", line 506, in __init__ |
 raise validation_error |
 pydantic.error_wrappers.ValidationError: 1 validation error for CicdScope
```
I don't know how to fix it. @CamilleTeruel  Can you help?

backend/python/pydevlake/pydevlake/subtasks.py
```diff
@@ -206,9 +207,13 @@ def process(self, tool_model: ToolModel, session: Session, ctx: Context):
...
             domain_model.id = tool_model.domain_id()
         session.merge(domain_model)
 
     def delete(self, session, ctx):
-        pass
\ No newline at end of file
+        domain_models = self.stream.domain_tables
+        if domain_models is not None:
+            for domain_model in domain_models:
+                session.execute(sql.delete(domain_model).where(domain_model.raw_data_params == self._params(ctx)))
```
CamilleTeruel, line 219: missing new line

backend/python/pydevlake/pydevlake/model.py
```diff
@@ -125,13 +128,17 @@ def set_tool_origin(self, tool_model: 'ToolModel'):
 
 
 class NoPKModel(RawDataOrigin):
-    created_at: Optional[datetime] = Field(
+    created_at:  Optional[datetime] = Field(
```
CamilleTeruel, line 131: extra space

backend/python/plugins/azuredevops/azuredevops/streams/jobs.py
```diff
@@ -28,6 +28,7 @@ class Jobs(Substream):
     tool_model = Job
     domain_types = [DomainType.CICD]
     parent_stream = Builds
+    domain_tables = [devops.CICDTask]
```
CamilleTeruel, line 31: Rename to `domain_models` in case you keep that.

d4x1: Good advice.

backend/python/pydevlake/pydevlake/model.py
```diff
@@ -125,13 +128,17 @@ def set_tool_origin(self, tool_model: 'ToolModel'):
 
 
 class NoPKModel(RawDataOrigin):
-    created_at: Optional[datetime] = Field(
+    created_at:  [datetime] = Field(
```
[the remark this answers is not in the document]
d4x1: I want to remove `Optional` in `created_at` to solve problems like this:
1. collect and store data in domain layers' tables with `created_at` = some value.
2. collect and store data again, then  new `yield` data with `created_at` = None will make records in database tables `created_at` = `NULL`. (it works during `session.merge()`)

So I cannot remove `Optional` in `created_at`. 

Do you have ideas about how to solve this problem?

backend/python/pydevlake/pydevlake/subtasks.py
d4x1: Updated.

backend/python/pydevlake/pydevlake/model.py
d4x1: Updated.

backend/python/plugins/azuredevops/azuredevops/streams/jobs.py
d4x1: Updated.

backend/python/pydevlake/pydevlake/model.py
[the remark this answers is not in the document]
CamilleTeruel: > collect and store data again, then new yield data with created_at = None will make records in database tables created_at = NULL. (it works during session.merge())

OK, I get it now, the merge overrides the previous value. I didn't understand how it ended up NULL.

First, about `Optional`. With SQLModel, the model gets validated at instanciation. But fields like `created_at` and `updated_at` are set when saving the model to the database. They need to be `Optional`, otherwise we always have to provide them when creating models.
So here, maybe a bit counter-intuitively, `Optional` means that we don't have to provide a value at instanciation but it doesn't necessarily mean that we allow the column value to be `NULL` when saved.

So you can either keep like you did and put `Optional` back or look for an alternative to `session.merge`, like loading saved model if any, write some merge logic, and saving the model back (although it might make collection slower...).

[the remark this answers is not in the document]
d4x1: 1. `created_at`should be optional, argeed. 
2. I haven't find any appropriate method to replace `session.merge()`, fortunately `created_at` is not very neccessary so far.  I think we should just leave it as it's for the present. loading and merging manually is so ungraceful.

CamilleTeruel (approved):

PR #7124, opened 2024-03-06
pr-type/bug-fix 7114 - wire up the page size event to alter page size

### LINKED ISSUE #7114: [Feature][config-ui] Connections pagination isn't wired up to page size drop down
### Search before asking

### What happened

We have hundreds of repositories and I needed to add ac config scope to them all however you can only select 'all' for the currently displayed page of 10. there is a drop down to select larger page sizes but it does not work.

I expected the drop down to reload the page with the selected number of items per page

### How to reproduce

have more than 10 repo connection. select the drop down on the bottom right and select any page size other than 10

### Anything else

this is the git diff patch for the fix:
--- a/config-ui/src/routes/connection/connection.tsx
+++ b/config-ui/src/routes/connection/connection.tsx
@@ -53,7 +53,7 @@ export const Connection = () => {
   const [operating, setOperating] = useState(false);
   const [version, setVersion] = useState(1);
   const [page, setPage] = useState(1);
-  const [pageSize] = useState(10);
+  const [pageSize, setPageSize] = useState(10);
   const [scopeId, setScopeId] = useState<ID>();
   const [scopeIds, setScopeIds] = useState<ID[]>([]);
   const [scopeConfigId, setScopeConfigId] = useState<ID>();
@@ -97,7 +97,7 @@ export const Connection = () => {
   const handleHideDialog = () => {
     setType(undefined);
   };
-
+
   const handleShowDeleteDialog = () => {
     setType('deleteConnection');
   };
@@ -332,6 +332,10 @@ export const Connection = () => {
             pageSize,
             total,
             onChange: setPage,
+            onShowSizeChange: (current,size)=> {
+              setPage(1);
+              setPageSize(size);
+            }
           }}
           rowSelection={{
             selectedRowKeys: scopeIds,

### Version

master

### Code of Conduct

wouldd:
### ⚠️ Pre Checklist

### Summary
What does this PR do?
As per the ticket I raised #7114 this just wires up the existing pageSize value to the relevant event handler from the pagination library so that the page does change the number of items shown when a user selects them from the existing drop down box

Closes #7114

### Screenshots

### Other Information


config-ui/src/routes/connection/connection.tsx
```diff
@@ -332,10 +332,14 @@ export const Connection = () => {
...
             total,
             onChange: setPage,
+            onShowSizeChange: (current,size)=> {
+              setPage(1);
+              setPageSize(size);
+            }
           }}
           rowSelection={{
             selectedRowKeys: scopeIds,
-            onChange: (selectedRowKeys) => setScopeIds(selectedRowKeys as ID[]),
```
mintsweet, line 338: Can you rollback this line? You can try executing yarn lint under config-ui, which will prompt you with an error.

wouldd:
regards your comment about it modifying the layout of a page... are you saying that you did not previously see the page dropdown at all? because for me in chrome I always saw the page dropdown. it just did not do anything, hence I considered it a bug.

mintsweet:
@wouldd sorry, I understand that currently there is no operation to switch the page size, but it does not affect page turning. Why do you think this is a issue?

wouldd:
@mintsweet page turning is fine yes. but you can only associated config scopes with teh repos visibile on teh current page. selecting 'all' only selects those on the current page. so when the page size selector doesn't do anything you cannot have any other page size. So in my example where I linked ~800 repositories which I then needed to associate config scopes to. having to do that 10 at a time then manualy change pages is incredibly slow.
potentially it would ultimately be better to allow for a proper means to select all unmapped repos regardless of hte pages. but as quick and easy improvement. just wiring up the already present page size selector to actually change the page size means you can at least do this operation 100 at a time.

mintsweet:
@wouldd I see. Awesome PR, please wait patiently for another commiter's reply.

wouldd:
@mintsweet is there anyone specific that should be tagged for approving?

mintsweet:
@wouldd Sorry, we are preparing for 1.0, I will help you advance this PR, please be patient.

Startrekzky:
Hi @wouldd . Thanks for your contribution. I just reviewed the change and the PR itself is good. However, since the change will cause the inconsistency of the pagination between different pages, I'm not sure if we should merge it.  

Let me know your thoughts on this. Thank you.

wouldd:
@Startrekzky when you say inconsistency, you mean it will only fix it on one page and not all pages? Do you want me to apply the same fix to other places?. It's worth noting that in the majority of instances of you redenering a table it is for things that you are unlikely to have that many of, so the impact of the broken page size is considerably less than here where you can easily have 1000 entries.

Startrekzky:
@wouldd I agree that some lists will have more entries than the others, and your improvement is good. The question was whether we should merge it now or merge it when all similar pages are updated accordingly.

We can merge it now, but I suggest you update the project list's pagination later. Does it make sense?

Cc. @mintsweet

mintsweet (approved):

PR #7195, opened 2024-03-20
feat: add ready and health probes

klesh:
### Summary
Added the following probing endpoints:
1. `/ready`: returns `200 OK` when system is fully initialized, or `503 Service Unavailable` otherwise
2. `/health`: returns `200 OK` when system is health, or `500 Server Insternal Error` when loading pipelines from the database


ZhangNing10 (approved):
LGTM

PR #7210, opened 2024-03-23
fix: populate missing RepoUrl in CiCDPipelineCommit

### LINKED ISSUE #7193: [Bug][Dora-Cahnge_lead_time_calculator] incorrect linking of commits to deployments when using azuredevops and webhook
### Search before asking

### What happened

I assume that the issue I'm seeing is somehow unique to me using the azuredevops_go plugin and the webhook for reporting deployments from octopus. However the 'bug' I've identified looks to me like it would effect everyone. So probably the real issue is somewhere else but all I can say is that for me the DORA metrics are all wildly out because it relates every commit that has ever been made against a repo to the most recent deploy. This means the cycle time calculation for each commit is showing as just the age of that commit until the most recent deploy.
I believe the offending line is this one: https://github.com/apache/incubator-devlake/blob/cb69f35b5319f57dd3d6c8534121a6ca069cd175/backend/plugins/dora/tasks/change_lead_time_calculator.go#L215C2-L215C3
the commits_diffs table has 3 columns, commit_sha, old_commit_sha and new_commit_sha. For me I need the filter in this query to filter where the mergeSha = cd.new_commit_sha in order to work correctly.
I've tested this in my setup and that change has the expected behaviour. However, I would obviously expect this to be broken for everyone doing DORA metrics at all, and it seems unlikely no one else noticed this. So I guess maybe the intention of that commits_diffs table is to have information in it differently and perhaps the azure plugin is populating it incorrectly?

In my case I have lots of quite old 4+years minimum repos. but I've only just started feeding deploy events in. so I would expect only to have data for the relatively few new commits since this went into place. and I would expect the cycle time calculations for those to be right.

### How to reproduce

configure a long standing repo with lots of historical commtis against it. create a webhook call for the most recent commit/commtis in teh repo and expect to see dora calculations for those commits only. However you should see that it links all of the old commits to the new deployment.
Indeed when I run with a version in which I changed the above filter check then I get the behavior I expect. all the numbers make sense and the pr_cycle_time calculations show up numbers for only commits that have been reported as part of deployments

### Anything else

Happens every time for all commits.

### Version

master

### Code of Conduct

mr-ks:
### Summary
The `enrichPrevSuccessDeploymentCommits` task of the dora plugin relies on `cicd_deployment_commits.repo_url` to fetch  all successful deployments in a given project. Deployments without a specified `repo_url` will not be included in the subsequent calculation steps. Which will lead to problems in the form of #7193.

This PR populates the missing `repo_url`column by joining `_tool_azuredevops_go_builds` and `_tool_azuredevops_go_repos` in the `convertApiBuilds` task of the Azure DevOps Go plugin.

Not sure if this closes #7193 but is definitely related

### DIFF
--- a/backend/plugins/azuredevops_go/e2e/build_test.go
+++ b/backend/plugins/azuredevops_go/e2e/build_test.go
@@ -52,6 +52,7 @@ func TestAzuredevopsBuildDataFlow(t *testing.T) {
 
 	dataflowTester.FlushTabler(&models.AzuredevopsBuild{})
 	dataflowTester.ImportCsvIntoRawTable("./raw_tables/_raw_azuredevops_go_api_builds.csv", "_raw_azuredevops_go_api_builds")
+	dataflowTester.ImportCsvIntoTabler("./raw_tables/_tool_azuredevops_go_repos.csv", &models.AzuredevopsRepo{})
 	dataflowTester.Subtask(tasks.ExtractApiBuildsMeta, taskData)
 
 	// Omit the datetime columns to avoid test failures caused by the varying precision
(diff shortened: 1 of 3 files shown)


klesh:
@mr-ks Hi, this seems to be a legit fix to me, would you like to rebase it so  I can merge it? Thanks.

mr-ks:
@klesh, done. Thanks for the reminder

klesh (approved):

Issue #5633, opened 2023-07-05
[Feature][Jenkins-Data] Jenkins plugin option to add all the multibranch job to be monitored


KrPetrovDev:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

The multibranch job is used for github pull requests which will lead for every pull request created will be created on the fly a new pipeline in the multibranch job which will not be included in the data scopes of the related jenkins connection. 

### Related issues

New pipelines created in the multibranch pipeline will not be added as a data scope in the jenkins connection automatically, they will be required to be inserted manually from the data scope UI of the jenkins connection

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

klesh:
Thanks for the report, however, I'm having trouble understanding the problem since I never used it before, do you have any concrete example of how it works?

KrPetrovDev:
Yes, as an example I have a multi branch job in jenkins, which inside It executes multiple separated pipelines, these pipelines are created on demand as we create pull requests in github but the issue related to devlake is that the pipelines that are generated on demand are not triggered automatically by the jenkins connection in devlake, so every time I have to go to the jenkins connection, add the new pipelines and run again the process of recollecting the data from jenkins.

As all of these pipelines pertains to the same multi branch job I think It would be possible to develop a feature that will 
recollect data from all the existing pipelines, inclusive the new pipelines generated after the previous recollection of metrics from this multi branch job instead of only recollecting data only from the pipelines that were initially selected when the jenkins connection was defined in the devlake project.

I hope that this explanation solves your doubt about how It works.

klesh:
@KrPetrovDev Thanks for the comprehensive explanation, I get the idea of it.
However, to evaluate the possible support, we will need a working jenkins server with proper fake data so we can observe how and what the jenkins API offers, would you like to set up one?

KrPetrovDev:
@klesh I cannot provide a server but I can provide you with a minimum installation guide which will help you to simulate part of my jenkins environment for PRs and to configure a github repository so Jenkins can communicate with It.

This is the guide where explains how to setup an existing jenkins server:
- https://www.youtube.com/watch?v=aDmeeVDrp0o&ab_channel=CloudBeesTV

[...]

PR #7213, opened 2024-03-25
feat: collect and extract jenkins jobs from multibranch pipelines

gustavobini:
### Summary

Collects and extracts job, build, and stages data from multi-branch Jenkins pipelines.

#5633


klesh:
Hi, @gustavobini , Sorry, I have been working on some performance optimization recently, may I come back to you maybe next week?

gustavobini:
> Hi, @gustavobini , Sorry, I have been working on some performance optimization recently, may I come back to you maybe next week?

yes that's perfectly fine!

backend/plugins/jenkins/api/remote_api.go
```diff
@@ -53,16 +53,19 @@ func listJenkinsRemoteScopes(
...
+				Type:     api.RAS_ENTRY_TYPE_SCOPE,
+				Id:       jenkinsJob.ScopeId(),
+				Name:     jenkinsJob.ScopeName(),
+				FullName: jenkinsJob.ScopeFullName(),
+				Data:     jenkinsJob,
 				ParentId: parentId,
 			})
-		} else {
+		case "org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject":
 			// this is a scope
```
klesh, line 69: Is it a `scope` or a `group`?

backend/plugins/jenkins/api/blueprint_v200.go
```diff
@@ -86,8 +86,11 @@ func makeDataSourcePipelinePlanV200(
 			subtaskMetas,
 			scopeConfig.Entities,
 			tasks.JenkinsOptions{
-				ConnectionId: connection.ID,
-				FullName:     jenkinsJob.FullName,
+				ConnectionId:       connection.ID,
+				ConnectionEndpoint: connection.Endpoint,
```
klesh, line 90: Can we store the extra information in the `_tool_jenkins_jobs` table instead of passing it around?

backend/plugins/jenkins/api/remote_api.go
klesh: While reading this part of the code, I thought you were to convert a multibranch to a group and let users choose multiple jobs that belong to it and keep the extractor and collectors as they were, sth feels like a semi-automation.
But it doesn't seem like it because collectors and extractors are changed to work with multiple jobs. 😂

backend/plugins/jenkins/api/blueprint_v200.go
gustavobini: The `class` is used at the tasks as the input between multi-branch and other jobs, so it needs to be available before. The `ConnectionEndpoint` cannot be stored in the `_tool_jenkins_jobs` db, as that will break the plugin in case the connection endpoint changes.

backend/plugins/jenkins/api/remote_api.go
gustavobini: I guess it's a scope in the sense that it needs to be stored in the db, but a group in the sense that job's data is not collected directly from it. 😓 

As of now, multi-branch jobs are groups. Jobs can already be selected from them, but since branches are dynamic, that's the issue. The big issue about multi-branch workflows is there can be as many jobs as there are branches, and that both of those are short-lived. In addition, Jenkins' instances are often configured (as they should be) to delete stale jobs, to keep jobs for a maximum amount of days, to keep a maximum amount of jobs, or all of these.

Finding the jobs dynamically is a must.

klesh: Aah, I see.
Thanks for the explanation.

backend/plugins/jenkins/api/blueprint_v200.go
klesh: I see.
How about that we load the information during the [PrepareTaskData](https://github.com/apache/incubator-devlake/blob/17221aef1b9cb62b2c66e9c19600d28333b7f70f/backend/plugins/jenkins/impl/impl.go#L104) phase?
By doing so, we could rerun the plan even after the connection endpoint changed without the need to update every related blueprints, especially for recurring Advanced Blueprint

gustavobini: @klesh I've moved the properties. I'm not happy with how it is, but I think it's the simplest. All the job names and path validations and assignments being done around make me hesitant to do more. I actually wanna get that refactored at some point if possible.

klesh: Thanks for the update.
Would you like me to merge it now so you may take time for the refactor?

klesh (dismissed):
LGTM

backend/plugins/jenkins/api/blueprint_v200.go
gustavobini: @klesh yes please! and thank you

gustavobini: I can't tell what's wrong with the lint-commit-message error 🤔

klesh:
lint commit message checks if all commit messages of the PR fit https://www.conventionalcommits.org/en/v1.0.0/, it is optional.
Please resolve the conflict and I will merge it ASAP

gustavobini:
> lint commit message checks if all commit messages of the PR fit https://www.conventionalcommits.org/en/v1.0.0/, it is optional. Please resolve the conflict and I will merge it ASAP

@klesh done!

klesh (approved):

Issue #6000, opened 2023-08-31
Jenkins associate input parameter with git ref


saada:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

The current Jenkins connection assumes that the repo clone by the Jenkins job is the same as the business logic repo that we want Dota metrics from.

In this use case, we want to support separated Jenkins repositories that services leverage to build their app. 

Let’s assume we have two git repos: “webapp” and “jenkinsfiles”. If I create a devlake project for the webapp, the GitHub connection in that project points to repo1 (webapp) while the Jenkins connection points to repo2 (jenkinsfiles). When Jenkins runs a new deployment job, it will clone a git SHA from repo2 (jenkinsfiles) not repo1 (webapp). The only link Jenkins knows about back to repo1 is a Jenkins input parameter called “TAG” that gets passed to the job. TAG is a git ref in repo1.  

Note: putting the GitHub and Jenkins connections in same project is not enough today. So a deployment is performed with a "devops-setting commit" in repo 2(jenkinsfiles) and a "business-code commit" in both repo 1(webapp), and we should use the latter commit to measure DORA.

If we could specify in the Jenkins association which input parameter defines the git tag/sha, then we would get correct metrics.

### Description

If we could specify in the Jenkins association which input parameter defines the git tag/sha, then we would get correct metrics

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

d4x1:
@saada  Could you show me how to pass tags to jenkins' job and where the tag parameter will appear?

saada:
Given a jenkinsfile like this

```groovy

pipeline {
    agent any
    parameters {
        string(name: 'PERSON', defaultValue: 'Mr Jenkins', description: 'Who should I say hello to?')
        text(name: 'RELEASE_TAG', defaultValue: '', description: 'Enter git ref for web app repo')
        choice(name: 'CLUSTER', choices: ['qa', 'prof'], description: 'Pick cluster')
    }
    stages {
        stage('deploy to cluster') {
            steps {
                echo "Deploying ${params.RELEASE_TAG} TO ${params.CLUSTER}"

               ...
            }
        }
    }
}
```

[...]

PR #7249, opened 2024-03-29
feat: collect jenkins build parameters

matrixji:
### Summary
Collet jenkins build parameters into raw data.

It may helpful for #6000 if someone continue working based on this PR. 

### Screenshots
N/A

### Other Information
Out current idea is collect build parameter into raw data, then we using dbt or some external scripts to fix the `environment` field in `cicd_deployment_commmits` table

### DIFF
--- a/backend/plugins/jenkins/tasks/build_collector.go
+++ b/backend/plugins/jenkins/tasks/build_collector.go
@@ -68,9 +68,10 @@ func CollectApiBuilds(taskCtx plugin.SubTaskContext) errors.Error {
 				UrlTemplate: fmt.Sprintf("%sjob/%s/api/json", data.Options.JobPath, data.Options.JobName),
 				Query: func(reqData *helper.RequestData, createdAfter *time.Time) (url.Values, errors.Error) {
 					query := url.Values{}
+					actionFields := "lastBuiltRevision[SHA1,branch[name]],remoteUrls,mercurialRevisionNumber,causes[*],parameters[name,value]"
 					treeValue := fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


klesh:
Hi, @matrixji , I see we are collecting more information from the API with this patch.
Seems like you have the whole plan of achieving what you wanted, would you like to share it as a document on the website so others could also learn the idea? Or I am afraid it is too subtle for others to understand how to do so.

matrixji:
I have the same requirement with #6000.
In our use of Jenkins, we use the parameter to specify which env to deploy, e.g. devploy_env="prod" as Production deployment, while deploy_env="test" as the QA deployment.

Currently in the DORA metrics process, while doing data transformation from _ to _, it just uses Jenkins Jobs' name to decide which environment is used. So my plan to support this requirement may be split into below steps:
- Collect the `parameters` into raw data
- Add parameters as string field into `JenkinsBuild`, this may rebuild kv pairs as a string: k1=v1;k2=v2;k3=v3
- In the transformation step, besides deciding if this job is a production env, also check the parameters(maybe matched with {name}~{parameters}):
    - Right now, the job name needs to be set as Deploy_Foo_Service_Prod, we use the pattern `.*_Prod` to match a production deployment.
    - After this is done, we could use the job name Deploy_Foo_Service with the parameter deploy_env=prod, we could set the pattern as `deploy_env=prod` to match a production deployment.


I turned this PR as a draft, I'll try to provide the above 3 commits as a whole PR to close #6000

klesh:
Looking forward to it.

PR #7464, opened 2024-05-13
fix(jenkins): incorrect class type on multibranch jobs

gustavobini:
The input for builds is a workflow job and not a workflow multibranch project.

### Summary
What does this PR do?

This fixes extraction of multibranch builds as it was incorrectly checking for the class type to be of the multibranch project, instead of a workflow job. Only jobs created from multibranch projects have `Input` defined.

I also noted I had only added tests covering the "jobs" aspect of the multibranch flow. This covers the rest.


gustavobini:
@klesh can you check this please? related to #7213

klesh (approved):

klesh:
LGTM, @gustavobini Thanks for your contribution.

Issue #7367, opened 2024-04-24
[Bug][SQL] URL column too short


bok11:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

My pipeline fails to get the latest pipeline run, as it cannot insert the URL into the database.
Here is my URL, with some info replaced with X's but the length is still representative:
https://dev.azure.com/ccccxxxxxxxxxxx/xxxxxx/_apis/build/builds?repositoryId=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx&repositoryType=tfsgit&deletedFilter=excludeDeleted&queryOrder=finishTimeDescending&api-version=7.0&continuationToken=2024-02-06T10%3A35%3A26.7061112Z

As you can see the URL is 264 chars long, while the database schema is 255 chars, making devlake throw an exception:

```
level=error msg=" [pipeline service] [pipeline #90] [task #1212] subtask collectAzuredevopsBuilds ended unexpectedly
  Wraps: 
    (2) remote error response:
      | ERROR: DataError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
      | (MySQLdb.DataError) (1406, \"Data too long for column 'url' at row 1\")
      | [SQL: INSERT INTO _raw_azuredevops_builds (params, data, url, input, created_at) VALUES (%!s(MISSING), %!s(MISSING), %!s(MISSING), %!s(MISSING), %!s(MISSING))]
[...]

PR #7493, opened 2024-05-21
fix: increase max URL length

bok11:
Supports longer azure devops organization names described in #7367

### Summary
Introduce migrations to update the column lengths for azure devops URLS, to allow up to 500 characters.

Closes 7367

### DIFF
--- a/backend/core/models/migrationscripts/20240521_modify_cicd_deployment_commits_repo_url_length.go
+++ b/backend/core/models/migrationscripts/20240521_modify_cicd_deployment_commits_repo_url_length.go
@@ -1,0 +1,48 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package migrationscripts
+
+import (
+	"github.com/apache/incubator-devlake/core/context"
+	"github.com/apache/incubator-devlake/core/errors"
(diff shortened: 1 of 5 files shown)


bok11:
Tested manually with the URL: https://dev.azure.com/bokwill0431xxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

abeizn:
@bok11 Thank you very much for your contribution. I have a little doubt. The error seems to be in the url field in the _raw_azuredevops_builds table, but the processing logic of the code is mainly in the tool layer and domain layer.
image

bok11:
Hi @abeizn  not sure if i am understanding correctly, do you mean why i added the golang side migrations?

In that case, i was debugging with dbeaver initially, and just changed the column lengths manually to identify where it was used.
The URL seems to be saved throughout the system, so in  cicd_pipeline_commits, i end up having the long URL in the repo_url column as well.
I found the same to be true for cicd_pipeline_commits

As i initially did this a month ago, i be happy to try to only change the azuredevops table and see if it works :)

abeizn:
@bok11 No, what I mean is that the error report should be here: https://github.com/apache/incubator-devlake/blob/main/backend/python/pydevlake/pydevlake/model.py#L106. 
So I would change this to: url: str = Field(default=b'', sa_column=Column(Text))
And migrate the url field length in the _raw_azuredevops_builds table to text (provided that the _raw_azuredevops_builds table exists)

bok11:
@abeizn sorry for the delay, life got in the way :)

I did the change to text , but i do have some issue migrating the raw table, as it does not exists before you set up azure devops.
I could not find any other migrations of _raw tables, so i added a more complex SQL query to do the alteration if the table exists.

Let me know if you had something else in mind

abeizn:
@bok11 It is indeed the case that the raw layer table for Azure DevOps only exists if the Azure DevOps plugin (Python) has been run. Therefore, when migrating the script, you need to first check if the _raw_azuredevops_builds table exists. If it does, then proceed with the migration (modify the length of the url field). If it does not exist, it indicates that the user has not run the Azure DevOps plugin (Python), so you only need to modify https://github.com/apache/incubator-devlake/blob/main/backend/python/pydevlake/pydevlake/model.py#L106.
Consider changing it to
`url: str = Field(default=b'', sa_column=Column(Text))`

abeizn (approved):
lgtm

PR #7553, opened 2024-05-31
fix: startedDate and finishedDate should be required

klesh:
### Summary
fix: startedDate and finishedDate should be required according to @louis


Startrekzky (approved):
LGTM

PR #7559, opened 2024-06-02
adding pull_request_id index to  pull_request_commits/comments tables

### LINKED ISSUE #7557: [Refactor][DORA] Refactor CalculateChangeLeadTimeMeta subtask
## What and why to refactor
CalculateChangeLeadTimeMeta subtask is slow when there are a lot of records in the tables that are used to calculate the needed metrics that will eventually end up in the 'project_pr_metrics' table.
Queries that are slow for us are the ones that are called in `getFirstCommit` and `getFirstReview` methods.
They are querying `pull_request_commits` and `pull_request_comments` tables. 
Our fix was to add an index to `pull_request_id` on both tables.

How to refactor?
Introduce index by pull_request_id for `pull_request_commits` and `pull_request_comments` tables.

## Related issues
Please link any other

sstojak1:
### ⚠️ Pre Checklist

- [ ] I have read through the [Contributing Documentation](https://devlake.apache.org/community/).

### Summary
What does this PR do?
Adding pull_request_id index to `pull_request_commits` and `pull_request_comments` tables.

Closes [7557](https://github.com/apache/incubator-devlake/issues/7557)

### Screenshots

### Other Information


backend/core/models/migrationscripts/20240603_add_pull_request_id_index_for_pr_commits_table.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"github.com/apache/incubator-devlake/core/context"
+	"github.com/apache/incubator-devlake/core/errors"
+	"github.com/apache/incubator-devlake/helpers/migrationhelper"
+)
+
+type addPullRequestIdIndexToPullRequestCommits struct{}
+
+type pullRequestCommits20240602 struct {
+	PullRequestId string `gorm:"index"`
```
d4x1, line 12: https://github.com/apache/incubator-devlake/blob/main/backend/core/models/domainlayer/code/pull_request_commit.go#L28,  `PullRequestId` and `CommitSha` are the primary keys, and it works as expected in db tables.

So I think there is no need to add a new `index` in this field.

sstojak1: true in case you are filtering by `CommitSha` or `CommitSha` + `PullRequestId`.
If you are filtering only by `PullRequestId` the composite index won't be used - [docs](https://dev.mysql.com/doc/refman/8.4/en/multiple-column-indexes.html)

image
image

sstojak1: with `PullRequestId`  index:
image

klesh: Ahh.. The order of the primary key columns seems off to me, the `PullRequestId` should be placed before the `CommitSha` column.
However, adding a secondary index is acceptable IMO if adjusting the order is too hard to do so.

sstojak1: This could work I think:
`ALTER TABLE pull_request_commits DROP PRIMARY KEY;
ALTER TABLE pull_request_commits ADD PRIMARY KEY (pull_request_id, commit_sha);`
However, this change might impact the rest of the application, especially if there are custom queries that end users have that expect `commit_sha` to be first in the index sequence.

What do you think?

klesh: @sstojak1 True, we should consider which one is more likely to happen. It is easier to predict someone might need to group records by `pull_request_id` while it is harder to think of how `commit_sha` being prioritized could be beneficial for any case, maybe "Counting how many PRs contain the same commit?" which doesn't make too much sense.

sstojak1: I agree. I changed the pull_request_commits script. Now it's changing the primary key columns' order.
image
image

klesh (commented):
Why are two tables handled differently?
Besides, can we support postgresql as well? Apache DevLake supports pg as a storage option.

sstojak1:
> Why are two tables handled differently? Besides, can we support postgresql as well? Apache DevLake supports pg as a storage option.

Those two tables don't have the same primary key definition. That's the reason why migration scripts are different. 

Is the PostgreSQL supported? There are also a few scripts where this kind of schema check is added...
Also, check this out:
https://devlake-io.slack.com/archives/C03APJ20VM4/p1713164942109869
https://devlake-io.slack.com/archives/C03APJ20VM4/p1716386100745249

sstojak1:
Hi @klesh any updates here?

klesh:
@sstojak1 Hi, sorry for the late reply.
We support using PostgreSQL as a storage option for advanced users who prefer writing their own dashboards without any preset dashboards.
In other words: the `devlake` implementation written in Golang supports both PG and Mysql while the embedded dashboards for grafana only available for Mysql

klesh:
@sstojak1 And yes, we have some scripts designated MySQL, sometimes because they are MySQL specific nuances, which were unnecessary for PG from what  I know.

sstojak1:
@klesh makes sense. I have added support for PostgreSQL. Please take a look.

klesh (dismissed):
LGTM

klesh:
Hi, @sstojak1 we are so close, there is one conflict left before I can merge the PR. Would you like to resolve it? Thanks.

sstojak1:
@klesh done!

sstojak1:
@klesh can we merge it? :)

klesh (approved):

abeizn:
@sstojak1 Already released in [v1.0.1-beta3](https://github.com/apache/incubator-devlake/releases/tag/v1.0.1-beta3)

PR #7657, opened 2024-06-23
feat(azuredevops): support disabled Repos

### LINKED ISSUE #7504: [Bug][Blueprints] 400 error when adding Azure DevOps scope to existing project
### Search before asking

### What happened

On a project with an existing github connection (such as #7503), adding an Azure DevOps connection to it gives a 400 error in the UI:

image

Inside lake microservice container, the error is the below:

```
time="2024-05-24 12:40:03" level=error msg="HTTP 400 error\n\tcaused by: error patching the blueprint (400)\n\tWraps: (2)\n\tWraps: (3) fork/exec ./run.sh: argument list too long\n\tWraps: (4) fork/exec ./run.sh\n\tWraps: (5) argument list too long\n\tError types: (1) *hintdetail.withDetail (2) *hintdetail.withDetail (3) *hintdetail.withDetail (4) *fs.PathError (5) syscall.Errno"
[GIN] 2024/05/24 - 12:40:03 | 400 | 4.889660007s | 10.253.240.52 | PATCH "/blueprints/2"
```

AzureDevops scopes are added to existing project.

### How to reproduce

- Create a project
- Successfully add data scopes from GitHub
- Try to add Azure DevOps scopes

### Anything else

Same environment where #7503 is happening.

### Version

1.0.0-beta6

### Code of Conduct

mr-ks:
### Summary
It is now possible to select disabled repositories as a remote scope. However, disabled repositories are not included in tasks related to domain `code` and `code review`, as the Azure DevOps API returns a 4xx status code for the required API calls.

A new column called `is_disabled` has been added to the `_tool_azuredevops_go_repos` table to support disabled repositories. For existing entries, the default value for this column is set to `false`.
Closes #7504


backend/plugins/azuredevops_go/tasks/pr_commit_collector.go
```diff
@@ -39,7 +39,7 @@ var CollectApiPullRequestCommitsMeta = plugin.SubTaskMeta{
 	EntryPoint:       CollectApiPullRequestCommits,
 	EnabledByDefault: true,
 	Description:      "Collect PullRequestCommits data from Azure DevOps API.",
-	DomainTypes:      []string{plugin.DOMAIN_TYPE_CROSS, plugin.DOMAIN_TYPE_CODE_REVIEW},
```
mr-ks, line 42: Hello @klesh,
what should be the expected outcome if the user creates a scope configuration with only `Cross Domain` as the data entity? Should this also include the tasks of domain `Code Review` and `Source Code Management`?

backend/plugins/azuredevops_go/api/blueprint_v200.go
```diff
@@ -71,52 +69,39 @@ func makeScopeV200(
 	sc := make([]plugin.Scope, 0, 3*len(scopeDetails))
 
 	for _, scope := range scopeDetails {
-		azuredevopsRepo, scopeConfig := scope.Scope, scope.ScopeConfig
-		if azuredevopsRepo.Type != models.RepositoryTypeADO {
+		repo, scopeConfig := scope.Scope, scope.ScopeConfig
+
+		if len(scopeConfig.Entities) == 0 {
```
klesh, line 74: Currently, we are treating this condition as Selected All because it has no point to collect a scope without any entity.

backend/plugins/azuredevops_go/api/blueprint_v200.go
```diff
@@ -126,62 +111,91 @@ func makePipelinePlanV200(
...
 ) (coreModels.PipelinePlan, errors.Error) {
 	plans := make(coreModels.PipelinePlan, 0, 3*len(scopeDetails))
+
 	for _, scope := range scopeDetails {
-		azuredevopsRepo, scopeConfig := scope.Scope, scope.ScopeConfig
-		var stage coreModels.PipelineStage
-		var err errors.Error
+		repo, scopeConfig := scope.Scope, scope.ScopeConfig
+
+		if len(scopeConfig.Entities) == 0 {
```
klesh, line 118: scopeConfig.Entities are considered as "All Entities".

backend/plugins/azuredevops_go/api/blueprint_v200.go
```diff
@@ -126,62 +111,91 @@ func makePipelinePlanV200(
...
+			logger.Printf("Precondition failed. Found empty ScopeConfig for Scope: %v. Skipping",
+				repo.Name)
+			continue
+		}
 
 		options := make(map[string]interface{})
```
klesh, line 124: We have some utility functions to help generate `tasks` and `scopes` , not sure if it could help but you may want to take a look https://github.com/apache/incubator-devlake/blob/5e3a4d7337ded3b27db1931e074a5a46cb1aaca8/backend/plugins/pagerduty/api/blueprint_v200.go

backend/plugins/azuredevops_go/models/migrationscripts/20246008_support_disabled_repos.go
```diff
@@ -0,0 +1,54 @@
...
+		return err
+	}
+
+	return baseRes.GetDal().UpdateColumn(
+		&SupportDisabledRepos{}, "is_disabled", false,
+		dal.Where("is_disabled IS NULL"))
+}
+
+func (*disabledRepos) Version() uint64 {
+	return 20240806100000
```
klesh, line 49: the numbers from the file name and the version are different and both incorrect.

backend/plugins/azuredevops_go/tasks/pr_commit_collector.go
klesh: "Code review" should be included alongside "Cross Domain".
"Cross Domain" means it would produce "accounts" records as a byproduct.
"Source Code Management" should not be included because it represents the `git repo` code base itself.
"Domain Types" are meant to select which subtasks to run but not limit what they could produce. It is simpler this way and most likey more desirable for users.

backend/plugins/azuredevops_go/api/blueprint_v200.go
mr-ks: Thanks for the hint, fixed.

backend/plugins/azuredevops_go/api/blueprint_v200.go
mr-ks: Thanks for the hint, fixed.

backend/plugins/azuredevops_go/api/blueprint_v200.go
mr-ks: Thanks for highlighting this. Unfortunately, we can't use it here as we need to select tasks based on the repository status. For example, a disabled Azure DevOps repo won't allow us to query its pull requests.

backend/plugins/azuredevops_go/models/migrationscripts/20246008_support_disabled_repos.go
mr-ks: Fixed and updated the date to today.

backend/plugins/azuredevops_go/tasks/pr_commit_collector.go
mr-ks: Thanks for clarifying. If we classify both the `collect-,convert-,extractAccounts` tasks and the `pullRequests` tasks as `Cross Domain`, we will face issues selecting the correct tasks for disabled ADO repos (and GH repos). Collecting Accounts will work fine, but collecting pull requests will fail. What are your thoughts on this?

Maybe something like this?

```
entities := scopeConfig.Entities without Cross Domain

tasks := helper.MakePipelinePlanSubtasks(subtaskMetas, entities)
tasks = append(tasks, collect-,convert-,extractAccounts)
```

klesh: @mr-ks Good question, I think we could use the [IgnoreAndContinue](https://github.com/apache/incubator-devlake/blob/6dca288d5e33bb55a8b98995df97ead71b76a082/backend/plugins/jenkins/tasks/stage_collector.go#L116) technique for certain HTTP status codes as a workaround.

mr-ks: @klesh  can you please check again? I've integrated your suggestions.

d4x1 (approved):

klesh (approved):
LGTM, thanks for your contribution

PR #7776, opened 2024-07-23
adjust dockerfile to use numeric uid instead of username

ZhangNing10:
### Summary
adjust dockerfile to use numeric uid instead of username

Closes xx

### Screenshots

### Other Information


Startrekzky (approved):
LGTM

PR #7777, opened 2024-07-24
fix: no error message when hideToast is true

mintsweet:
### Summary
Fixed no error message when hideToast is true.


abeizn (approved):

mintsweet:
Manual cherry-pick completed.

PR #7792, opened 2024-07-30
fix: prevent updates to provisioned dashboards to avoid overwriting user modifications

klesh:
# Summary

Currently, users may edit the provisioned dashboards and save them to the database, this may look nice but the modification from the user will get overwritten if the dashboards have been changed in the newer version of Apache DevLake during an upgrade.

To address the problem, we came to a conclusion: **All provisioned dashboards can be edited but should not be saved, it makes more sense for users to save the edited dashboard as a new one.**

This PR achieves the goal by simply forbidding updating the provisioned dashboards according to https://grafana.com/docs/grafana/latest/administration/provisioning/

# Screenshots

I tested it on my local machine and it seems to be working as expected:

1. When users click on the "Save" button, a window pops up instead of saving it to the database
image
2. Users may go to dashboard settings and click the "Save as" to make a copy
image
image
3. The provisioned dashboard was updated as expected while the copied one stayed intact
image
image

### DIFF
--- a/grafana/provisioning/dashboards/dashboard.yml
+++ b/grafana/provisioning/dashboards/dashboard.yml
@@ -33,7 +33,7 @@ providers:
   # <int> how often Grafana will scan for changed dashboards
   updateIntervalSeconds: 5
   # <bool> allow updating provisioned dashboards from the UI
-  allowUiUpdates: true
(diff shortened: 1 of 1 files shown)


abeizn (approved):
lgtm

PR #7820, opened 2024-08-01
fix(circleci-plugin): pipeline collector time range

### LINKED ISSUE #7797: [Bug][CircleCI Plugin] CircleCI pipelines collected from before time range
### Search before asking

### What happened

When collecting CircleCI pipelines, the time range specified in the sync policy has no effect on the data collected - pipelines are collected from before the specified date.

E.g. Sync policy settings set to collect from 1st June 2024:
Screenshot 2024-07-30 at 16 34 46

Screenshot 2024-07-30 at 16 53 41

Excerpt of `data` JSON blob from top row - has `created_date` and `updated_date` of 1st Feb 2024 (ie. 180 days ago from todays date - 2024-07-30):

```
{
    "id" : "eae60b4c-7dcc-4293-8b00-45f18a494881",
    "updated_at" : "2024-02-01T14:11:31.988Z",
    "created_at" : "2024-02-01T14:11:31.988Z",
    ...
    "trigger" : {
      "received_at" : "2024-02-01T14:11:31.481Z",
      "type" : "webhook",
    ...
    },
    ...
  }
```

No CircleCI pipelines, workflows or jobs are collected from before the time range start point.

### How to reproduce

1. Set the time frame to a date before any CircleCI data retention period ends (e.g. if retention period is 90 days, set this to 30 days - see below).
2. Run the DevLake pipeline
3. Sort the `_raw_circleci_api_pipelines` table by the `created_at` JSON property of the `data` column:
```
SELECT *
FROM _raw_circleci_api_pipelines
ORDER BY STR_TO_DATE(JSON_UNQUOTE(JSON_EXTRACT(CONVERT(data USING utf8mb4), '$.created_at')), '%Y-%m-%dT%H:%i:%s.%fZ') ASC;
```
4. Compare the `created_at` property to that set in the sync policy time range.

### Anything else

This is in part due to the recent pagination fix on the plugin (#7770) - the pagination works but as the CircleCI API does not offer any date range pagination controls, the collector now loops through the pages until `next_page_token` is `null`, which is whenever the data retention limit is hit for the account (e.g. for me it is 180 days, but could be less/more, see [here](https://support.circleci.com/hc/en-us/articles/5645222646939-Data-retention-policy)).

When subsequently attempting to collect the relevant workflows & jobs for the pipeline, this will return a 404 and error the DevLake pipeline for any data points that fall outside of the data retention range in a race condition vs. CircleCI cleaning up build data:

```
subtask collectWorkflows ended unexpectedly Wraps: (2) Retry exceeded 3 times calling /v2/pipeline/6b7c4513-56bd-4e0c-ad72-d562df7513b1/workflow. The last error was: Http DoAsync error calling [method:GET path:/v2/pipeline/6b7c4513-56bd-4e0c-ad72-d562df7513b1/workflow query:map[]]. Response: {:message "Pipeline not found"} (404) Error types: (1) *hintdetail.withDetail (2) *errors.errorString
```

There needs to be an additional check that the `created_at` property of the returned pipelines is not before the specified time range starting point. 

### Version

44c3ecb8194d008dd4e3914bb534d2a26eeefdae

### Code of Conduct

Nickcw6:
### Summary
- Fixes bug whereby CircleCI pipeline collector would continue to collect data and not consider the `timeAfter` config of the DevLake pipeline
- Handle case where circleci workflow or job has been deleted

Closes #7797 

### Screenshots
Before:
Running DevLake pipeline with time range set to 2024-06-01 would pull in results from Feb (ie back 180 days until hitting circleci's data retention policy for my org):
image
image

After:
Only results from 2024-06-01 onwards are considered:
Screenshot 2024-07-31 at 19 26 56

### Other Information
~~May have uncovered another bug in the CircleCI plugin whereby existing data is overwritten by a subsequent collection, see [this comment](https://github.com/apache/incubator-devlake/issues/7797#issuecomment-2262846077)~~ Confirmed as intended behaviour for a full sync plugin


backend/plugins/circleci/tasks/shared.go
```diff
@@ -121,3 +121,12 @@ func ParseCircleciPageTokenResp(res *http.Response) ([]json.RawMessage, errors.E
 	err := api.UnmarshalResponse(res, &data)
 	return data.Items, err
 }
+
+func ignoreDeletedBuilds(res *http.Response) errors.Error {
+	// CircleCI API will return a 404 response for a workflow/job that has been deleted
+	// due to their data retention policy. We should ignore these errors.
+	if res.StatusCode == http.StatusNotFound {
+		return api.ErrIgnoreAndContinue
```
klesh, line 129: Shouldn't it be `api.ErrFinishCollect` ? There would be no more data to be collected if the reason was "data retention",  would it?

Nickcw6: With `api.ErrFinishCollect` it potentially misses workflow/jobs that are still available if the ci data isn't processed exactly in order from newest -> oldest. With `api.ErrIgnoreAndContinue` if that does occur those rows would still be collected.

E.g. say there are 600 pipelines to be collected, 1 has been deleted. These end up out of order and the deleted record gets processed as 300/600. The devlake pipeline would end with 300 rows uncollected.

There shouldn't be a significant number of deleted workflows & jobs for attempted collection (as the associated pipelines will also have been deleted), so the increase in devlake pipeline duration should be minimal & worth the reduced risk of missing rows.

Nickcw6:
Hey @klesh, is there an ETA on when this PR might be reviewed/merged?

backend/plugins/circleci/tasks/shared.go
klesh: Aah,  I see.  It makes total sense.
Sorry for taking so long I was on vacation and somehow missed the notification.

klesh (approved):
LGTM

klesh:
@Nickcw6 Thanks for your contribution.
Will release a new beta on Friday.

PR #8000, opened 2024-09-04
feat: add deployment URL mapping to webhook plugin

ktoublanc:
### Summary
This PR add mapping for deployment URL when creating a deployment through web hook plugin.

I'm planning a second merge request on the documentation repository if accepted

I did not see any relevant tests for the web hook plugin but I am willing to contribute as to this as I'm planning to propose other changes (a fix on the deployment name computation which can exceed 255 chars when multiple commits are present and the ability to delete issues created from the web hook)

No open issues for this change

### DIFF
--- a/backend/plugins/webhook/api/deployments.go
+++ b/backend/plugins/webhook/api/deployments.go
@@ -43,6 +43,7 @@ type WebhookDeploymentReq struct {
 	Result       string `mapstructure:"result"`
 	Environment  string `validate:"omitempty,oneof=PRODUCTION STAGING TESTING DEVELOPMENT"`
 	Name         string `mapstructure:"name"`
+	Url          string `mapstructure:"url"`
 	// DeploymentCommits is used for multiple commits in one deployment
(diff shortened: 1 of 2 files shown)


klesh:
LGTM.
Don't forget to write a migration script to add the column to the database.

ktoublanc:
> LGTM. Don't forget to write a migration script to add the column to the database.

Hey, thanks for the review.

I don't get what migration script I should write...
From my point of view, I'm only adding a field in the Webhook API which will value the Url field of the deployment model.
This field is already created in the database (cicd_deployments.url) which is nullable and I can't set a default value for the field and there are no associated _tool_webhook_... table associated.

Am I getting it right ?

ktoublanc:
@klesh did you have time to check my last comment ?

klesh:
@ktoublanc You're absolutely right. I wasn't aware that the field already existed—thank you for the clarification!

klesh (approved):

klesh:
Merged, thanks for your contribution. 😄

PR #8084, opened 2024-09-20
fix: the start and end times are reversed

mintsweet:
### Summary
Fixed the start and end times are reversed.


Startrekzky (approved):
LGTM

Issue #7655, opened 2024-06-21
[Bug][Devlake][GitHub] GitHub app connection data source usability issues


petercort:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

Creating a GitHub App connection isn't working as expected. 
1. No organizations or repositories are found 'naturally' after the connection is created
2. Can't multiselect to add repositories when they are specified into the datasources search 

![Image](https://github.com/user-attachments/assets/f4a2aa92-38af-48c8-aa4b-b6f34896b864)
![Image](https://github.com/user-attachments/assets/ea3cae2c-6e86-4cc8-9f7e-a914635b58a8)
![Image](https://github.com/user-attachments/assets/6510a14b-8836-4b28-a3f1-804ea0067346)
![Image](https://github.com/user-attachments/assets/a5d3dc25-a083-4ddc-8951-cfe4fe599382)
![Image](https://github.com/user-attachments/assets/207a6333-fa2f-4a7f-bac5-3178ab4e38ae)

### What do you expect to happen

I would expect that the repositories in the organization where the app has been installed would auto populate and I could select multiples to add as a data source.  

### How to reproduce

1. Configure a GitHub App as a connection
2. Add data sources - no 

### Anything else

I was able to reproduce this behavior with a GitHub app installed into an organization and my personal account. 

I'm willing to take a crack at a PR if it's not a massive thing. I'll spend some time setting up a debugger and seeing if I can get more info. Might need someone more skilled than me (cough cough copilot) to fix this. 

I also tested this with the v0.21.0@ac4f7fe and it's reproducible there. 

Here's the error message formatted nicely: 

```
level=error msg="HTTP 400 error
    caused by: attached stack trace
     -- stack trace:
        | github.com/apache/incubator-devlake/helpers/pluginhelper/api.(*DsScopeApiHelper[...]).PutMultiple
        | /app/helpers/pluginhelper/api/ds_scope_api_helper.go:114  
        | github.com/apache/incubator-devlake/plugins/github/api.PutScopes  
        | /app/plugins/github/api/scope_api.go:42
        | github.com/apache/incubator-devlake/server/api.handlePluginCall.func1
        | /app/server/api/router.go:135
        | github.com/gin-gonic/gin.(*Context).Next
        | /go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174
[...]

PR #8088, opened 2024-09-21
fix(github-plugin): removes the processed org check to return all the…

petercort:
… org repos, and pulls the repository data when adding scope

previous behavior would return a single repository. Also no repo data was returned so the user could not add the repo to devlake

fixes part of my issue in #7655

### ⚠️ Pre Checklist

### Summary
Removees the processed org check which was preventing multiple repositories from being pulled. The processed org check would trigger after a single repo was returned.  Also I've added in the repo data model because it wasn't returning data, and therefore data couldn't be added to the  scope!

Partial fix for #7655 

### Screenshots

### Other Information
I'm not sure what the initial purpose the processedOrgs check had, so deleting it entirely might be aggressive. But since the app connection is scoped to a single org I don't see a direct downside to removing it.

### DIFF
--- a/backend/plugins/github/api/remote_api.go
+++ b/backend/plugins/github/api/remote_api.go
@@ -202,18 +202,25 @@ func listGithubAppInstalledRepos(
 	}
 	var appRepos GithubAppRepo
 	errors.Must(api.UnmarshalResponse(resApp, &appRepos))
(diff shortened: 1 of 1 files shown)


abeizn (approved):
lgtm

KyriosGN0:
is this fix available in any version ?

PR #8104, opened 2024-09-25
feat: improve performance of the listing-scope-config-projects-and-scopes

klesh:
### Summary

**Problem:**
The current implementation of the API for listing projects and scopes related to a specific Scope Config may experience timeouts, especially when hundreds of scopes are associated with the same Scope Config.

**Solution:**
This PR addresses the issue by optimizing the database query, reducing it to a single query to improve response time.

### Screenshots

PixPin_2024-09-25_15-33-44


abeizn (approved):
lgtm

PR #8163, opened 2024-10-29
fix: make sql query PostgreSQL compatible

Tonkonozhenko:
### Summary
Make SQL queries work in PostgreSQL

PR #8167, opened 2024-10-30
fix: make sql query PostgreSQL compatible

Tonkonozhenko:
https://github.com/apache/incubator-devlake/pull/8163


klesh:
Nice, thanks for your contribution.

PR #8405, opened 2025-04-24
refactor: refactor StatefulApiExtractor to fix busy buffer and EOF errors

### LINKED ISSUE #7826: [Bug][Jira] issues disappearing from dataset, progressive load issue?
### Search before asking

### What happened

We have noticed in some of our graphs that sometimes we see a complete picture, and other times we're only seeing some fraction of the relevant jira issues.
In both cases this seems to impact jira projects that I know have quite a lot of issues, in once case >10k in the last year or so
When this was first brought up, I triggered a full-refresh , which did not seem to fix it, but then seemingly after another couple of 'normal' refreshes the issues did re-populate the db.
However now a couple of days later they've gone again. it seems mostly that I'm left with small numbers of issues from back at the begining of 2023, which makes me think maybe something is emptying out the previous data, kicking off process to work through re-import but then failing for some unknown reason.

I expected the jira issues to consistently be present, they're still in jira under the board filter etc. So there is no obvious reason that they should disspear from devlake

### How to reproduce

hard to say at this time. I'd suggest a long lived jira project with thousands of issues spread over several months is a good place to start. I guess run repeated refresh cycles and see if the data population in devlake is changing between them in ways it should not.

### Anything else

so far i've seen this happen for 2 specific projects (out of about 25 that we sync. though it's possible that the same issue happens else wehre it's just less obvious in a graph.
not sure if it matters but some of our pipelines are alson syncing quite a few azure devops repos, so the project pipeline iself can take >6 hours to run.
I tried looking in the container logs but did not spot anything that looked like an error.
I'm happy to try running with additional debug or whatever if it would be helpful to assist in understanging this issue better.

Obviously it rather undermines the business faith in the graphs to have them suddenly underreporting by hundreds of missing jira issues.

### Version

v1-custom

### Code of Conduct

caioq:
### Summary
This PR updates the `StatefulApiExtractor.Execute()` method to avoid using a long-lived cursor while writing to the database. Previously, the cursor was kept open during data iteration and insertions, which led to errors like `unexpected EOF` and `busy buffer`. This change introduces a two-phase process: first, we collect all raw data IDs using a cursor and then close it. After that, each row is loaded individually by ID and processed safely. This avoids concurrency issues with the database connection and improves stability for high-volume data extractions.

Closes https://github.com/apache/incubator-devlake/issues/7826

### Screenshots
Testing with Github connection
image

Testing with Jira connection
image

### Other Information
This issue was observed during the `extractIssues` step in the Jira plugin. It occurred consistently on subsequent executions of the task, but not on the first run after restarting the application.
This change ensures a more robust extraction flow without requiring increased `max_allowed_packet` settings or other MySQL tuning.

### DIFF
--- a/backend/helpers/pluginhelper/api/api_extractor_stateful.go
+++ b/backend/helpers/pluginhelper/api/api_extractor_stateful.go
@@ -125,27 +125,11 @@ func (extractor *StatefulApiExtractor[InputType]) Execute() errors.Error {
 	}
 	logger.Info("get data from %s where params=%s and got %d with clauses %+v", table, params, count, clauses)
 
-	// get cursor for IDs only
-	cursor, err := db.Cursor(clauses...)
-	if err != nil {
-		return errors.Default.Wrap(err, "error running DB query for IDs")
-	}
-	defer cursor.Close()
-
-	// collect all IDs
+	// get all IDs
 	var ids []uint64
-	for cursor.Next() {
-		var row struct {
-			ID uint64 `gorm:"column:id"`
-		}
-		err = db.Fetch(cursor, &row)
-		if err != nil {
-			return errors.Default.Wrap(err, "error fetching ID")
-		}
-		ids = append(ids, row.ID)
-	}
-	if err := cursor.Err(); err != nil {
-		return errors.Default.Wrap(err, "error during ID cursor iteration")
+	err = db.Pluck("id", &ids, clauses...)
+	if err != nil {
+		return errors.Default.Wrap(err, "error getting IDs")
 	}
 
 	// batch save divider
--- a/backend/helpers/pluginhelper/api/api_extractor_stateful.go
+++ b/backend/helpers/pluginhelper/api/api_extractor_stateful.go
@@ -102,7 +102,9 @@ func (extractor *StatefulApiExtractor[InputType]) Execute() errors.Error {
 	if !db.HasTable(table) {
 		return nil
 	}
+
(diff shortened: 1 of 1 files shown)


klesh (commented):
Does the error disappear after this patch?
I would recommend to use `db.Pluck` to fetch all ids at once instead of using `db.Cursor`

caioq:
> Does the error disappear after this patch? I would recommend to use `db.Pluck` to fetch all ids at once instead of using `db.Cursor`

Good idea, done https://github.com/apache/incubator-devlake/pull/8405/commits/a32009a2fdae230f3bdb50f3fdcc3e96f4f509ec

Yes, I tested it locally and the error disappeared.
I tested it with 8K Jira Issues and the following database configuration:
```yaml
mysql:
    image: mysql:8
    volumes:
      - mysql-storage:/var/lib/mysql
    restart: always
    ports:
      - 3306:3306
    environment:
      MYSQL_ROOT_PASSWORD: admin
      MYSQL_DATABASE: lake
      MYSQL_USER: merico
      MYSQL_PASSWORD: merico
      TZ: UTC
    command: --character-set-server=utf8mb4
      --collation-server=utf8mb4_bin
      --skip-log-bin
      --max_allowed_packet=9M
      --net_read_timeout=6
      --net_write_timeout=6
 ```

klesh (approved):

klesh:
Nice work. Thanks for your contribution.

PR #8522, opened 2025-07-31
fix: allow 0 idle database connection

carocad:
### Summary

Allow having an idle connection value of 0 or less as this is supported by the SQL driver documented [here](https://gorm.io/docs/connecting_to_the_database.html#Connection-Pool) which points [here](https://pkg.go.dev/database/sql#DB.SetMaxIdleConns).

No

### Other Information

In my setup we are using AWS Aurora Serverless v2. One important property of this technology is that it automatically scales the cluster depending on load. It can scale down to 0 and then we don't pay for the use of the database; which we want in order to save money. 

Unfortunately devlake doesn't allow having less than 1 idle connection. This PR fixes that by removing the condition from the code and relying on the SQL driver to manage the connection pooling.


carocad:
@klesh sorry for the ping. Any chance of this getting reviewed 🙏 ?

klesh:
It makes sense to me.
However, the fix changes the default value, which is a breaking change.
Not sure how it would impact the data collection / processing performance?
Or, should we move the default value to the `.env.example` and `helm-chart` while allowing people to change it as well?

carocad:
> It makes sense to me. However, the fix changes the default value, which is a breaking change. 

fair point, an alternative implementation would be to only default to 10 connections if the user didn't provide a value. The method `GetInt` returns `0` if there was nothing provided. According to the SQL driver [docs](https://pkg.go.dev/database/sql#DB.SetMaxIdleConns) `If n <= 0, no idle connections are retained`. So a user could specify `-1` as a value and remove idle connections that way. 

The current behavior would stay the same for users already providing a value `> 0`.

```go
idleConns := configReader.GetInt("DB_IDLE_CONNS")
if idleConns == 0 {
    idleConns = 10
}
```

klesh:
Nice, so. we don't need to change the code, right?
Maybe update https://devlake.apache.org/  ?

carocad:
> Nice, so. we don't need to change the code, right? Maybe update https://devlake.apache.org/ ?

small misunderstanding. I updated the PR to make the change clear. This would keep the default value for anyone that isn't providing one but would still allow 0 idle connection if the user explicitly wants to by using a value `< 0`

klesh:
> > Nice, so. we don't need to change the code, right? Maybe update https://devlake.apache.org/ ?
> 
> small misunderstanding. I updated the PR to make the change clear. This would keep the default value for anyone that isn't providing one but would still allow 0 idle connection if the user explicitly wants to by using a value `< 0`

I see. Thanks for the information. It makes sense to me.

klesh (approved):
LGTM

PR #8536, opened 2025-08-13
fix(github-graphql): fix empty startedAt field

### LINKED ISSUE #8490: [Bug][GitHub] Error extracting GitHub data - incorrect datetime value
### Search before asking

### What happened

During the data collection, I ran into this error when pulling from GitHub:

> subtask Extract Jobs ended unexpectedly Wraps: (2) error adding result to batch (500) Wraps: (3) Error 1292 (22007): Incorrect datetime value: '0000-00-00' for column 'started_at' at row 81 (500) Wraps: (4) Error 1292 (22007): Incorrect datetime value: '0000-00-00' for column 'started_at' at row 81 Error types: (1) *hintdetail.withDetail (2) *hintdetail.withDetail (3) *hintdetail.withDetail (4) *mysql.MySQLError

I was running version 1.0.0 and upgraded to 1.0.2, but the error is still happening. 

I was expecting the job to be completed without any failures. 

### How to reproduce

No repro steps.

### Anything else

_No response_

### Version

1.0.2

### Code of Conduct

petkostas:
### Summary
GitHub GraphQL API can return check runs with nil or a zero `StartedAt` (manual testing over GitHub API returns a zero `time.Time` value -0001-01-01-) time for jobs that were skipped immediately. These jobs have a CompletedAt time (when they were skipped) but no actual start time resulting to the extractor crashing.

Closes #8490 

### Screenshots
TODO: Add the screenshot of the output

### Other Information
We should decide if we should skip those jobs or if we should use the completedAt as the startedAt time.

### DIFF
--- a/backend/plugins/github_graphql/tasks/job_collector.go
+++ b/backend/plugins/github_graphql/tasks/job_collector.go
@@ -19,6 +19,7 @@ package tasks
 
 import (
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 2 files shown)


klesh (approved):

klesh:
Good work, the comment is clear and easy to understand.

PR #8547, opened 2025-08-21
fix(jira): update epic collector to use new API endpoint and include

### LINKED ISSUE #8544: [Bug][Jira] Integration fails due to deprecated /api/2/search endpoint (HTTP 410)
### Search before asking

### What happened

The DevLake Jira data collection is failing because it's using the deprecated `/api/2/search` API endpoint that Atlassian has removed.

**Error Details:**
- HTTP Status: 410 Gone
- Error Message: "The requested API has been removed. Please use the newer, enhanced search-based API instead"
- Affected Operation: `subtask collectEpics` failing after 3 retries

**Sample Error Log:**
`subtask collectEpics ended unexpectedly Wraps: (2) | Retry exceeded 3 times calling api/2/search. The last error was: Http DoAsync error calling [method:GET path:api/2/search query:map]. Response: {"errorMessages":["The requested API has been removed. Please use the newer, enhanced search-based API instead. Deprecation details are available at [https://developer.atlassian.com/changelog/#CHANGE-2046."],"errors":{](https://developer.atlassian.com/changelog/#CHANGE-2046.%22%5D,%22errors%22:%7B)}} (410)`

Jira data collection should complete successfully and retrieve issue data without API errors. The system should use the current supported Jira REST API endpoints for searching issues.

### How to reproduce

1. Configure DevLake with a Jira data source
2. Run data collection that includes epic collection (`collectEpics` subtask)
3. Observe the collection fails with HTTP 410 errors
4. Check logs for `/api/2/search` endpoint calls
5. Error occurs consistently on any Jira instance due to deprecated API usage

**Environment where reproduced:**
- Any Jira Cloud
- DevLake attempting to collect Jira data
- Issue occurs during epic collection phase

### Anything else

**Root Cause:** DevLake is calling the deprecated `/api/2/search` endpoint instead of the current Jira REST API search endpoint.

**Impact:** 
- High severity - Complete failure of Jira data collection
- Affects all Jira integrations using this endpoint

**References:**
- Atlassian Deprecation Notice: https://developer.atlassian.com/changelog/#CHANGE-2046
- Current Jira REST API documentation should be consulted for migration path

**Suggested Fix:**
Update the Jira plugin to use the current search API endpoint and ensure JQL queries and authentication methods are compatible.

### Version

v1.0.3-beta1@2ef56cf

### Code of Conduct

GoSimplicity:
Fixed the Epic collector to use the new Jira API endpoint and ensured all required fields are included when fetching Epic data. This resolves the synchronization failures caused by the deprecated endpoint (fix #8544 ).


klesh (commented):
hanks for your contribution!
This should resolve the JIRA Cloud data collection issue.
However, I’m concerned it may not work with older JIRA Server versions.
I think we should detect the JIRA version and call the appropriate APIs accordingly.

For reference, here’s some related code:
https://github.com/apache/incubator-devlake/blob/ac3b1d9cb317e6a08b1412ddef0de7bbb9b97c9e/backend/plugins/jira/tasks/epic_collector.go#L54

GoSimplicity:
> hanks for your contribution! This should resolve the JIRA Cloud data collection issue. However, I’m concerned it may not work with older JIRA Server versions. I think we should detect the JIRA version and call the appropriate APIs accordingly.
> 
> For reference, here’s some related code:
> 
> https://github.com/apache/incubator-devlake/blob/ac3b1d9cb317e6a08b1412ddef0de7bbb9b97c9e/backend/plugins/jira/tasks/epic_collector.go#L54

Based on your suggestion, I have made the new and old version endpoints compatible.

klesh (commented):
That was fast — great job!

However, I made a mistake in pointing to an inappropriate line of code for this case.  
The version check on line 54 was originally added because the performance of Jira versions ≤ 8 was quite poor, so we reduced the batch size to avoid overloading or crashing the Jira server.

In the current scenario, checking only the server type should be sufficient.  
I'm not sure whether Jira Server 9 supports the `api/v3` endpoint or not.

backend/plugins/jira/tasks/epic_collector.go
```diff
@@ -98,6 +117,10 @@ func CollectEpics(taskCtx plugin.SubTaskContext) errors.Error {
 			query.Set("startAt", fmt.Sprintf("%v", reqData.Pager.Skip))
```
narrowizard, line 117: Based on the information provided in the Atlassian developer changelog, the new `api/3/search/jql` endpoint has changed the way data is paginated. Specifically, random page access has been replaced with a continuation token API. This means that you won't be able to fetch multiple pages simultaneously using parallel threads. The `startAt` parameter has been replaced with `nextPageToken`.

Given these changes, simply updating the API endpoint without adjusting the pagination logic will result in incomplete data retrieval. The existing code that relies on the `startAt` parameter for pagination will no longer work as expected. It's crucial to implement the new nextPageToken mechanism to handle pagination properly and ensure that all data can be fetched correctly.

> We’ll replace random page access with a continuation token API. This means you won’t be able to get multiple pages at the same time with parallel threads. startAt parameter will be replaced with nextPageToken.

GoSimplicity:
> That was fast — great job!
> 
> However, I made a mistake in pointing to an inappropriate line of code for this case. The version check on line 54 was originally added because the performance of Jira versions ≤ 8 was quite poor, so we reduced the batch size to avoid overloading or crashing the Jira server.
> 
> In the current scenario, checking only the server type should be sufficient. I'm not sure whether Jira Server 9 supports the `api/v3` endpoint or not.

 You're right about the performance considerations for JIRA Server ≤ 8. I'll keep the reduced batch size for those versions. Regarding JIRA Server 9's support for api/v3, I'll need to investigate further and potentially add version-specific handling.

backend/plugins/jira/tasks/epic_collector.go
klesh: There’s already a mechanism for handling this kind of API. Here’s how you can enable it:

1. Specify `GetNextPageCustomData` in `ApiCollectorArgs` to extract the `nextPageToken`.  
   - This token will be stored in the `RequestData.CustomData` field.  
   - Note: this also forces the collector to run in sequential mode (no parallel fetching).  

2. In the `Query` function, read the `CustomData` and plug it into the query string.  

Here’s an example:  
https://github.com/apache/incubator-devlake/blob/52433a8bc098eba668a02d815d618409aa375b93/backend/plugins/bitbucket_server/tasks/pr_collector.go#L47-L48

klesh:
I see that there are new commits, and the code looks good to me. Just to confirm, is it ready for merging? Have you tested it against Jira Cloud, and does the pagination work as expected?

GoSimplicity:
> I see that there are new commits, and the code looks good to me. Just to confirm, is it ready for merging? Have you tested it against Jira Cloud, and does the pagination work as expected?

Yes, the code is ready for merging. I have done some testing and it works properly.

klesh (approved):
LGTM
Thanks for your contribution

Issue #8577, opened 2025-09-16
[Feature][PR Size] Exclude specified file types from PR Size (GitLab)


luoyutong47:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

In our workflow, we rely on DevLake’s PR Size metric to understand how much code is being changed by pull requests in a given time range. However, sometimes PRs include large modifications to non-code files (e.g. .md, .txt, .json), which inflates the PR Size metric and makes it harder to evaluate the actual code changes.

Our goal is to get a more accurate view of PR Size by excluding file types that are not relevant to code complexity or review workload.
(We are using GitLab as the data source.)

### Description

It would be helpful if DevLake could support ignoring user-specified file types when calculating PR Size:
- Users can configure a list of file extensions to exclude (e.g. [".md", ".txt"]).
- Lines of code changed in these files will not be counted toward PR Size.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

GoSimplicity:
I will try to complete the development of this new feature, and it may take some time. Thank you for your understanding and support.

GoSimplicity:
I have submitted a new PR. You can check it to see if it achieves the expected results

luoyutong47:
@GoSimplicity Thanks for the quick follow-up! I‘ve checked the PR and it looks exactly what we needed. Thanks for your effort!

PR #8586, opened 2025-09-23
feat(gitextractor): add support for excluding file extensions in commit stats

GoSimplicity:
add support for excluding file extensions in commit stats #8577


klesh (approved):
LGTM.
Very impressive work. Thanks for your contribution.

Issue #8579, opened 2025-09-17
[Feature][Jenkins Connection] Introduce filtering of branch builds


OG256:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

As a user of Apache Devlake I want the ability to only show deployment pipelines from Jenkins that I actually care about and that are relevant to me and my organisation.

### Description

The current Jenkins collector appears to fetch all pipelines - even if they are temporary branch/PR builds. Devlake should have the option to filter the collection results (or only collect certain results based on a regex). As an example there is a Jenkins collector configured in this project where only the following data scopes are selected:

<img width="782" height="594" alt="Image" src="https://github.com/user-attachments/assets/3ac3304c-01b5-460f-99a1-1b8db4df8d37" />

However when viewing/searching for the project in Grafana it shows on the available dropdown of a bunch of random PR branches that are not selected.

<img width="706" height="355" alt="Image" src="https://github.com/user-attachments/assets/2d1b01e0-f79d-4f7e-a9ab-a0ee65611f2b" />

I assume this has an additional underlying impact about redundant pipelines/scopes

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

GoSimplicity:
I have implemented this function. I wonder if it meets your expectations

klesh:
Hi, @OG256  Can you take a look at the PR and provide your insight? Thanks.

PR #8601, opened 2025-10-02
feat(jenkins): add branch filtering for Jenkins jobs

GoSimplicity:
add branch filtering for Jenkins jobs #8579


klesh (commented):
After reading issue #8579, could the problem be that `scope_api.go` and `remote_api.go` are returning incorrect scopes?

GoSimplicity:
> After reading issue #8579, could the problem be that `scope_api.go` and `remote_api.go` are returning incorrect scopes?

Thanks for the review I checked both `remote_api.go` and `scope_api.go`. You're right to question this.
From what I can see:
- `remote_api.go` looks fine - MultiBranchProject is treated as a scope (it's in `scopesWithJobs`), so its branch jobs won't show up as selectable scopes in the UI
- The issue in #8579 is different - when you select a MultiBranchProject, the collector grabs all branch jobs underneath it, converts them to `cicd_pipelines`, and they all show up in Grafana
My PR filters branches during collection so only matching ones get into Grafana. But you're right that we should check if `GetScopeList` is also returning these auto-collected branch jobs alongside user-selected scopes.
Should I add logic to distinguish user-selected scopes from auto-collected branch jobs? Or is filtering at collection time enough for this use case?

klesh:
Thanks for the explanation. 
@OG256 could you provide some context/input on this?

Issue #5207, opened 2023-05-16
[Feature] ArgoCD Integration


1ddo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

Most of our CD are done with Kubernetes ArgoCD and we wanted to add this data connection to DevLake so we could leverage DORA for our deployment frequency.

### Description

Add Kubernetes ArgoCD as new data connection to DevLake

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

mintsweet:
hi @1ddo, thank you for your report.

We have received your information and will consider your needs.

klesh:
@1ddo Hi, I see you checked the **Yes I am willing to submit a PR!**, would you like to work on it? If so, we could assign the issue to you and you can start working on it.

1ddo:
Hi @klesh , I'm currently working on it locally, I'm new to this stuff and I'm having difficulty in making devlake work locally in windows. Was able to run the components in Docker Windows however I'm getting API Down in the config UI (localhost:4000), the tutorial said just wait for a few minutes to let the API up and running but mines seems not working. Do you have any more docs on setting devlake locally using windows or any docs that could help me speed things up?

klesh:
@1ddo Hi, I'm using Windows too because of Feishu(Good collaborative software but lousy Linux support). Sadly, development in Native Windows is not supported, I am using WSL2, which is practically an Ubuntu Linux environment in conjunction with VSCode Remote WSL plugin, you may take a look, it is a far superior development experience than Native Windows.

1ddo:
Hi @klesh, thanks for the advice. Will try to set up a WSL2

abeizn:
It should be that your version is not the latest, you can use [v0.17.0-beta5 ](https://github.com/apache/incubator-devlake/releases/tag/v0.17.0-beta5) version, the installation method can refer to here: https://devlake.apache.org/docs/next/GettingStarted/DockerComposeSetup

klesh:
> Hi @klesh , do you know the reason why I don't have a Project menu in my config-ui:4000? I only see Data Connections, Blue Print, Disabled button for Connection and Dashboard menu options?

[...]

PR #8610, opened 2025-10-13
feat(plugins): initial implementation of argocd

rbstp:
### Summary
This commit implements a new ArgoCD plugin that enables deployment frequency tracking and DORA metrics calculation by collecting sync operation data from ArgoCD applications. It covers rollouts, deployments, replicaset, etc and are sorted per project.

Included is also a grafana dashboard to show simple statistics for the applications.

IMPORTANT: The applications in ArgoCD must be synced at least once for the plugin to collect data. Importing them via an application set does not count as a sync operation.

https://github.com/apache/incubator-devlake/issues/5207

### Screenshots
image
image
image
image

### Other Information
Since I used the .devcontainers, I also updated the version or else mockery would not work. The sorting of the plugins was also updated to keep them in order.


klesh (approved):
LGTM

Issue #8469, opened 2025-06-16
[Bug][GitHub GraphQL] Fetching Pull Requests fails with HTTP 502


robaca:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

In one of our projects, the GitHub GraphQL PR Collector fails reproducible due to HTTP 502. I already found this old bug #3708 and tried to further reduce the page size, but without success. It always seems to happen at the same collector page (depending on page size).

### What do you expect to happen

I would expect that the Job doesn't fail or at least gives more valuable information in the logs about the query which failed. I only see the response body (standard nginx 502 gateway error).

### How to reproduce

This seems to related to actual data of one private repository, so I can't share a way to reproduce it. I'm not familiar to Go programming, but can do a local docker backend rebuild and would appreciate getting help to further boil this down. 

### Version

v1.0.2-rc1

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Startrekzky:
Noted. Please take a look @klesh

klesh:
5xx errors indicate a server-side issue—in this case, with the GitHub API—where the server encounters a problem it cannot properly handle. These errors can be triggered by various factors such as page size or other request-specific conditions, so it's difficult to say definitively whether you're encountering the same issue as described in #3708.

Ideally, GitHub should address the underlying cause of the 5xx error and return a more appropriate response, such as a 4xx error indicating the request size is too large or something along those lines. I recommend reporting this issue to GitHub so they can investigate further.

That said, if you’d like to dig deeper into this on your end, feel free to reach out—I’m happy to help however I can.

robaca:
Not sure what changed, but this error didn't occur any more after a while.

robaca:
With a second clean installation with the same project setup, the error is occurring again, so I reopen the ticket. 

It would be helpful if I can get DevLake to at least skip the PR and continue with the job.

robaca:
I get the following stacktrace:
```
attached stack trace -- stack trace: 

| github.com/apache/incubator-devlake/server/services.RunTasksStandalone 
| /app/server/services/task.go:217 
| github.com/apache/incubator-devlake/server/services.(*pipelineRunner).runPipelineStandalone.func1 
[...]

Issue #8614, opened 2025-10-16
[Bug][github_graphql] Collector Runs Indefinitely on Large Repos likely due to `stream error: CANCEL`


danielemoraschi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

This is a follow-up to the discussion in #8576 

When running a pipeline on a project that includes a very large GitHub repository (5+ years of history, 100' of commits weekly), the `github_graphql` task for collecting job runs hangs for days. The logs show that the process eventually fails due _likely_ to a `stream error: CANCEL` received from the peer (GitHub's servers).

As identified by @klesh in the original thread, this is likely caused by the API response body size exceeding a limit on GitHub's end, which leads to the server terminating the connection.

Errors in the logs:
```
time="2025-10-15 13:01:52" level=warning msg=" [pipeline service] [pipeline #63] [task #16353] retry #1 graphql calling after 120s\n\tcaused by: stream error: stream ID 1; CANCEL; received from peer"
time="2025-10-15 13:04:03" level=warning msg=" [pipeline service] [pipeline #63] [task #16353] retry #2 graphql calling after 120s\n\tcaused by: non-200 OK status code: 502 Bad Gateway body: \"<html>\\r\\n<head><title>502 Bad Gateway</title></head>\\r\\n<body>\\r\\n<center><h1>502 Bad Gateway</h1></center>\\r\\n<hr><center>nginx</center>\\r\\n</body>\\r\\n</html>\\r\\n\""
```

### What do you expect to happen

The pipeline should handle large API responses either completing the data collection successfully or fail with a error message about exceeding API limits, rather than hanging indefinitely.

### How to reproduce

- Configure a DevLake project with a connection to a very large GitHub repository.
- Create and run a new pipeline that includes collecting GitHub Actions data.
- Monitor the devlake container logs.
- Observe that the pipeline hangs on the `github_graphql` task, specifically "Collect Job Runs".
- After a long period, the following error appears in the logs

### Anything else

Attaching here a new snapshot log file.
[task-16353-2-1-github_graphql.log](https://github.com/user-attachments/files/22946344/task-16353-2-1-github_graphql.log)

### Version

v1.0.3-beta6@44f2db2

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

[...]

PR #8616, opened 2025-10-17
Fix Pagination Bug in Github GraphQL Job Collector and add configurable page and batch size

### LINKED ISSUE #8615: [Bug][Github_graphql] Job_collector not paginating correctly
### Search before asking

### What happened

We have repositories with github actions pipelines that consistently have around 30+ Jobs to a run and we noticed we were missing many deployments (which was usually the last job of a run) to be reported in the DORA metrics.

All deployments should be identified correctly, even for github action workflow runs with more than 20 jobs.

### How to reproduce

Have multiple github actions workflow runs with more than 20 jobs representing a deployment

### Anything else

By peeking at the raw tables in the database we could determine, that these runs that were not detected correctly only ever had exactly 20 jobs in the database instead of the full 30+. So something seems wrong with the pagination of the graphql query.

### Version

v1.0.3-beta6

### Code of Conduct

FlomoN:
### ⚠️ Pre Checklist

### Summary

This PR fixes the Pagination bug from #8615 and adds the ability to adjust the collection paramaters of the Github GraphQL Job collector through environment variables. 

The bug was caused by a pagination implementation for workflow jobs in `job_collector.go` that didn't play well together with the simultaneous batching of multiple workflow runs. In `getPageInfo` the first workflow run that had `HasNextPage` true returned it's `EndCursor` which was then used for pagination for all Workflow runs in the batch, but only worked for the one it came from, therefore missing all the pages of other workflow runs in the same batch.

```go
func getPageInfo(query interface{}, args *helper.GraphqlCollectorArgs) (*helper.GraphqlQueryPageInfo, error) {
	queryWrapper := query.(*GraphqlQueryCheckRunWrapper)
	hasNextPage := false
	endCursor := ""
	for _, node := range queryWrapper.Node {
		if node.CheckSuite.CheckRuns.PageInfo.HasNextPage {
			hasNextPage = true
			endCursor = node.CheckSuite.CheckRuns.PageInfo.EndCursor // <- This cursor will be used for all workflow runs, since only one skipCursor variable exists.
			break // <- Then stops after the first was found
		}
	}
	return &helper.GraphqlQueryPageInfo{
		EndCursor:   endCursor,
		HasNextPage: hasNextPage,
	}, nil
}
```

Since some people reported having timeout issues with this plugin (@ClaudioMascaro and @robaca) for large repositories with large workflows I didn't want to take away either option. But combining pagination and batching at the same time didn't seem feasible without adding a lot of hard to maintain complexity to it (like keeping track of all produced EndCursors and submitting them to following collection, but only for those workflow runs that had more pages...).

I ended up implementing a configurable mode switch:

- Either you choose **BATCHING** mode, if you want a little bit better performance while collecting to reduce graphQL API calls, when your Workflow Runs dont have more than 20 Jobs (this number is also adjustable however).
- OR you choose **PAGINATING** mode, if your workflow runs typically have more than 20 Jobs and would potentially cause timeouts of the graphql API when trying to fetch all jobs at once.

I think Rate Limit wise it doesnt make a difference (If I understood that point system correctly), since both ways would in total consume about the same amount of points.

Since it was suggested in #8469 to have the batch size and page size configurable via environment variable, I also added this to give more flexibility to the users in finding the sweet spot for their setup.

Closes #8615

### Screenshots

### Other Information

Maybe it would also make sense to add these configuration options to Scope Config and thus make them configurable on a per scope config basis, but this would also need migration scripts and so on. So maybe an improvement for the future?

I will add tests aswell in the following days, if this approach seems appropriate to you?

### DIFF
--- a/backend/plugins/github_graphql/tasks/job_collector.go
+++ b/backend/plugins/github_graphql/tasks/job_collector.go
@@ -43,15 +43,75 @@ const (
 // Set the collection mode here
 // BATCHING: Query multiple runs at once, no pagination (may miss jobs if >20 per run)
 // PAGINATING: Query one run at a time with full pagination (complete data, more API calls)
-const JOB_COLLECTION_MODE = JOB_COLLECTION_MODE_PAGINATING
+const DEFAULT_JOB_COLLECTION_MODE = JOB_COLLECTION_MODE_BATCHING
 
 // Mode-specific configuration
 const (
-	BATCHING_INPUT_STEP   = 10 // Number of runs per request in BATCHING mode
-	BATCHING_PAGE_SIZE    = 20 // Jobs per run in BATCHING mode (no pagination)
-	PAGINATING_INPUT_STEP = 1  // Number of runs per request in PAGINATING mode
-	PAGINATING_PAGE_SIZE  = 50 // Jobs per page in PAGINATING mode (with pagination)
+	DEFAULT_BATCHING_INPUT_STEP  = 10 // Number of runs per request in BATCHING mode (must be > 1)
+	DEFAULT_BATCHING_PAGE_SIZE   = 20 // Jobs per run in BATCHING mode (no pagination)
+	PAGINATING_INPUT_STEP        = 1  // Number of runs per request in PAGINATING mode (always 1)
+	DEFAULT_PAGINATING_PAGE_SIZE = 50 // Jobs per page in PAGINATING mode (with pagination)
 )
+
+// JobCollectionConfig holds the configuration for job collection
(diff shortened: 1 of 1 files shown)


FlomoN:
@klesh here it is :)

klesh (approved):
Neat.
Could you kindly update the docs as well?
It is hosted at https://devlake.apache.org, the repository is https://github.com/apache/incubator-devlake-website

klesh:
The configurable `page_size` should help #8614 as well.

Issue #8630, opened 2025-10-31
[Feature][ArgoCD Plugin] Parse ArgoCD metadata for docker image


ppapp92:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar feature requirement.

### Use case

As DevOps Engineer I want to be able to write a blueprint that can correlate the Git Hash with the Docker Tag of an image that gets deployed by ArgoCD. 

### Description

Currently we have a GitLab pipeline that builds our Docker Images. However we use a different repo for the Helm Chart which ArgoCD uses to deploy the application. With the current configuration DevLake can't correlate the build and the deployment because they happen in different repos with different commit SHAs.

The idea with the change is that the image tags are the build SHA in our setup. So if query-able I could use the image tag and the commit hash to match the build and deployment.

The metadata is already ingested by the plugin under the `data` column in the `_raw_argocd_api_sync_operations` table. However it doesn't get processed into the `_tool_argocd_sync_operations` table and therefore can't be queried by the Blueprint.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

rbstp:
I can have a look at it this weekend.

rbstp:
If you can validate the PR https://github.com/apache/incubator-devlake/pull/8631 corresponds to your needs @ppapp92

ppapp92:
@rbstp This is perfect! Thank you so much for implementing this.

PR #8631, opened 2025-11-02
feat(plugins): add image extraction and revision image model to argocd

rbstp:
### Summary
Adds ArgocdRevisionImage model + migration, and extractor fallback logic. The fallback is used for Argo Rollouts, which  does not directly provide the image information, it must be fetched via the ArgoCD Application table instead. A new table is introduced to be able to keep a history of those type of images.

It also supports deployments with multiple images in the same pod.

Closes issue [8630](https://github.com/apache/incubator-devlake/issues/8630)

### Screenshots
New tiles in Grafana Dashboard:
image

In the database:
image

Also available via the sync operations table:
image

### Other Information
If you can validate it is what you need @ppapp92


klesh (approved):
LGTM.
Thanks for your contribution.

PR #8639, opened 2025-11-13
fix(devcontainer): make Config UI and Grafana accessible from host when using devcontainers

### LINKED ISSUE #8638: [Question][Devcontainer] config-ui and grafana not accessible from outside the dev container
## Question
I'm using the dev container for attempt at local development.
Using the default configuration and running `make configure-dev`, I wasn't able to access the config-ui from outside the container.

After some digging, I found it was because Vite's dev server binds to localhost by default, which is only reachable inside the container.

I was able to resolve it by adding the `--host` parameter to the `"start": "vite"` inside the `package.json` file.
This makes it listen on 0.0.0.0, which lets VS Code's port-forwarder hook in.

**EDIT**:
I also run into a problem with accessing and then querying grafana when proxied via config-ui due to exposure and then origin not allowed issues.
I'm able to resolve it by modifying the vite dev server so grafana proxy targets to `http://grafana:3000` and disabling changeOrigin.

My question is, am I missing a way to make it work without any changes?
I'm wondering if the original author of the dev container had it work in a different way.

If not, I'm willing to contribute a PR with this change that should allow config-ui and grafana access from outside the dev container out-of-the-box (without affecting local development).

Thanks!

amnir:
### ⚠️ Pre Checklist

### Summary
Make config-ui dev server accessible outside the container when started by listening on 0.0.0.0 with the `--host` flag.
Also ensure it uses port 4000 or fails to start with `--strictPort` to prevent silent port change that will mismatch with the exported dev container port.

In addition, it makes Grafana accessible from dev container' config-ui by making the Grafana origin and origin change configurable and injected via the dev container environment variables.

web socket proxying (ws) is enabled regardless of dev container or not as it allows for features like grafana query auto-complete to work via the dev server proxy.

Closes [#8638](https://github.com/apache/incubator-devlake/issues/8638)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,7 +15,6 @@
  * limitations under the License.
  *
  */
-
 {
   "name": "Go",
   "dockerComposeFile": "docker-compose.yml",
@@ -36,6 +35,10 @@
     4000,
     8080
   ],
+  "containerEnv": {
+    "VITE_GRAFANA_URL": "http://grafana:3000",
+    "VITE_GRAFANA_CHANGE_ORIGIN": "false"
+  },
   "postCreateCommand": "npm install commitizen -g",
   "customizations": {
     "vscode": {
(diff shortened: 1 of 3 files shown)


rbstp:
I can confirm the changes work well in devcontainer to expose Grafana and the UI on port 4000.

PR #8642, opened 2025-11-16
fix(plugins): add missing scope-config and missing projects in grafana for argocd

rbstp:
### Summary
The scope-config was missing in the initial implementation, which prevent the DORA metrics to be collected.

In Grafana, you could not select the projects, so only ALL was available as an option. I also moved the graphs under the correct sections.

For the comment in the issue https://github.com/apache/incubator-devlake/issues/5207#issuecomment-3537781398

### Screenshots
scope-config
image

Grafana
image

### Other Information
Used the config from https://github.com/apache/incubator-devlake/pull/8639 for dev-containers and it worked without issue.


ayushkmr (approved):
LGTM

  Minor Notes:
  - Helper function firstNonEmpty() could be reusable across plugins

rbstp:
@klesh could you have a look at the PR please

klesh (approved):
LGTM

PR #8706, opened 2026-02-10
fix(plugins): add missed DomainTypes in ArgoCD plugin

abdonrd:
### Summary

Add missed `DomainTypes` in ArgoCD plugin to fix this problem: https://github.com/apache/incubator-devlake/issues/5207#issuecomment-3843222503

### Does this close any open issues?

https://github.com/apache/incubator-devlake/issues/5207#issuecomment-3843222503

PR #8712, opened 2026-02-12
fix(plugins): add missed repoURL in ArgoCD plugin

abdonrd:
### Summary

Testing ArgoCD plugin (https://github.com/apache/incubator-devlake/issues/5207), after:
- https://github.com/apache/incubator-devlake/pull/8706

I noticed that the Median Lead Time for Changes metric in DORA wasn't being reflected.

I saw that in `cicd_deployment_commits`, the `repo_url` field contained the ArgoCD app name name instead of the repository URL... And then I saw that the `repo_url` field in `_tool_argocd_applications` was null.


rbstp:
the change looks good 👍🏻

klesh (approved):

PR #8731, opened 2026-02-25
[fix][SonarQube]: allow project list with non-admin (Browse) token (fixes #8710)

### LINKED ISSUE #8710: [Question][SonarQube] Is Enterprise Edition v2025.1.3 supported
## Question
Hi. I was wondering whether SonarQube Enterprise Edition v2025.1.3 is supported ?

I have the problem that I can add the connection in DevLake (SonarQube Server) and using "Test Connection" it tells me that "Test Connection Succesfully". 

But when I want to add a data scope, I only see an empty box with no options (see screenshot).

Is there anything known or anything to do for me here ?

## Screenshots

Image

## Additional context
Add any other context here.

jbsmith7741:
### Summary
When adding a SonarQube data scope, the project list was empty even though "Test Connection" succeeded when using a non-admin token. The plugin was calling the projects/search API, which requires admin (or equivalent) permissions. This PR switches to components/search_projects, which works with a token that has Browse permission, so users can list projects and add scopes without an admin token.

Backend: use `components/search_projects` instead of `projects/search` in `api/blueprint_v200.go` (GetApiProject) and `api/remote_api.go` (querySonarqubeProjects). 

Config UI: add a subLabel on the token field stating that a token with Browse permission is sufficient for listing projects and collecting issues, hotspots, and file metrics; listing Accounts may still require a system admin token on some instances.

Closes #8710 

### Screenshots

### Other Information
No change to connection validation; "Test Connection" behavior is unchanged.

### DIFF
--- a/backend/plugins/sonarqube/api/blueprint_v200.go
+++ b/backend/plugins/sonarqube/api/blueprint_v200.go
@@ -131,7 +131,8 @@ func GetApiProject(
 	}
 	query := url.Values{}
 	query.Set("q", projectKey)
-	res, err := apiClient.Get("projects/search", query, nil)
+	// Use components/search_projects for consistency and normal-token (Browse) support.
+	res, err := apiClient.Get("components/search_projects", query, nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


klesh (approved):
LGTM

PR #8737, opened 2026-02-28
fix(q_dev): prevent data duplication in user_report and user_data tables

warren830:
## Summary
- Replace auto-increment ID (`common.Model`) with composite primary keys (`common.NoPKModel`) on `_tool_q_dev_user_report` and `_tool_q_dev_user_data` tables to enable proper deduplication
- Switch `db.Create()` to `db.CreateOrUpdate()` in s3_data_extractor so re-extracted data updates existing rows instead of inserting duplicates
- Add migration to drop/rebuild tables with new PKs and reset `s3_file_meta.processed` flag to trigger clean re-extraction

## Test plan
- [x] `go build ./plugins/q_dev/...` passes
- [x] `go test ./plugins/q_dev/...` passes
- [ ] Verify migration runs cleanly on dev environment
- [ ] Verify `SELECT ... GROUP BY ... HAVING COUNT(*) > 1` returns no duplicates after re-extraction
- [ ] Verify Grafana dashboard "Total Credits Used" shows correct value

### DIFF
--- a/backend/plugins/q_dev/models/migrationscripts/20260228_fix_dedup_user_tables.go
+++ b/backend/plugins/q_dev/models/migrationscripts/20260228_fix_dedup_user_tables.go
@@ -1,0 +1,60 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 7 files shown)


abeizn (approved):
lgtm

PR #8747, opened 2026-03-03
feat(gh-copilot): add support for organization daily user metrics

ReeceXW:
### Summary
The Copilot implementation only supports enterprise daily user metrics but a near identical endpoints exists for organization daily user metrics. This will conditionally use this if no enterprise connection is configured.

The report json files appear identical (I can't pull a real sample because I don't have enterprise, but the structure already defined appears compatible) except for an enterprise id field; it is an organization id field instead

N/A

### Screenshots
N/A nothing fundamentally changed


klesh (approved):
Looks like a good addition to me. 
@ewega  What do you think?

ewega:
Thanks for the tag @klesh ! @ReeceXW this looks great left a nitpick but LGTM. Awesome work

klesh:
Thank you all.

PR #8743, opened 2026-03-01
fix(gitlab): add missing repos scope in project_mapping

spiffaz:
## Summary
- GitLab's `makeScopeV200` did not create a `repos` scope when `scopeConfig.Entities` was empty or only contained `CROSS`, causing `project_mapping` to have no `table='repos'` row
- This broke downstream DORA metrics, PR-issue linking, and all Grafana PR dashboard panels that join on `project_mapping` with `table='repos'`
- The fix aligns GitLab with the GitHub plugin's `makeScopesV200` (lines 155-160) by:
  1. Defaulting empty entities to `plugin.DOMAIN_TYPES`
  2. Adding `DOMAIN_TYPE_CROSS` to the repo scope condition

Closes #8742

## Changes
- `backend/plugins/gitlab/api/blueprint_v200.go` - two-line fix in `makeScopeV200`
- `backend/plugins/gitlab/api/blueprint_V200_test.go` - two new test cases:
  - `TestMakeScopesWithEmptyEntities` - verifies empty entities default to all domain types
  - `TestMakeScopesWithCrossEntity` - verifies CROSS entity triggers repo scope creation

## Test plan
- [x] All existing tests pass (`TestMakeScopes`, `TestMakeDataSourcePipelinePlanV200`)
- [x] New tests pass for both empty entities and CROSS entity scenarios
- [x] Tests verified in `mericodev/lake-builder:latest` Docker container (same as CI)

## Note
The same issue also affects Bitbucket (missing empty default) and Azure DevOps (missing both). If this fix is accepted, I am happy to submit follow-up PRs for those plugins.

PR #8750, opened 2026-03-03
fix(bitbucket): default empty entities to all domain types in makeScopesV200

### LINKED ISSUE #8748: [Bug][Bitbucket] project_mapping missing repos scope when entities empty
## Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues) and found no similar issues.

## What happened

In `backend/plugins/bitbucket/api/blueprint_v200.go`, the `makeScopesV200` function does not default `scopeConfig.Entities` to `plugin.DOMAIN_TYPES` when entities are empty. This means that when a blueprint is created without specifying entities (which is common in the UI), no scopes are generated at all, and the `project_mapping` table ends up with no repo rows.

This is the same bug that was fixed for GitLab in #8743 (which fixed #8742). While Bitbucket Cloud already has `DOMAIN_TYPE_CROSS` in the repo scope condition (unlike GitLab before the fix), it is still missing the empty-entities default.

When `scopeConfig.Entities` is empty, it should default to `plugin.DOMAIN_TYPES` so that all domain scopes (repos, boards, cicd_scope) are created in `project_mapping`.

## How to reproduce

1. Create a Bitbucket Cloud connection and configure a blueprint without specifying entities (leave entities empty)
2. Run the blueprint
3. Check the `project_mapping` table - no rows are created for this scope

## Anything else

Follow-up from #8742 / #8743. The fix is to add the same empty-entities default that was applied to GitLab.

## Version

Latest main branch

spiffaz:
## Summary

- When `scopeConfig.Entities` is empty (common when no entities are explicitly selected in the UI), `makeScopesV200` produced zero scopes, leaving `project_mapping` with no repo rows
- Adds the same empty-entities default applied to GitLab in #8743
- Adds unit tests for normal entities, empty entities, and CROSS entity scenarios

## How has this been tested?

```bash
docker run --rm -v $(pwd):/app -w /app/backend mericodev/lake-builder:latest \
  bash -c "go test ./plugins/bitbucket/api/ -v -run TestMakeScopes"
```

All 3 tests pass:
- `TestMakeScopes` - CODE + TICKET + CICD = 3 scopes
- `TestMakeScopesWithEmptyEntities` - empty entities defaults to all domain types = 3 scopes
- `TestMakeScopesWithCrossEntity` - CROSS + TICKET = 2 scopes (repos + boards)

## What is the fix?

After generating the domain ID, default empty entities to `plugin.DOMAIN_TYPES`:

```go
if len(scopeConfig.Entities) == 0 {
    scopeConfig.Entities = plugin.DOMAIN_TYPES
}
```

Bitbucket Cloud already had `DOMAIN_TYPE_CROSS` in the repo scope condition (unlike GitLab before #8743), so only the empty-entities default was needed.

Closes #8748

### DIFF
--- a/backend/plugins/bitbucket/api/blueprint_v200.go
+++ b/backend/plugins/bitbucket/api/blueprint_v200.go
@@ -154,6 +154,11 @@ func makeScopesV200(
 		scope, scopeConfig := scopeDetail.Scope, scopeDetail.ScopeConfig
 		id := idgen.Generate(connection.ID, scope.BitbucketId)
 
+		// if no entities specified, use all entities enabled by default
+		if len(scopeConfig.Entities) == 0 {
+			scopeConfig.Entities = plugin.DOMAIN_TYPES
+		}
+
 		if utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CODE_REVIEW) ||
 			utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CODE) ||
 			utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CROSS) {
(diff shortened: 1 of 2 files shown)


klesh (approved):
LGTM. Thanks for your contribution.

PR #8751, opened 2026-03-03
fix(azuredevops): default empty entities and add CROSS to repo scope in makeScopeV200

### LINKED ISSUE #8749: [Bug][Azure DevOps] project_mapping missing repos scope when entities empty or CROSS-only
## Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues) and found no similar issues.

## What happened

In `backend/plugins/azuredevops_go/api/blueprint_v200.go`, the `makeScopeV200` function has two issues:

1. It does not default `scopeConfig.Entities` to `plugin.DOMAIN_TYPES` when entities are empty. When a blueprint is created without specifying entities (common in the UI), no scopes are generated and `project_mapping` ends up with no rows.

2. The repo scope condition checks for `DOMAIN_TYPE_CODE_REVIEW` and `DOMAIN_TYPE_CODE` but not `DOMAIN_TYPE_CROSS`. This means that when only `CROSS` is selected, the repo scope is missing from `project_mapping`, which breaks DORA metrics that depend on the cross-domain join.

These are the same bugs that were fixed for GitLab in #8743 (which fixed #8742).

1. When `scopeConfig.Entities` is empty, it should default to `plugin.DOMAIN_TYPES` so that all domain scopes are created in `project_mapping`.
2. The repo scope condition should also check for `DOMAIN_TYPE_CROSS`, consistent with GitLab and Bitbucket.

## How to reproduce

1. Create an Azure DevOps connection and configure a blueprint without specifying entities (leave entities empty)
2. Run the blueprint
3. Check the `project_mapping` table - no rows are created for this scope

Or:

1. Configure an Azure DevOps blueprint with only `CROSS` entity
2. Run the blueprint
3. Check `project_mapping` - the repo scope is missing

## Anything else

Follow-up from #8742 / #8743. The fix applies the same pattern used in GitLab.

## Version

Latest main branch

spiffaz:
## Summary

- When `scopeConfig.Entities` is empty (common when no entities are explicitly selected in the UI), `makeScopeV200` produced zero scopes, leaving `project_mapping` with no rows
- The repo scope condition did not check for `DOMAIN_TYPE_CROSS`, so selecting only CROSS would not create a repo scope, breaking DORA metrics
- Adds the same fixes applied to GitLab in #8743
- Adds unit tests for empty entities and CROSS entity scenarios

## How has this been tested?

```bash
docker run --rm -v $(pwd):/app -w /app/backend mericodev/lake-builder:latest \
  bash -c "go test ./plugins/azuredevops_go/api/ -v -run TestMakeScopes"
```

All tests pass (new + existing):
- `TestMakeScopes` - CODE + TICKET + CICD = 3 scopes
- `TestMakeScopesWithEmptyEntities` - empty entities defaults to all domain types = 3 scopes
- `TestMakeScopesWithCrossEntity` - CROSS + TICKET = 2 scopes (repos + boards)
- `TestMakeDataSourcePipelinePlanV200` - existing test still passes
- `TestMakeRemoteRepoScopes` - existing test still passes

## What is the fix?

1. After generating the domain ID, default empty entities to `plugin.DOMAIN_TYPES`:

```go
if len(scopeConfig.Entities) == 0 {
    scopeConfig.Entities = plugin.DOMAIN_TYPES
}
```

2. Add `DOMAIN_TYPE_CROSS` to the repo scope condition:

```go
if utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CODE_REVIEW) ||
    utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CODE) ||
    utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CROSS) {
```

Closes #8749

### DIFF
--- a/backend/plugins/azuredevops_go/api/blueprint_v200.go
+++ b/backend/plugins/azuredevops_go/api/blueprint_v200.go
@@ -77,8 +77,14 @@ func makeScopeV200(
 		}
 		id := didgen.NewDomainIdGenerator(&models.AzuredevopsRepo{}).Generate(connectionId, azuredevopsRepo.Id)
 
+		// if no entities specified, use all entities enabled by default
+		if len(scopeConfig.Entities) == 0 {
+			scopeConfig.Entities = plugin.DOMAIN_TYPES
+		}
+
 		if utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CODE_REVIEW) ||
-			utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CODE) {
+			utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CODE) ||
+			utils.StringsContains(scopeConfig.Entities, plugin.DOMAIN_TYPE_CROSS) {
 			// if we don't need to collect gitex, we need to add repo to scopes here
 			scopeRepo := code.NewRepo(id, azuredevopsRepo.Name)
 			sc = append(sc, scopeRepo)
--- a/backend/plugins/azuredevops_go/api/blueprint_v200_test.go
+++ b/backend/plugins/azuredevops_go/api/blueprint_v200_test.go
@@ -78,6 +78,64 @@ func TestMakeScopes(t *testing.T) {
 	assert.Equal(t, actualScopes[2].ScopeId(), expectDomainScopeId)
 }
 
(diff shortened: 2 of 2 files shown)


klesh (approved):
LGTM

PR #8701, opened 2026-02-08
fix: CircleCI column naming inconsistency (#8497)

varsis:
Fixes #8497

## Summary
Fixed CircleCI plugin incremental data collection failure due to column naming inconsistency.

## Root Cause
- CircleCI API v2 returns `created_at`/`updated_at` (snake_case)
- Pipeline model used `CreatedDate`/`UpdatedDate` → DB columns `created_date`/`updated_date`
- Workflow/Job models correctly use `CreatedAt`/`UpdatedAt` → DB columns `created_at`/`updated_at`
- Collectors queried workflows/jobs with wrong column name `created_date` (didn't exist in those tables)

## Changes
- Fixed workflow/job collector queries: `created_date` → `created_at`
- Standardized pipeline model fields to match API: `CreatedDate`/`UpdatedDate` → `CreatedAt`/`UpdatedAt`
- Created DB migration to rename pipeline table columns
- Updated archived model for consistency
- Fixed test CSV headers and removed obsolete ignore fields

## Breaking Changes
- DB schema change (column renames for pipelines table)
- Requires migration for existing deployments
- No API contract changes

## Test Plan
- [x] Code builds successfully
- [x] Migration handles both MySQL and PostgreSQL
- [x] Safe for clean installs (checks column existence before renaming)
- [x] Preserves data during migration

PR #8766, opened 2026-03-13
fix(circleci): correct column name in incremental collection WHERE clause

### LINKED ISSUE #8497: [Bug][CircleCI] plugin fails on second (incremental) data collection due to SQL error: unknown column created_date
### Search before asking

### What happened

On running data collection pipeline second time (incremental mode), the first CircleCI task fails with an SQL error:
` | /usr/local/go/src/runtime/asm_amd64.s:1598 Wraps: (4) Error 1054 (42S22): Unknown column 'created_date' in 'where clause' (500) Wraps: (5) Error 1054 (42S22): Unknown column 'created_date' in 'where clause' Error types: (1) *withstack.withStack (2) *errutil.withPrefix (3) *withstack.withStack (4) *errutil.withPrefix (5) *mysql.MySQLError`

On running 'Collect Data in Full Refresh Mode' everything goes fine again, and the data collection after that fails with the same error.

The incremental data collection to run successfully without SQL errors. If a column like created_date is needed, I’d expect it to be created automatically or handled in a way that avoids failures.

### How to reproduce

1. Set up a new DevLake project.
2. Add CircleCI connection.
3. Run the first data collection pipeline in full refresh mode – it completes successfully.
4. Run the same pipeline again (incremental mode) without changing any config.

### Anything else

The same happens in a new project on the first run with the same connection. And works fine in the Full Refresh Mode

### Version
v1.0.3-beta1(Docker Image)
v1.0.3-beta2(Docker Image)
UPD: v1.0.2 is also impacted.

### Code of Conduct

jbsmith7741:
### Summary

Fixes a MySQL error that causes every incremental CircleCI sync to fail
after the first successful run.

Both `workflow_collector.go` and `job_collector.go` build a
`WHERE created_date > ?` clause when `isIncremental = true`, but the
actual column in `_tool_circleci_workflows` is `created_at` (derived
from the Go struct field `CircleciWorkflow.CreatedAt`). 

Fix: change the two `WHERE` clause strings from `created_date` to
`created_at` to match the actual schema. No model changes, no
migration scripts, and no other files are touched.

Closes #8497

### Screenshots
NA

### Other Information

alternative solution would be to rename the column to create_date which would require a DB migration script and changes through the code. If desired I can switch to that solution, but this was the simplest fix with the less amount of code changes.

### DIFF
--- a/backend/plugins/circleci/tasks/job_collector.go
+++ b/backend/plugins/circleci/tasks/job_collector.go
@@ -60,7 +60,7 @@ func CollectJobs(taskCtx plugin.SubTaskContext) errors.Error {
 				}
 
 				if isIncremental {
-					clauses = append(clauses, dal.Where("created_date > ?", createdAfter))
+					clauses = append(clauses, dal.Where("created_at > ?", createdAfter))
 				}
 
 				db := taskCtx.GetDal()
--- a/backend/plugins/circleci/tasks/workflow_collector.go
+++ b/backend/plugins/circleci/tasks/workflow_collector.go
@@ -61,7 +61,7 @@ func CollectWorkflows(taskCtx plugin.SubTaskContext) errors.Error {
 				}
 
 				if isIncremental {
-					clauses = append(clauses, dal.Where("created_date > ?", createdAfter))
+					clauses = append(clauses, dal.Where("created_at > ?", createdAfter))
 				}
 
 				db := taskCtx.GetDal()


klesh:
`created_at` is reserved for devlake to track when was the record got created in the database.
I think a better solution is to rename the column to `created_date`.
Duplicated to #8701

jbsmith7741:
This issue has been resolved in another pull request.

PR #8777, opened 2026-03-17
feat(linker): link when branch names contain issue keys

ReeceXW:
### Summary
When branch names include an issue key they should be linked. I've added some new test cases and validated that the head ref lookup is working as expected.


klesh (approved):

klesh:
Thanks for your contribution.

PR #8785, opened 2026-03-20
fix(qdev): configuration and execution not working with iam auth

ReeceXW:
### Summary
In the UI you are able to choose IAM authentication instead of an access key and secret but the API doesn't actually support this; it will just fail. I've added the type parameter to the API and the necessary AWS configuration to make this work.


klesh:
@warren830, could you please review this PR when you have a moment? Thank you!

warren830 (commented):
Thanks for the PR — overall clean and well-scoped. IAM credential chain handling is correct, backward compatibility is solid (defaults to `access_key`, migration backfills existing rows), and test coverage hits the key paths.

One additional concern not tied to a specific diff line:

**Potential bug (medium):** `MergeFromRequest` in `models/connection.go` uses `DecodeMapStruct` to overwrite fields from the request body. If a PATCH request omits `authType`, the field may be decoded as `""`, which then silently falls back to `"access_key"` via the validation default. This could unexpectedly flip an `iam_role` connection back to `access_key`. Consider adding the same guard pattern used for `SecretAccessKey` — preserve the existing `AuthType` when the incoming value is empty.

A few inline suggestions below — #1 (mutation in validation) and the `MergeFromRequest` issue above are worth addressing before merge; the rest are nice-to-haves.

ReeceXW:
@warren830 I've implemented those suggestions, tests all passing :)

klesh:
Close for inactivity, feel free to reopen if you wish.

PR #8746, opened 2026-03-02
feat: GitHub App token refresh

lrf-nitro:
### ⚠️ Pre Checklist

> Please complete _ALL_ items in this checklist, and remove before submitting

- [x] I have read through the [Contributing Documentation](https://devlake.apache.org/community/).
- [x] I have added relevant tests.
- [x] I have added relevant documentation.
- [x] I will add labels to the PR, such as `pr-type/bug-fix`, `pr-type/feature-development`, etc.

### Summary
This PR adds support for GitHub App token refresh. It is based on issues we faced due to token expiration with large projects. The initial idea was to fix the token refresh issue only, but once that was working other issues appeared - e.g., the connections request from the UI to the GitHub plugin returned incorrect/corrupt token values which also had to be fixed. 

Claude Opus 4.6 was used during the work.

### Does this close any open issues?
No.

### Screenshots
Include any relevant screenshots here.

### Other Information
Summary notes:

[PR_SUMMARY.md](https://github.com/user-attachments/files/25696778/PR_SUMMARY.md)

PR #8791, opened 2026-03-22
Fix/graphql client token refresh

### LINKED ISSUE #8788: [Bug][github_graphql] GitHub App installation token not refreshed for GraphQL client, panic and pipeline crash on large repos
### Search before asking

### What happened

When collecting data from a large repository using GitHub App authentication, the pipeline crashes after approximately 1 hour with the following panic:

```
panic: non-200 OK status code: 401 Unauthorized body: "{
  \"message\": \"Bad credentials\",
  \"documentation_url\": \"https://docs.github.com/rest\",
  \"status\": \"401\"
}"
goroutine 11186 [running]:
github.com/apache/incubator-devlake/helpers/pluginhelper/api.(*GraphqlAsyncClient).updateRateRemaining.func1()
	/app/helpers/pluginhelper/api/graphql_async_client.go:129
created by github.com/apache/incubator-devlake/helpers/pluginhelper/api.(*GraphqlAsyncClient).updateRateRemaining
	/app/helpers/pluginhelper/api/graphql_async_client.go:117
```

There are two independent bugs that combine to cause this:

**Bug 1: Process crash on any error in the rate-limit goroutine**

`graphql_async_client.go:129` calls `panic(err)` inside the background goroutine that periodically polls the GitHub GraphQL rate limit. Any error, including a transient 401, immediately kills the entire DevLake process.

**Bug 2: GraphQL HTTP client holds a frozen, never-refreshed token (root cause)**

In `backend/plugins/github_graphql/impl/impl.go` (lines 183-209), the GraphQL HTTP client is built with `oauth2.StaticTokenSource`:

```go
tokens := strings.Split(connection.Token, ",")
src := oauth2.StaticTokenSource(
    &oauth2.Token{AccessToken: tokens[0]},
)
httpClient := oauth2.NewClient(oauthContext, src)
```

This freezes the token to the value at task start time. The client has no connection to the `TokenProvider` / `RefreshRoundTripper` machinery that already handles token refresh for the REST client.

By contrast, `CreateApiClient` (called just above at line 173) correctly wires up `NewAppInstallationTokenProvider` + `RefreshRoundTripper` and auto-refreshes on 401. This was added in #8746. The GraphQL client was not updated as part of that PR and still bypasses the refresh infrastructure with its own isolated `oauth2` HTTP client.

Since GitHub App installation tokens expire after 1 hour, any pipeline running longer than that will encounter 401s on all GraphQL requests. The rate-limit polling goroutine then hits the 401 and panics, crashing the process.

Note: the documentation page at https://devlake.apache.org/docs/Configuration/GitHub/#github-apps-beta currently states "The GitHub App token expires every hour and does not auto-renew." This was accurate before #8746 but is now stale for the REST path. The GraphQL path is the remaining gap.

- The GitHub App installation token should be transparently refreshed for GraphQL requests, the same way it already is for REST requests.
- If token refresh fails or any other error occurs in the background rate-limit goroutine, the error should be logged and retried, not cause a process-wide panic.
- Pipelines against large repositories should complete successfully without requiring users to fall back to Personal Access Tokens.

### How to reproduce

1. Configure a GitHub connection using **GitHub App** authentication (not PAT).
2. Create a blueprint targeting a large repository whose full collection takes more than 1 hour.
3. Trigger the pipeline.
4. After ~1 hour, observe the panic in the logs and the pipeline crashing.

### Anything else

**Affected files:**

| File | Issue |
|---|---|
| `backend/helpers/pluginhelper/api/graphql_async_client.go:129` | `panic(err)` in background goroutine crashes the process on any rate-limit poll error |
| `backend/plugins/github_graphql/impl/impl.go:183-209` | `oauth2.StaticTokenSource` freezes the token at task start; GraphQL client is not wired to `TokenProvider` / `RefreshRoundTripper` |

**Suggested fix:**

1. In `graphql_async_client.go`, replace `panic(err)` with graceful error handling (log the error and reschedule the next poll) so a transient 401 does not crash the process.

2. In `github_graphql/impl/impl.go`, instead of creating a new `oauth2.StaticTokenSource`-backed HTTP client, reuse `apiClient.GetClient()` -- the REST client's underlying `*http.Client` -- which already carries the `RefreshRoundTripper`. This is safe: `ApiAsyncClient` embeds `*ApiClient` which exposes `GetClient()`, and proxy is already configured inside `NewApiClientFromConnection` so the `oauthContext` proxy block becomes redundant. Token injection is handled by the `RefreshRoundTripper`, making the `oauth2` wrapper unnecessary.

Related: #8746 (added token refresh for REST path)

### Version

v1.0.3-beta10

### Code of Conduct

Ke-vin-S:
### Summary
This PR fixes two issues in the GitHub GraphQL client related to stability and authentication during long-running pipelines:

1. **Prevent process crash on rate-limit polling errors**
   - Replaces `panic` in `GraphqlAsyncClient` background goroutine with graceful error handling.
   - On failure, errors are logged and the system retries in the next polling cycle while retaining the last known rate limit.

2. **Enable automatic token refresh for GraphQL requests**
   - Replaces `oauth2.StaticTokenSource`-based HTTP client with the underlying `http.Client` from `ApiAsyncClient`.
   - This allows GraphQL requests to reuse the `RefreshRoundTripper`, enabling automatic token refresh for GitHub App installation tokens.

3. **Refactor authentication into shared HTTP client**
   - Extracts auth and transport setup into `CreateAuthenticatedHttpClient`.
   - Enables reuse across REST and GraphQL clients.

4. **Introduce GraphQL client factory**
   - Adds `CreateGraphqlClient` using shared authenticated HTTP client.
   - Replaces manual client setup while preserving existing behavior.

5. **Support static and refreshable tokens**
   - Adds `StaticRoundTripper` for PAT authentication.
   - Ensures correct handling for both token types.

---

Closes #8788

---

### Screenshots
N/A

---

#### Design decisions

- **Avoid panic in background goroutines**  
  Rate-limit polling is non-critical and should not crash the entire pipeline. Errors are now handled gracefully.

- **Separate handling of initial vs runtime failures**
  - Initial rate-limit fetch failure → fallback to default (`5000`) to prevent deadlock.
  - Runtime failure → retain last known `rateRemaining` to preserve correctness.

- **Use eventual consistency for rate limiting**  
  Rate-limit polling is best-effort; failures should not block or terminate execution.

- **Centralize transport layer for consistency**  
  Both GraphQL and REST clients now use a shared authenticated HTTP client, ensuring consistent authentication, proxy, and retry behavior.

- **Avoid `StaticTokenSource`**  
  Static tokens break long-running pipelines due to expiration. Using `RefreshRoundTripper` ensures seamless token rotation.

- **Handle static vs refreshable tokens explicitly**  
  Shared transport distinguishes between PAT (static) and installation (refreshable) tokens to avoid incorrect refresh/retry behavior.

---


Ke-vin-S:
@klesh I’ve pushed an update for this—could you please take a look? Happy to make changes if needed.

backend/plugins/github/tasks/http_client.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if tp != nil {
+		baseClient.Transport = token.NewRefreshRoundTripper(baseTransport, tp)
+		logger.Info(
+			"Installed token refresh round tripper for connection %d (authMethod=%s)",
+			connection.ID,
+			connection.AuthMethod,
+		)
+
+	} else if connection.Token != "" {
```
danielemoraschi: DevLake allows multiple comma-separated tokens in `connection.Token` to bypass rate limits (handled by a core round-robin transport in the REST ApiClient). Adding this forces `Authorization: Bearer token1,token2`, which will cause 401 Unauthorized errors from the GitHub API breaking the round-robin logic. Original logic only added a RoundTripper when `tp != nil`.
The implementation in `impl.go` explicitly handled this via `strings.Split(connection.Token, ",")[0]`, injecting the unparsed comma-separated string breaks existing multi-token configurations.

danielemoraschi (commented):
Left some comments for your review. Thanks!

klesh:
@Ke-vin-S Great job! Would you like to take a look at @danielemoraschi’s comments?

Ke-vin-S:
> @Ke-vin-S Great job! Would you like to take a look at @danielemoraschi’s comments?

Yes, currently working on it

klesh:
Thanks for your contribution! Could you please take a look at the failed checks?

zaiddialpad:
Hi — just wanted to share that we're actively hitting both issues this PR addresses in production, and this fix would be hugely impactful for us.

## Our setup

- **DevLake v1.0.3-beta9** on GKE (Helm chart)
- **GitHub App auth** across 4 connections (408 repos total)
- Largest repo: ~20K PRs, ~1,200 currently OPEN

## What we're experiencing

### 1. The `updateRateRemaining` panic (the primary fix in this PR)

Our lake pod is at **42 restarts** — almost entirely from this panic. The sequence is:

1. The `github_graphql` "Collect Pull Requests" subtask enters Phase 2 (refetching all OPEN PRs)
2. Phase 2 sends heavy compound GraphQL queries (100 PRs per batch) that trigger **502 Bad Gateway** from GitHub
3. The retry loop runs for 10+ retries × 120s each, easily exceeding 1 hour
4. Our GitHub App token expires (~1hr lifetime)
5. The background `updateRateRemaining` goroutine calls `/rate_limit` with the expired token → **401 Unauthorized**
6. `panic()` at `graphql_async_client.go:129` → pod crash

```
panic: non-200 OK status code: 401 Unauthorized body: "Bad credentials"

goroutine 3660 [running]:
github.com/apache/incubator-devlake/helpers/pluginhelper/api.(*GraphqlAsyncClient).updateRateRemaining.func1()
    /app/helpers/pluginhelper/api/graphql_async_client.go:129 +0x166
```

### 2. GraphQL client token refresh (the second fix in this PR)

Since the GraphQL client uses a `StaticTokenSource` that freezes the token at task start, any task running longer than ~1 hour with GitHub App auth is guaranteed to eventually fail. For us, the most important repo's `github_graphql` task takes 2-3 hours even when it succeeds, so we hit this on every single pipeline run.

The REST client was already fixed for this via #8746, but without this PR the GraphQL client remains vulnerable.

## Impact on us

- **Largest Repo has failed on 4 of the last 6 daily pipeline runs** due to this combination
- Each failure wastes 15+ hours of pipeline time (retries + pod restarts + re-queuing)
- The pod crashes affect **all** running tasks, not just the one that triggered the panic


## Summary

This PR addresses exactly the two bugs compounding against us:
1. The panic → replaced with graceful error handling ✅
2. The static token → wired up to `RefreshRoundTripper` for automatic renewal ✅

Would love to see this merged. Happy to help test if that would be useful — we have a production environment that reliably reproduces both issues on every nightly sync.

Ke-vin-S:
Fixed commit message formatting - changed underscores to hyphens in scopes to pass the lint check. Force-pushed to update all commits.

backend/plugins/github/tasks/http_client.go
Ke-vin-S: Implemented that in static round tripper

Ke-vin-S:
Hey @danielemoraschi , addressed your feedback. Would appreciate a second look.

Ke-vin-S:
> Hi — just wanted to share that we're actively hitting both issues this PR addresses in production, and this fix would be hugely impactful for us.
> ## Our setup
> 
>     * **DevLake v1.0.3-beta9** on GKE (Helm chart)
> 
>     * **GitHub App auth** across 4 connections (408 repos total)
> 
>     * Largest repo: ~20K PRs, ~1,200 currently OPEN
> 
> 
> ## What we're experiencing
> ### 1. The `updateRateRemaining` panic (the primary fix in this PR)
> 
> Our lake pod is at **42 restarts** — almost entirely from this panic. The sequence is:
> 
>     1. The `github_graphql` "Collect Pull Requests" subtask enters Phase 2 (refetching all OPEN PRs)
> 
>     2. Phase 2 sends heavy compound GraphQL queries (100 PRs per batch) that trigger **502 Bad Gateway** from GitHub
> 
>     3. The retry loop runs for 10+ retries × 120s each, easily exceeding 1 hour
> 
>     4. Our GitHub App token expires (~1hr lifetime)
> 
>     5. The background `updateRateRemaining` goroutine calls `/rate_limit` with the expired token → **401 Unauthorized**
> 
>     6. `panic()` at `graphql_async_client.go:129` → pod crash
> 
> 
> ```
> panic: non-200 OK status code: 401 Unauthorized body: "Bad credentials"
> 
> goroutine 3660 [running]:
> github.com/apache/incubator-devlake/helpers/pluginhelper/api.(*GraphqlAsyncClient).updateRateRemaining.func1()
>     /app/helpers/pluginhelper/api/graphql_async_client.go:129 +0x166
> ```
> 
> ### 2. GraphQL client token refresh (the second fix in this PR)
> 
> Since the GraphQL client uses a `StaticTokenSource` that freezes the token at task start, any task running longer than ~1 hour with GitHub App auth is guaranteed to eventually fail. For us, the most important repo's `github_graphql` task takes 2-3 hours even when it succeeds, so we hit this on every single pipeline run.
> 
> The REST client was already fixed for this via #8746, but without this PR the GraphQL client remains vulnerable.
> ## Impact on us
> 
>     * **Largest Repo has failed on 4 of the last 6 daily pipeline runs** due to this combination
> 
>     * Each failure wastes 15+ hours of pipeline time (retries + pod restarts + re-queuing)
> 
>     * The pod crashes affect **all** running tasks, not just the one that triggered the panic
> 
> 
> ## Summary
> 
> This PR addresses exactly the two bugs compounding against us:
> 
>     1. The panic → replaced with graceful error handling ✅
> 
>     2. The static token → wired up to `RefreshRoundTripper` for automatic renewal ✅
> 
> 
> Would love to see this merged. Happy to help test if that would be useful — we have a production environment that reliably reproduces both issues on every nightly sync.

If you get a chance to test against your setup before it's merged, that'd be great — but no pressure

klesh:
Hi, @danielemoraschi Would you like to take another look?

backend/plugins/github/token/round_tripper.go
```diff
@@ -93,3 +95,36 @@ func (rt *RefreshRoundTripper) roundTripWithRetry(req *http.Request, refreshAtte
...
+	}
+	if len(tokens) == 0 {
+		tokens = []string{rawToken}
+	}
+	return &StaticRoundTripper{base: base, tokens: tokens}
+}
+
+func (rt *StaticRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// always overrides headers put by SetupAuthentication, to make sure the token is always injected
+	tok := rt.tokens[rt.idx.Add(1)%uint64(len(rt.tokens))]
```
danielemoraschi: Uint64 starts at 0, `Add(1)` returns 1, so the first request uses `tokens[1]`, not `tokens[0]`. For single-token lists this is harmless (1%1==0), but multi-PAT connections skip the first token on first request. The old SetupAuthentication started at index 0.

backend/helpers/pluginhelper/api/graphql_async_client.go
```diff
@@ -126,7 +154,9 @@ func (apiClient *GraphqlAsyncClient) updateRateRemaining(rateRemaining int, rese
 		case <-time.After(nextDuring):
 			newRateRemaining, newResetAt, err := apiClient.getRateRemaining(apiClient.ctx, apiClient.client, apiClient.logger)
 			if err != nil {
-				panic(err)
+				apiClient.logger.Info("failed to update graphql rate limit, will retry next cycle: %v", err)
+				apiClient.updateRateRemaining(apiClient.rateRemaining, nil)
```
danielemoraschi: Potential deadlock on GitHub Enterprise? The Signal() at L139 only fires when `rateRemaining > 0`, so all goroutines waiting in Query (L183: for rateRemaining <= 0 { Wait() }) block forever. Since Github returns this error every cycle rate limit never recovers.

Check if use `max(apiClient.rateRemaining, defaultRateLimitConst)` instead of `apiClient.rateRemaining` works, to ensure the fallback is always positive and Signal() fires.

danielemoraschi (commented):
Just a couple of comments. But verified the proxy fix, token splitting, debug removal, and rate limit priority ordering and all seems good.

Ke-vin-S:
@klesh, @danielemoraschi 
Apologies for the delay on this PR. I've now implemented requested changes and would love to get anymore changes if any

klesh:
@Ke-vin-S Thank you for following up. Great job!  
@danielemoraschi Would you like to take another look? Thank you.

Ke-vin-S:
Closed and reopened to reset the PR base SHA. The lint failure is a false positive caused by stale CI metadata comparing against an older main head, flagging commits already merged upstream

klesh (approved):
It has been 2 weeks. I am merging the PR.
@Ke-vin-S  Thank you for your contribution and patience.

danielemoraschi:
Apologies, I was not that available the last few weeks 🙏 
Glad to see it landed!

klesh:
> Apologies, I was not that available the last few weeks 🙏 Glad to see it landed!

Don't be. Your input was important, and you are not obligated.

PR #8807, opened 2026-03-24
feature: Copilot per team metrics rework

la-tamas:
### ⚠️ Pre Checklist

### Summary

The PR adds functionality to:
- ingest teams and per team metrics from the GitHub API
- display ingested data on new dashboards filterable by teams and teams/users

None

### Screenshots

### Other Information


ewega (changes requested):
Thanks for the contribution friend! Seems like for this team sync you're using the org endpoints in GitHub. GitHub provides enterprise teams as well: https://docs.github.com/en/enterprise-cloud@latest/admin/concepts/enterprise-fundamentals/teams-in-an-enterprise

I think we have 2 options for this PR:
1) We should make that scoping explicit in behavior to avoid errors: if a connection is enterprise-only, these subtasks should be skipped rather than allowed to run and potentially fail. 
2) Adding Enterprise teams in this PR (but this increases the scope)

I suggest 1! Just wanted to check with you. If you think so as well I'll just open an issue. What do you think?

ewega:
Oh also noticed one thing we should also probably track separately: Looks like teams deleted or renamed in GitHub could linger in the dashboard selector after sync. Not blocking I'll just make sure to open a separate small issue for this one we merge this one

klesh:
The migration script linting is failing—could you please fix it?

la-tamas:
> Oh also noticed one thing we should also probably track separately: Looks like teams deleted or renamed in GitHub could linger in the dashboard selector after sync. Not blocking I'll just make sure to open a separate small issue for this one we merge this one

We also noticed this. I will deliver a solution here as well.

ewega:
Let's merge #8807 prior to this one given the change so we can resolve conflicts

PR #8825, opened 2026-04-08
fix(gh-copilot): throwing on parsing for a 204 response

ReeceXW:
### Summary
It's possible the user metrics endpoint can return a 204 No Content, but we are only ignoring a 404. This leads to trying to parse a response that isn't valid and collection stops.

I've added the exclusion and some more logging


klesh (approved):
LGTM.
Thanks for your contribution.

PR #8830, opened 2026-04-10
fix(ci): replace Poetry runtime setup with uv

### LINKED ISSUE #8776: [Refactor][CI/Docker] Replace poetry with uv for Python package management
## What and why to refactor

**Problem**: CI job "build image" fails when installing Poetry via the official installer.

The issue occurs in `backend/Dockerfile` during the `base` stage build:

```dockerfile
RUN curl -sSL https://install.python-poetry.org | python3 -
```

which results in:

```
Installing Poetry (2.3.2): Creating environment
Installing Poetry (2.3.2): Installing Poetry
Installing Poetry (2.3.2): An error occurred. Removing partial environment.
Poetry installation failed.
```

**Why refactor now**:
- Poetry installer is unreliable in CI
- Alternative approach (pipx) also has issues

Replace poetry with [uv](https://github.com/astral-sh/uv), a Rust-based Python package manager:

**Advantages:**
- 10-100x faster than poetry
- Compatible with pyproject.toml (no config changes needed)
- Minimal overhead in QEMU environments
- Actively maintained by Astral (ruff creators)

## Related issues

- Original issue: poetry timeout in CI arm64 builds
- Python 3.9 EOL concerns

## Additional context

**Root cause identified**: Poetry 2.3.x is incompatible with Python 3.9. The installer silently fails because it cannot create a working environment.

**Attempted workarounds:**

1. **curl installer fails** - Original method using `install.python-poetry.org`:
   ```
   Installing Poetry (2.3.2): Creating environment
   Installing Poetry (2.3.2): Installing Poetry
   Installing Poetry (2.3.2): An error occurred. Removing partial environment.
   Poetry installation failed.
   ```
   - **Root cause**: Poetry 2.3.x requires Python >= 3.10, but the devlake base image uses Python 3.9

2. [**Version pinning to 2.2.1**](https://github.com/apache/incubator-devlake/pull/8775) - Locking to last Python 3.9 compatible version:
   ```bash
   curl -sSL https://install.python-poetry.org | python3 - --version 2.2.1
   ```
   - Installation succeeds with Poetry 2.2.1
   - However, `poetry install` still times out in CI arm64 builds:
     ```
     Creating virtualenv azuredevops-ANnMAkq9-py3.9 in /app/.cache/pypoetry/virtualenvs
     failed to query /usr/local/bin/python3.9 with code -1 err: 'timed out'
     ```
   - Timeout occurs during virtualenv creation when Poetry queries the Python interpreter
   - This is exacerbated by QEMU emulation in cross-platform CI builds

3.[ **pipx install poetry**](https://github.com/apache/incubator-devlake/pull/8774) - Using pipx with version pinning:
   - Installation succeeded: `python3 -m pipx install poetry`
   - But `poetry install` still fails in CI arm64 builds:
     ```
     Creating virtualenv azuredevops-ANnMAkq9-py3.9 in /app/.cache/pypoetry/virtualenvs
     failed to query /usr/local/bin/python3.9 with code -1 err: 'timed out'
     ```
   - Root cause: Poetry queries Python interpreter during virtualenv creation, which times out in QEMU-emulated arm64 environments
   - Local arm64 builds succeed because they run on native hardware

aviralgarg05:
### Summary

This PR removes the runtime dependency on Poetry from the Python build path and switches the backend, test, and container workflows to `uv`.

The change is intentionally narrow:
- added a small shared helper to bootstrap `uv`, create or reuse project-local virtual environments, and run Python or pytest commands consistently
- updated the Azure DevOps and fake plugin build and run scripts to use that helper instead of Poetry
- updated the backend Python test runner to use the same flow and to avoid collecting tests from `.venv` directories
- updated the backend image and the lake-builder image to install `uv` instead of Poetry
- fixed the backend image copy step so the non-root `devlake` user can create project-local virtual environments during image builds

This addresses the CI and Docker breakage caused by the Poetry-based Python 3.9 setup, especially in cross-platform and emulated build paths.

Closes #8776

### Screenshots

No UI changes.

### Other Information

Validation completed against the updated flow:
- `make build-python` passed in the CI-style `mericodev/lake-builder:latest` container
- `make unit-test-python` passed in the same container
- `docker buildx build --platform linux/amd64 --target base ... backend` passed
- `docker build --target base ... backend` passed on the native platform

Relevant test results:
- Azure DevOps Python tests: 8 passed, 1 skipped
- PyDevLake Python tests: 10 passed

Suggested labels:
- `pr-type/bug-fix`


narrowizard:
Sorry, I'm not familiar with Python package management or DevLake's Python plugins. I only opened the issue because I encountered a timeout problem at that time, and I don't have the capability to review this modification. @klesh @aviralgarg05

klesh:
@keon94 Would you like to take a look?

klesh (approved):
Thanks for your contribution.

PR #8832, opened 2026-04-13
Fix/8787 cancel pipeline keeps running

### LINKED ISSUE #8787: [Bug][Pipeline][Gitextractor] Cancel pipeline returns 200 OK but pipeline keeps running
### Search before asking

### What happened

Clicking the cancel button in the UI (which fires `DELETE /api/pipelines/:id`) always returns
"Operation successfully completed" (HTTP 200), but the pipeline continues running. Observable
for 30+ minutes after the cancel request before it stops on its own.

This is a regression of #5585 and a go-git counterpart to #4188 (which only fixed libgit2).
Both were closed but the underlying causes were not fully addressed.

Three independent bugs combine to produce this behaviour:

---

**Bug 1: `CancelPipeline` silently discards errors from `CancelTask`**

`server/services/pipeline.go`:

```go
for _, pendingTask := range pendingTasks {
    _ = CancelTask(pendingTask.ID)  // error thrown away
}
```

`CancelTask` calls `runningTasks.Remove(taskId)`. If the task is not in the in-memory map
(race between pipeline stages, task not yet registered, or pod restart), `Remove` returns
`errors.NotFound`. The error is discarded, `cancel()` is never called, the goroutine keeps
running, and the API still returns 200 OK.

A second consequence: tasks in future pipeline stages (`TASK_CREATED`) are never in
`runningTasks`, so `CancelTask` silently fails for all of them. They remain `TASK_CREATED`
in the database and the pipeline status stays `TASK_RUNNING` until the goroutine naturally
finishes.

---

**Bug 2: `storeRepoSnapshot` in gitextractor ignores context cancellation (go-git path)**

`plugins/gitextractor/parser/repo_gogit.go`:

```go
func (r *GogitRepoCollector) storeRepoSnapshot(subtaskCtx plugin.SubTaskContext, commitList []*object.Commit) error {
    ctx := subtaskCtx.GetContext()
    for _, commit := range commitList {      // ← no ctx.Done() check between commits
        // ...
        for _, p := range patch.Stats() {
            blameResults, err := gogit.Blame(commit, fileName)  // ← no context parameter
```

`gogit.Blame()` has no context parameter, it performs a full in-process blame computation
and cannot be interrupted. For large repositories with thousands of commits, each touching
many files, this loop runs for **30+ minutes** and is completely unresponsive to context
cancellation. This is the primary cause of the long delay observed after pressing cancel.

Issue #4188 fixed the same problem for the libgit2 implementation (`repo_libgit2.go`) but
the go-git implementation was never addressed.

---

**Bug 3: Cancelled tasks are marked `TASK_FAILED` instead of `TASK_CANCELLED`**

`core/runner/run_task.go`: the deferred status update always writes `TASK_FAILED` when
`err != nil`, with no special case for context cancellation:

```go
dbe := db.UpdateColumns(task, []dal.DalSet{
    {ColumnName: "status", Value: models.TASK_FAILED},  // wrong for cancellations
    ...
})
```

The final pipeline status also becomes `TASK_FAILED` or `TASK_PARTIAL` rather than
`TASK_CANCELLED`, making it impossible to distinguish a failed run from a cancelled one in
the UI or dashboards.

- Pressing cancel on a running pipeline stops it promptly (within seconds for HTTP-based plugins)
- The pipeline and all its tasks (running and not-yet-started) are immediately marked `TASK_CANCELLED` in the database
- The API returns a non-200 or a meaningful error if cancellation could not be applied
- A cancelled run is distinguishable from a failed run in the UI

### How to reproduce

**For the 30+ minute hang (Bug 2):**
1. Configure a blueprint with a large git repository (thousands of commits)
2. Trigger the pipeline and wait for `collectCommits` / blame subtask to begin
3. Click cancel
4. Observe: "Operation successfully completed" in the UI, but pipeline status stays `RUNNING` for 30+ minutes

**For the silent cancel failure (Bug 1):**
1. Run a multi-stage pipeline
2. Click cancel immediately after one stage completes and before the next stage's tasks appear in progress
3. Observe: cancel returns 200 OK, next stage starts and runs to completion

### Anything else

Affected files:

| File | Issue |
|---|---|
| `server/services/pipeline.go:464` | `_ = CancelTask(...)` silently discards errors; unstarted tasks never marked cancelled |
| `plugins/gitextractor/parser/repo_gogit.go:526` | No `ctx.Done()` check in commit loop; `gogit.Blame()` has no context |
| `core/runner/run_task.go:91` | Context-cancelled tasks written as `TASK_FAILED` instead of `TASK_CANCELLED` |

Suggested fixes:
- Log or return errors from `CancelTask` instead of discarding them
- In `CancelPipeline` for a running pipeline: immediately set all `TASK_CREATED` tasks and the pipeline itself to `TASK_CANCELLED` in the DB
- In `storeRepoSnapshot`: add a `ctx.Done()` check at the top of the commit loop; investigate whether go-git exposes a context-aware blame API
- In `RunTask`: use `TASK_CANCELLED` when `errors.Is(err, context.Canceled)` is true

Related: #5585, #4188

### Version

b68c102f2

### Code of Conduct

danielemoraschi:
### ⚠️ Pre Checklist

### Summary
Clicking the cancel button in the UI (which fires DELETE /api/pipelines/:id) always returns "Operation successfully completed" (HTTP 200), but the pipeline continues running. Observable for 30+ minutes after the cancel request before it stops on its own.

Three independent bugs were causing this:

1. **`CancelPipeline` silently discarded errors from `CancelTask`** (`pipeline.go`). The cancel-func was never invoked for running tasks, and pending tasks were not batch-updated in the DB. Fixed by classifying tasks into running vs pending, then delegating to `cancelRunningTasks` (triggers context cancellation) and `cancelPendingTasksInDB` (batch DB update with status guard).

2. **`gitextractor` ignored context cancellation in `storeRepoSnapshot`** (`repo_gogit.go`). `gogit.Blame()` has no context parameter, so once a blame started it ran to completion. Added `select { case <-ctx.Done() }` checks at the top of the outer commit loop and inner `patch.Stats()` loop so cancellation is respected between iterations.

3. **Cancelled tasks were recorded as `TASK_FAILED`** (`run_task.go`, `pipeline_runner.go`). The deferred status-update block didn't distinguish cancellation from failure. Now detects cancellation via `errors.Is(err, context.Canceled) || ctx.Err() == context.Canceled`, sets status to `TASK_CANCELLED`, and skips writing `failed_sub_task`. `ComputePipelineStatus` also now returns `TASK_CANCELLED` when the pipeline was cancelled, taking priority over other statuses.

Additional correctness improvements:
- `cancelPendingTasksInDB` includes `TASK_RESUME` in the status filter so tasks awaiting restart recovery are also cancelled
- `cancelRunningTasks` uses `err.GetType() == errors.NotFound` instead of `.As(errors.NotFound)` to avoid walking the entire error chain
- `SkipOnFail` guard in `run_task.go` reuses the `isCancelled` variable instead of recomputing

Closes #8787

### Screenshots
N/A

### Other Information
Additional improvements:
- `cancelPendingTasksInDB` includes `TASK_RESUME` in the status filter so tasks awaiting restart recovery are also cancelled
- `cancelRunningTasks` uses `err.GetType() == errors.NotFound` instead of `.As(errors.NotFound)` to avoid walking the entire error chain
- `SkipOnFail` guard in `run_task.go` reuses the `isCancelled` variable instead of recomputing

Tests:
- Added cancellation tests to `TestComputePipelineStatus` verifies `TASK_CANCELLED` is returned when `isCancelled=true`, regardless of individual task statuses
- Added `TestCancelPipeline` with 3 subtests: cancels all pending tasks, leaves completed tasks unchanged, returns error for non-existent pipeline

### DIFF
--- a/backend/core/runner/run_task.go
+++ b/backend/core/runner/run_task.go
@@ -77,7 +77,12 @@ func RunTask(
 		}
 		finishedAt := time.Now()
 		spentSeconds := finishedAt.Unix() - beganAt.Unix()
+		isCancelled := errors.Is(err, gocontext.Canceled) || ctx.Err() == gocontext.Canceled
 		if err != nil {
+			taskStatus := models.TASK_FAILED
+			if isCancelled {
+				taskStatus = models.TASK_CANCELLED
+			}
 			lakeErr := errors.AsLakeErrorType(err)
 			subTaskName := "unknown"
 			if lakeErr = lakeErr.As(errors.SubtaskErr); lakeErr != nil {
@@ -87,16 +92,19 @@ func RunTask(
 			} else {
 				lakeErr = errors.Convert(err)
 			}
-			dbe := db.UpdateColumns(task, []dal.DalSet{
-				{ColumnName: "status", Value: models.TASK_FAILED},
+			columns := []dal.DalSet{
+				{ColumnName: "status", Value: taskStatus},
 				{ColumnName: "message", Value: lakeErr.Error()},
 				{ColumnName: "error_name", Value: lakeErr.Messages().Format()},
 				{ColumnName: "finished_at", Value: finishedAt},
 				{ColumnName: "spent_seconds", Value: spentSeconds},
-				{ColumnName: "failed_sub_task", Value: subTaskName},
-			})
+			}
+			if taskStatus == models.TASK_FAILED {
+				columns = append(columns, dal.DalSet{ColumnName: "failed_sub_task", Value: subTaskName})
+			}
+			dbe := db.UpdateColumns(task, columns)
 			if dbe != nil {
-				logger.Error(dbe, "failed to finalize task status into db (task failed)")
+				logger.Error(dbe, "failed to finalize task status into db (task %s)", taskStatus)
 			}
 		} else {
 			dbe := db.UpdateColumns(task, []dal.DalSet{
@@ -116,7 +124,7 @@ func RunTask(
 			dal.Where("id=?", task.PipelineId),
 		))
 		// not return err if the `SkipOnFail` is true and the error is not canceled
-		if dbPipeline.SkipOnFail && !errors.Is(err, gocontext.Canceled) {
+		if dbPipeline.SkipOnFail && !isCancelled {
 			err = nil
 		}
 	}()
--- a/backend/plugins/gitextractor/parser/repo_gogit.go
+++ b/backend/plugins/gitextractor/parser/repo_gogit.go
@@ -524,6 +524,11 @@ func (r *GogitRepoCollector) storeRepoSnapshot(subtaskCtx plugin.SubTaskContext,
 	ctx := subtaskCtx.GetContext()
 	snapshot := make(map[string][]string) // {"filePathAndName": ["line1 commit sha", "line2 commit sha"]}
 	for _, commit := range commitList {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
 		commitTree, firstParentTree, err := r.getCurrentAndParentTree(ctx, commit)
 		if err != nil {
 			return err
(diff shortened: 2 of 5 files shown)


klesh (approved):
LGTM
Thanks for your contribution.

PR #8779, opened 2026-03-17
fix: Make gh-copilot plugin database agnostic

sclausson:
### ⚠️ Pre Checklist

> Please complete _ALL_ items in this checklist, and remove before submitting

- [x] I have read through the [Contributing Documentation](https://devlake.apache.org/community/).
- [x] I have added relevant tests.
- [x] I have added relevant documentation.
- [x] I will add labels to the PR, such as `pr-type/bug-fix`, `pr-type/feature-development`, etc.

### Summary
Fixes issue where gh-pilot plugin prevents Devlake from starting when using Postgres.

### Does this close any open issues?
Closes #8778 

### Screenshots
N/A

### Other Information
Any other information that is important to this PR.

PR #8836, opened 2026-04-14
fix(asana): remove MySQL-specific gorm type tags for PostgreSQL compatibility

### LINKED ISSUE #8835: bug: Asana plugin migration uses MySQL-specific "double" type, fails on PostgreSQL
## Search before asking

- [x] I had mass searched in the [issues](https://github.com/apache/incubator-devlake/issues) and found no similar issues.

## What happened

The Asana plugin migration `20250212_add_task_transformation_fields` uses `gorm:"type:double"` on the `StoryPoint` field. PostgreSQL does not have a `double` type (it uses `double precision`), so the migration fails with:

```
ERROR: type "double" does not exist (SQLSTATE 42704)
```

The failing DDL:
```sql
ALTER TABLE "_tool_asana_tasks" ALTER COLUMN "story_point" TYPE double USING "story_point"::double
```

The migration should use a database-agnostic type. Removing the explicit `gorm:"type:double"` tag and letting GORM infer the type from `*float64` would work correctly on both MySQL and PostgreSQL.

## How to reproduce

1. Configure DevLake v1.0.3-beta10 with a PostgreSQL backend (`DB_URL=postgres://...`)
2. Trigger database migration via `/proceed-db-migration`
3. Migration fails on the Asana plugin step

## Anything else

The same class of bug was fixed for the copilot plugin's `datetime` type in #8779. This is the same pattern: a MySQL-specific type in a GORM struct tag that breaks PostgreSQL.

**Affected file:** `backend/plugins/asana/models/migrationscripts/20250212_add_task_transformation_fields.go:35`

```go
StoryPoint *float64 `gorm:"type:double"`
```

**Fix:** Remove the explicit type tag:
```go
StoryPoint *float64
```

GORM will infer `double precision` on PostgreSQL and `double` on MySQL.

**Workaround:** Create a PostgreSQL domain alias before running migrations:
```sql
CREATE DOMAIN double AS double precision;
```

## Version

v1.0.3-beta10

JAORMX:
## Summary

- Remove explicit `gorm:"type:double"` and `gorm:"type:int unsigned"` tags from the Asana task migration struct in `20250212_add_task_transformation_fields.go`
- PostgreSQL does not have a `double` type (uses `double precision`) and does not support `UNSIGNED` integer qualifiers — both tags cause migration failures on PostgreSQL
- Without explicit tags, GORM infers the correct column type per dialect automatically (`double`/`double precision` for `*float64`, `bigint unsigned`/`bigint` for `*uint`)

Follows the same pattern as #8779 which fixed the equivalent `datetime` issue in the copilot plugin.

closes #8835

## Test plan

- [ ] Verify `go build ./plugins/asana/...` passes (confirmed locally)
- [ ] Verify GORM AutoMigrate succeeds on PostgreSQL without errors
- [ ] Verify existing MySQL installations are unaffected (migration is gated by version number and GORM infers `double` / `bigint unsigned` for MySQL — matching existing column types)

### DIFF
--- a/backend/plugins/asana/models/migrationscripts/20250212_add_task_transformation_fields.go
+++ b/backend/plugins/asana/models/migrationscripts/20250212_add_task_transformation_fields.go
@@ -32,9 +32,9 @@ type asanaTask20250212 struct {
 	StdType         string   `gorm:"type:varchar(255)"`
 	StdStatus       string   `gorm:"type:varchar(255)"`
 	Priority        string   `gorm:"type:varchar(255)"`
-	StoryPoint      *float64 `gorm:"type:double"`
+	StoryPoint      *float64
 	Severity        string   `gorm:"type:varchar(255)"`
-	LeadTimeMinutes *uint    `gorm:"type:int unsigned"`
+	LeadTimeMinutes *uint
 }
 
 func (asanaTask20250212) TableName() string {


klesh:
@copilot resolve the merge conflicts in this pull request

klesh (approved):
LGTM.
Thanks for your contribution.

PR #8849, opened 2026-04-24
fix(sonarqube): use filter parameter for project search

### LINKED ISSUE #8805: [Bug][SonarQube] Search in Sonarqube connection seems wrong
### Search before asking

### What happened

The following behavior was found in v1.0.3-beta10 in a sonaqube connection, it might impact other searches too.

The search is not working properly or inuititevely. The maximal number of results returned seems to be limited to 20 without an option to load more results. Also when using the search field, it is unclear how the projects are mapped.

E.g. searching `backend` returns projects that have no word similiar to backend in the name, so results seems to be random. E.g. I have a project called `translation-backend`. Searching 'translation' does not show the project, searching 'backend' does. But this is not cohersive. So another time searching 'translation' returns my project amongst the others, but 'backend' not.

If there is a way to confidentially share a screenshot of this, please let me know, as I cannot share the list of projects unfortunately here.

- When searching for a pattern, all projects that contains this pattern are displayed.
- If the number of results are bigger than a threshold, there should be an option to page next results.
- An explanation on how to search for projects and what formats are accepted (like regex or simply word matching)

### How to reproduce

Create a sonarqube connection to a repo with more than 20 projects and start search for some projects

### Anything else

_No response_

### Version

v1.0.3-beta10

### Code of Conduct

jbsmith7741:
Summary
[Bug](https://github.com/apache/incubator-devlake/issues/8805) remote project search for SonarQube used the wrong query style for components/search_projects and a fixed small page; this PR uses SonarQube’s filter=query = "…" form, exact key matching where needed, and more reliable loading of additional pages in the data-scope UI. That aligns results with the connection’s projects and makes “load more” behave as users expect when there are many projects.

Does this close any open issues?
Fixes #8805

Screenshots
sonarQube_search

Other Information

### DIFF
--- a/backend/plugins/sonarqube/api/blueprint_v200.go
+++ b/backend/plugins/sonarqube/api/blueprint_v200.go
@@ -130,7 +130,9 @@ func GetApiProject(
 		Data []models.SonarqubeApiProject `json:"components"`
 	}
 	query := url.Values{}
-	query.Set("q", projectKey)
+	query.Set("p", "1")
+	query.Set("ps", "100")
+	query.Set("filter", sonarqubeSearchProjectsQueryFilter(projectKey))
 	// Use components/search_projects for consistency and normal-token (Browse) support.
 	res, err := apiClient.Get("components/search_projects", query, nil)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


klesh (approved):
LGTM
Thanks for your contribution.

PR #8850, opened 2026-04-24
fix: disable API client caching for GitHub App connections

### LINKED ISSUE #8847: [Bug][github] GitHub App installation token not refreshed for remote scopes / connection APIs, causing silent "zombie" connection after about 1 hour
### Search before asking

### What happened

When using **GitHub App authentication (**not PAT) on a GitHub connection, the connection silently becomes non-functional for administrative operations after approximately 1 hour, even when no requests have been made during that period.

**Observed symptoms:**

- The private GitHub organization no longer appears when trying to add a new scope through the Config UI.
- The API endpoint used to search/list repositories before creating a scope returns empty results, making it impossible to add new scopes.
- No error is logged in the devlake-lake pod — the failure is completely silent.
- Data collection on already-configured scopes continues to work normally during the same period. Blueprints run successfully, data is collected, pipelines complete without error.
- Editing the connection and re-submitting the same GitHub App private key restores the listing functionality for another ~1 hour.
- Restarting the devlake-lake pod (without re-editing the connection) also restores functionality for ~1 hour.

**Key discriminating observation:**

Once the connection is in the "zombie" state, REST data collection tasks keep succeeding on existing scopes while the remote scopes listing API returns empty. This decoupling strongly indicates that two different HTTP clients with independent token lifecycles are in use: one is being refreshed correctly, the other is not.

The GitHub App installation token should be transparently refreshed for all HTTP clients used by the GitHub plugin — including the ones backing the remote scopes API (org/repo listing) and the Test Connection endpoint — the same way it is already refreshed for REST data collection tasks since #8746.

The connection should remain fully usable indefinitely without requiring a pod restart or a manual re-edit of the private key.

### How to reproduce

1. Deploy DevLake v1.0.3-beta10 (Helm chart) on Kubernetes.
2. Create a GitHub connection using GitHub App authentication (App ID + installation ID + private key). Do not use a PAT.
3. Create a project and add at least one scope pointing to a private repository of the installed organization. Verify that data collection works.
4. Wait ~1 hour without making any request against the connection.
5. In the Config UI, attempt to add a new scope to the project: the organization no longer appears in the repository picker.
6. Via the API, attempt to search repositories through the remote scopes endpoint before creating a scope: the response is empty.
7. Trigger a blueprint on an existing scope: data collection still succeeds, proving the installation token itself is being refreshed somewhere in the process but not on the admin/listing code path.
8. Restart the devlake pod (no connection edit, no DB change): listing works again for another ~1 hour.

### Anything else

**Root cause hypothesis**
This appears to be the same architectural pattern as #8788, but on a different code path.

PR #8746 introduced NewAppInstallationTokenProvider + RefreshRoundTripper and wired them into CreateApiClient so that REST data collection tasks get transparent token refresh. Issue #8788 identified that the GraphQL client in backend/plugins/github_graphql/impl/impl.go was not updated as part of #8746 and still uses oauth2.StaticTokenSource, which freezes the token at task start time.

The present bug strongly suggests that the HTTP client(s) backing the remote scopes / connection admin APIs (org listing, repo search, test connection) suffer from the same issue: they were not recabled to the RefreshRoundTripper machinery in #8746. The installation token captured when the process starts (or when the connection is last edited) is cached in memory and never refreshed, so after the 1-hour GitHub installation token TTL expires, all admin API calls fail silently — likely because GitHub returns 401 Bad credentials and the listing code path interprets the empty/errored response as "no orgs available" without surfacing an error.

This would explain every observed characteristic:

- Why it is silent: the error is swallowed at the listing layer instead of being logged (unlike the GraphQL path in #8788 which panics loudly).
- Why collection still works: the REST path uses the client correctly refreshed by #8746.
- Why a pod restart fixes it temporarily: the in-memory frozen token is replaced by a freshly minted one when the process starts.
- Why re-editing the connection fixes it temporarily: it likely forces a new connection object to be built, including a new client with a fresh token.
- Why other environments with the same version may not see it: pods that are restarted more frequently (rolling deploys, aggressive liveness probes, autoscaling) never reach the 1-hour threshold in steady state.

### Version

 v1.0.3-beta10 

### Code of Conduct

fxgagnon:
GitHub App installation tokens expire after ~1 hour, causing admin APIs (remote-scopes, test-connection) to fail silently after token expiry.

This fix overrides GetHash() to return empty string for GitHub App connections, disabling client caching and ensuring fresh tokens are fetched for each admin request.

PAT connections continue to use default caching behavior.

Fixes #8847

### ⚠️ Pre Checklist

## Summary
Fixes GitHub App installation token expiration issue causing admin APIs to fail silently after 1 hour.

## Problem
GitHub App connections use installation tokens that expire after ~1 hour. Admin APIs (remote-scopes, test-connection) were reusing cached HTTP clients with expired tokens, causing silent failures.

## Root Cause
`DsRemoteApiProxyHelper.getApiClient()` caches HTTP clients using `connection.GetHash()`. Once cached, clients are reused indefinitely with their embedded tokens.

## Solution
Override `GetHash()` method in `GithubConnection` to return empty string for GitHub App connections, disabling client caching and ensuring fresh tokens are fetched for each admin request.

Follows the same pattern as Zentao plugin for connections with expiring credentials.

## Changes
- `backend/plugins/github/models/connection.go`: Added `GetHash()` method
- `backend/plugins/github/models/connection_test.go`: Added test for `GetHash()` logic

## Testing
- Unit tests pass for both GitHub App and PAT connections
- No impact on data collection (uses separate client creation path)

Fixes #8847

### DIFF
--- a/backend/plugins/github/models/connection.go
+++ b/backend/plugins/github/models/connection.go
@@ -134,6 +134,16 @@ func (connection GithubConnection) TableName() string {
 	return "_tool_github_connections"
 }
 
+func (connection GithubConnection) GetHash() string {
+	if connection.AuthMethod == AppKey {
+		// GitHub App installation tokens expire after ~1 hour, disable API client caching
+		// to ensure fresh tokens are fetched for admin APIs (remote-scopes, test-connection)
+		return ""
+	}
+	// Use default caching for PAT connections (they don't expire)
+	return connection.BaseConnection.GetHash()
+}
+
 func (connection *GithubConnection) MergeFromRequest(target *GithubConnection, body map[string]interface{}) error {
 	modifiedConnection := GithubConnection{}
 	if err := helper.DecodeMapStruct(body, &modifiedConnection, true); err != nil {
--- a/backend/plugins/github/models/connection_test.go
+++ b/backend/plugins/github/models/connection_test.go
@@ -18,6 +18,7 @@ limitations under the License.
 package models
 
 import (
+	"github.com/apache/incubator-devlake/core/models/common"
 	"github.com/apache/incubator-devlake/helpers/pluginhelper/api"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


klesh (approved):
LGTM.
Thanks for your contribution.

PR #8854, opened 2026-04-26
feat(auth): implement oidc authentication and session management

rbstp:
### Summary
This PR adds OIDC login behind a default-off `AUTH_ENABLED` switch. Providers are configured via `OIDC_PROVIDERS` plus per-name env blocks, so a deployment can offer Entra ID and Google (or any other OIDC IdP) side by side. Sessions are signed JWTs in HttpOnly cookies, backed by a new `auth_sessions` table so they can be revoked server-side without rotating `SESSION_SECRET`.

CSRF is enforced on state-changing requests when the caller is using the session cookie; API-key, Basic auth, and proxy-header callers are unaffected. Azure Workload Identity is supported for the Entra provider so AKS deployments can drop `CLIENT_SECRET` in favor of a federated SA-token assertion.

When `AUTH_ENABLED=false`, which is the default so no need to add it, the system behaves exactly as before.

`env.example` contains the variables needed to add the OIDC login.

RBAC was not implemented since there are no users roles currently, but it could be added in the future.

Refs: https://github.com/apache/incubator-devlake/discussions/8694
Refs: https://devlake-io.slack.com/archives/C03AJH1HBAT/p1776917914734679

### Screenshots
Example with Google, validated with Entra ID as well.
image
image
image
image

### Other Information
Validated with Entra ID (AAD) and Google Auth, but it should work with other providers as well, as long as they respect the OIDC spec.

Expect a 401 in DevTools on first load before the redirect to `/login`. It is the axios interceptor bouncing unauthenticated requests through the OIDC flow, not a bug.


klesh (approved):
Neat! This is a much-anticipated feature for the community. Thank you for the contribution!

vaujo6y:
There might be regression?

Our previous deployment we didn't have any authentication. Anyone can access devlake.

After upgrading to the latest release I now get

image

Env Vars

image

redeployed with `AUTH_ENABLED` and we get back the original functionality of no auth challenge. 

image

p.s nice feature.

rbstp:
@vaujo6y thanks for bringing up the issue, you are correct that it is a regression, it should have been disabled by default. I made a fix in https://github.com/apache/devlake/pull/8951

PR #8855, opened 2026-04-26
fix(github_graphql): keep requestedIssues out of GraphQL query in Collect Issues

### LINKED ISSUE #8853: [Bug][github_graphql] `Collect Issues` fails on second phase with `Field 'requestedIssues' doesn't exist on type 'Query'`
### Search before asking

### What happened

In `v1.0.3-beta12` the second phase of `github_graphql` `Collect Issues` (the stale-local-issue cleanup added in [#8820](https://github.com/apache/incubator-devlake/pull/8820)) aborts the subtask with `Field 'requestedIssues' doesn't exist on type 'Query'`:

```
[Collect Issues] ended Graphql collector execution with error
        Wraps: (2) Field 'requestedIssues' doesn't exist on type 'Query'
        Error types: (1) *hintdetail.withDetail (2) graphql.DataError
```

The unexported `requestedIssues` field added to `GraphqlQueryIssueDetailWrapper` in [#8820](https://github.com/apache/incubator-devlake/pull/8820) ([`issue_collector.go` L54-L62](https://github.com/apache/incubator-devlake/blob/v1.0.3-beta12/backend/plugins/github_graphql/tasks/issue_collector.go#L54-L62)) is intended as Go-side per-request metadata — populated before the request (L198) and read in the response callback (L221, L231) to detect issues that disappeared upstream. The `merico-ai/graphql` marshaller appears not to skip unexported fields, so `requestedIssues` is emitted as a top-level selection on `Query` and GitHub rejects it.

`Collect Issues` completes both phases when `_tool_github_issues` already has rows.

### How to reproduce

1. `Collect Data` once so `_tool_github_issues` has rows for the scope.
2. `Collect Data` again — phase 2 fails on the first request.

A scope with an empty `_tool_github_issues` skips phase 2 and does not surface the bug.

### Anything else

The right fix is to move `requestedIssues` out of `GraphqlQueryIssueDetailWrapper` and keep it on the closure or on `GithubTaskData`. `merico-ai/graphql`'s `writeQuery` ([`query.go` L111-L173](https://github.com/merico-ai/graphql/blob/master/query.go#L111-L173)) does not special-case `graphql:"-"` (the tag value would just be rendered as the field name `-`) and does not skip unexported fields, so there is no struct-tag-only escape hatch on the wrapper as written.

### Version

v1.0.3-beta12

### Code of Conduct

yamoyamoto:
### ⚠️ Pre Checklist

### Summary

Putting the per-batch metadata on the wrapper struct caused merico-ai/graphql to emit it as a top-level GraphQL field. Moved it off the wrapper into a `sync.Map` kept in the surrounding closure.

Closes #8853 

### Screenshots
N/A

### DIFF
--- a/backend/plugins/github_graphql/tasks/issue_collector.go
+++ b/backend/plugins/github_graphql/tasks/issue_collector.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/apache/incubator-devlake/core/dal"
@@ -52,8 +53,7 @@ type GraphqlQueryIssueWrapper struct {
 }
 
 type GraphqlQueryIssueDetailWrapper struct {
-	requestedIssues map[int]missingGithubIssueRef
-	RateLimit       struct {
+	RateLimit struct {
 		Cost int
 	}
 	Repository struct {
@@ -182,6 +182,7 @@ func CollectIssues(taskCtx plugin.SubTaskContext) errors.Error {
 		return err
 	}
 	issueUpdatedAt := make(map[int]time.Time)
+	requestedIssuesByQuery := sync.Map{}
 	err = apiCollector.InitGraphQLCollector(api.GraphqlCollectorArgs{
 		GraphqlClient: data.GraphqlClient,
 		Input:         iterator,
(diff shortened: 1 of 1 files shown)


klesh (approved):
Good catch!

PR #8861, opened 2026-04-28
fix(argoCD): avoid duplication of latest  deployments

antarasevich:
## Summary

Fix duplicate `ArgocdSyncOperation` records caused by ArgoCD exposing the same completed sync in two places in its API response: `status.history[]` and `status.operationState`.

## Problem

ArgoCD includes the most recent completed sync in both:
- `status.history[]` — the authoritative log of all past syncs, each with `deployedAt`
- `status.operationState` — a mirror of the last operation, with `startedAt`

The DevLake collector feeds both into the raw table. The extractor then processes each independently, using different timestamp fields as `DeploymentId`:

| Source | Field used | Example value | DeploymentId |
|---|---|---|---|
| `history[id=30]` | `deployedAt` | `2026-04-16T09:02:16Z` | `1776330136` |
| `operationState` | `startedAt` | `2026-04-16T09:02:15Z` | `1776330135` |

Because `startedAt` and `deployedAt` differ by one second (the sync duration), the two records get different primary keys and both are persisted — resulting in duplicate deployment entries in DORA metrics and deployment frequency charts.

## Fix

Extract the skip logic into a dedicated `isRedundantOperationState(phase string) bool` function and call it from inside `Extract` before processing any `operationState` row.

A `Succeeded` `operationState` is always already present in `status.history[]` so it is skipped. In-flight (`Running`, `Terminating`) and failure (`Failed`, `Error`) states are not yet — or may never be — written to history, so they continue to be processed normally.

```go
// isRedundantOperationState reports whether an operationState entry should be
// skipped because it duplicates the most recent history entry.
func isRedundantOperationState(phase string) bool {
    return phase == "Succeeded"
}
```

```go
// inside Extract:
isOperationState := apiOp.Phase != ""

if isOperationState && isRedundantOperationState(apiOp.Phase) {
    return nil, nil
}
```

## Behaviour by phase

| `operationState` phase | Already in `history[]`? | Action |
|---|---|---|
| `Succeeded` | yes | skip — history entry is authoritative |
| `Running` / `Terminating` | no | keep — only source for in-flight sync |
| `Failed` / `Error` | no | keep — failure may never reach history |
| ` ` (empty) | n/a — history entry, not operationState | keep — guard never fires |

## Testing

Added unit tests for `isRedundantOperationState` covering all phase variants. Verified against production ArgoCD API responses for multiple applications where `operationState.phase = Succeeded` — duplicate `ArgocdSyncOperation` records are eliminated after the fix.

## Related

- Affects: `backend/plugins/argocd/tasks/sync_operation_extractor.go`
- Reproduces with: any ArgoCD application whose last sync completed successfully (i.e. the vast majority of applications in a healthy cluster)

### DIFF
--- a/backend/plugins/argocd/tasks/sync_operation_extractor.go
+++ b/backend/plugins/argocd/tasks/sync_operation_extractor.go
@@ -188,6 +188,13 @@ func ExtractSyncOperations(taskCtx plugin.SubTaskContext) errors.Error {
 
 			isOperationState := apiOp.Phase != ""
 
+			// *** THE FIX ***
+			// operationState always represents the same sync as the latest history entry
+			// (when phase=Succeeded). Skip it to avoid the duplicate DeploymentId.
+			if isOperationState && apiOp.Phase == "Succeeded" {
+				return nil, nil
+			}
+
 			// For multi-source apps ArgoCD sets revisions[] instead of revision. Resolve
 			// the single commit SHA we care about before deciding whether to skip this entry.
 			if apiOp.Revision == "" {
--- a/backend/plugins/argocd/tasks/sync_operation_extractor.go
+++ b/backend/plugins/argocd/tasks/sync_operation_extractor.go
@@ -191,7 +191,7 @@ func ExtractSyncOperations(taskCtx plugin.SubTaskContext) errors.Error {
 			// *** THE FIX ***
 			// operationState always represents the same sync as the latest history entry
 			// (when phase=Succeeded). Skip it to avoid the duplicate DeploymentId.
-			if isOperationState && apiOp.Phase == "Succeeded" {
+			if isOperationState && isRedundantOperationState(apiOp.Phase) {
 				return nil, nil
 			}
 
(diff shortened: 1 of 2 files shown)


backend/plugins/argocd/tasks/sync_operation_extractor.go
```diff
@@ -188,6 +188,13 @@ func ExtractSyncOperations(taskCtx plugin.SubTaskContext) errors.Error {
 
 			isOperationState := apiOp.Phase != ""
 
+			// *** THE FIX ***
```
rbstp, line 191:
```suggestion
```

rbstp (changes requested):

backend/plugins/argocd/tasks/sync_operation_extractor.go
rbstp: Thanks for the contribution! Can you please remove the above comment, and fix the failing test.

antarasevich:
invalid. To be updated

PR #8880, opened 2026-05-16
fix: require secret for forwarded user auth

klesh:
### Summary
Reject spoofed X-Forwarded-User headers unless a configured shared secret matches the forwarded secret header.


Startrekzky (approved):
LGTM

PR #8890, opened 2026-05-28
fix(dora): count failed deployments, not incidents, in Change Failure Rate

spiffaz:
## What does this PR do?

The Change Failure Rate dashboard (`grafana/dashboards/DORADetails-ChangeFailureRate.json`) computes the rate as `sum(has_incident) / count(deployment_id)`, where `has_incident` is `count(distinct i.id)` per deployment, i.e. the number of incidents a deployment caused.

When one deployment causes more than one incident, it is counted multiple times in the numerator, so the rate exceeds the DORA definition of Change Failure Rate (the share of deployments that caused a failure, not the number of incidents).

This makes `has_incident` a 0/1 flag (`CASE WHEN count(distinct i.id) > 0 THEN 1 ELSE 0 END`) so each failed deployment counts once, and relabels the companion stat panel field from "incident count" to "failed deployment count" so the displayed counts stay consistent with the rate.

## Validation

Against a production dataset, a 30-day window had 6 incidents spread across 4 distinct deployments out of 23 total:
- Before: 6 / 23 = 26.1%
- After: 4 / 23 = 17.4%

Only the SQL in the two CFR panels changed; no datasource, layout, or variable changes.


klesh:
@Startrekzky Please take a look when you find time, thx.

Startrekzky (approved):
LGTM. Should be counting the number of failed deployments

PR #8892, opened 2026-05-29
Allow Rootly plugin to collect incidents without service identifiers assigned

### LINKED ISSUE #8891: [Bug][Rootly] Incidents without service tags are silently dropped during collection
### Search before asking

### What happened

Description:

The Rootly plugin collects incidents by filtering per-service using filter[service_ids]. However, in practice many (or all) Rootly incidents are created without service associations. This causes three compounding issues that silently discard valid incidents:

Bug 1 — Collector always sends service filter
buildIncidentsQuery always appends filter[service_ids]=<serviceId> to every API request. The Rootly API returns 0 results for incidents with no service tag, so untagged incidents are never stored in the raw table.

Bug 2 — ValidateTaskOptions rejects empty ServiceId
There is no way to trigger a global (unfiltered) collection because an empty serviceId returns a validation error: "not enough info for Rootly execution".

Bug 3 — Extractor service-membership guard drops incidents when ServiceId is empty
In extractRootlyIncident, there is a safety check that filters out any incident whose service list doesn't contain the current ServiceId. When ServiceId == "", containsServiceId always returns false, so any incident that does have service relationships attached gets dropped as a false positive.

Impact:
 - Any incident not tagged with a specific service is permanently lost — never reaches the tool layer or domain layer

Allow ServiceId to be empty (opt-in global collection)
Skip filter[service_ids] in buildIncidentsQuery when ServiceId is empty
Skip the extractor service-membership guard when collecting globally

### How to reproduce

 1. Create a Rootly incident without associating it to any service
 2. Run the Rootly collector for any service scope
 3. The incident does not appear in _tool_rootly_incidents or the incidents domain table

### Anything else

_No response_

### Version

v1.0.3-beta12

### Code of Conduct

brianfeucht:
### ⚠️ Pre Checklist

- [X] I have read through the [Contributing Documentation](https://devlake.apache.org/community/).
- [X] I have added relevant tests.
- [X] I have added relevant documentation.
- [X] I will add labels to the PR, such as `pr-type/bug-fix`, `pr-type/feature-development`, etc.

### Summary
What does this PR do?
Allows the Rootly plugin to collect incidents missing service identifiers

Closes [8891](https://github.com/apache/incubator-devlake/issues/8891)

### DIFF
--- a/backend/plugins/rootly/tasks/incidents_collector.go
+++ b/backend/plugins/rootly/tasks/incidents_collector.go
@@ -126,7 +126,9 @@ func CollectIncidents(taskCtx plugin.SubTaskContext) errors.Error {
 // `filter[service_ids]`) is caught by a unit test.
 func buildIncidentsQuery(serviceId string, pageSize, pageNumber int, createdAfter *time.Time) url.Values {
 	query := url.Values{}
-	query.Set("filter[service_ids]", serviceId)
+	if serviceId != "" {
+		query.Set("filter[service_ids]", serviceId)
+	}
 	query.Set("page[size]", fmt.Sprintf("%d", pageSize))
 	// Rootly's JSON:API pagination is 1-based.
 	query.Set("page[number]", fmt.Sprintf("%d", pageNumber))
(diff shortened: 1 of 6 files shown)


klesh (approved):
LGTM.
Thanks for your contribution.

PR #8893, opened 2026-05-30
feat: require authentication for /proceed-db-migration endpoint if auth was enabled

klesh:
### Summary
This PR updates the /proceed-db-migration API endpoint to require authentication if authentication is enabled. Previously, this endpoint was accessible without authentication, which could pose a security risk in protected environments.

### Changes
- Added the auth.RequireAuth() middleware to the /proceed-db-migration route in the API server.
- Removed /proceed-db-migration from the list of public paths in the authentication middleware.
- Updated test helper comments to note that credentials are now required for this endpoint when AUTH_ENABLED=true.

### Motivation
Securing the database migration endpoint ensures that only authorized users can trigger migrations, aligning with best practices for production deployments.


Startrekzky (approved):
LGTM

PR #8894, opened 2026-05-30
fix(github): create domain accounts for non-committer authors (#8886)

### LINKED ISSUE #8886: GitHub plugin: domain `accounts` row missing for non-contributor authors, leaving `issues.creator_id` / `pull_requests.author_id` as orphan FKs
The GitHub plugin writes `issues.creator_id` and `pull_requests.author_id` for every author it sees, but only creates the corresponding `accounts` row when that user has also committed code to the tracked repo. Anyone who only filed an issue, or who opened a PR without commits to the repo, ends up with a foreign key that points to nothing. A `LEFT JOIN accounts ON issues.creator_id = accounts.id` returns NULL even though the login is sitting one column over in `issues.creator_name`.

### Reproduction

1. Configure the GitHub plugin against a public repo with external issue traffic. `stacklok/toolhive` is a convenient one.
2. Pick an issue filed by someone with no commits in that repo. In `stacklok/toolhive`, `#4297` (filed by `milichev`), `#4301`, `#4359`, and `#4413` all qualify.
3. After a full sync, query the domain layer:

```sql
SELECT i.url, i.creator_id, i.creator_name, a.user_name
  FROM issues i
  LEFT JOIN accounts a ON a.id = i.creator_id
 WHERE i.url = 'https://github.com/stacklok/toolhive/issues/4297';
```

`creator_id` is populated, `creator_name` is `milichev`, and `a.user_name` is NULL. Expected: `a.user_name` is `milichev`. The same shape affects `pull_requests.author_id` for PRs from non-committer authors.

### Root cause

`backend/plugins/github/tasks/issue_extractor.go` emits a `GithubRepoAccount` for `body.User` via `convertAccount(...)` but does not emit a `GithubAccount` row. `_tool_github_accounts` is populated separately by `account_collector.go`, which only iterates `/repos/{owner}/{repo}/contributors`. Users who never committed never land in `_tool_github_accounts`.

`account_convertor.go::ConvertAccounts` converts only what is in `_tool_github_accounts`:

```go
dal.Select("_tool_github_accounts.*"),
dal.From(&models.GithubAccount{}),
dal.Where("repo_github_id = ? and _tool_github_accounts.connection_id=?", ...),
dal.Join(`left join _tool_github_repo_accounts gra on (...)`),
```

The `LEFT JOIN _tool_github_repo_accounts` with a `WHERE` on the right-table column behaves like an inner join, so users present only in `_tool_github_repo_accounts` (the rows `issue_extractor` inserts via `convertAccount`) get filtered out.

`issue_convertor.go::ConvertIssues` writes `domainIssue.CreatorId = accountIdGen.Generate(connId, issue.AuthorId)` unconditionally for any non-zero AuthorId, producing the orphan FK. `pr_extractor.go` has the same pattern, with the same outcome for PR authors and merged-by users.

The denormalized columns `issues.creator_name` and `pull_requests.author_name` are populated correctly from the API response, so the login is recoverable. Any consumer that joins to `accounts` for author attribution will silently under-attribute external contributors. Bot filters that key on `accounts.user_name LIKE '%[bot]'` will miss bot-filed issues for the same reason.

Reproduced on `v1.0.3-beta10`. The same code paths exist on `main`; the recent refactor of `ConvertAccounts` from `NewDataConverter` to `NewStatefulDataConverter` did not change the filter. Likely present in every prior release.

### Suggested fix

Have the issue and PR extractors emit a `GithubAccount` alongside the `GithubRepoAccount`. In `issue_extractor.go`, where `body.User` is processed:

```go
if body.User != nil {
    githubIssue.AuthorId = body.User.Id
    githubIssue.AuthorName = body.User.Login
    repoUser, err := convertAccount(body.User, data.Options.GithubId, data.Options.ConnectionId)
    if err != nil { return nil, err }
    results = append(results, repoUser)

    results = append(results, &models.GithubAccount{
        ConnectionId: data.Options.ConnectionId,
        Id:           body.User.Id,
        Login:        body.User.Login,
    })
}
```

Same change in `pr_extractor.go` for `body.User` and `body.MergedBy`. The upsert into `_tool_github_accounts` is idempotent: if the user later shows up as a contributor, `account_extractor` overwrites with the richer record. The alternative is to rewrite the `ConvertAccounts` query to UNION in author IDs from `_tool_github_issues` and `_tool_github_pull_requests`, which is larger and changes the contract of that pipeline.

### Workaround for consumers

`COALESCE(NULLIF(a.user_name, ''), i.creator_name)` for issues, or `pr.author_name` for PRs. The denormalized columns are populated before the orphan-FK chain runs, so the login is always there.

JAORMX:
## Problem

`issues.creator_id` and `pull_requests.author_id` are written for every author, but a domain `accounts` row is only created for users we collected a full profile for (effectively, committers). Authors who only filed an issue, or opened a PR without committing, become orphan foreign keys:

```sql
SELECT i.url, i.creator_id, i.creator_name, a.user_name
  FROM issues i
  LEFT JOIN accounts a ON a.id = i.creator_id
 WHERE i.url = 'https://github.com/stacklok/toolhive/issues/4297';
```

`creator_id` is set and `creator_name` is `milichev`, but `a.user_name` is NULL. Bot filters keying on `accounts.user_name LIKE '%[bot]'` miss those rows for the same reason.

Fixes #8886.

## Root cause

The issue and PR extractors already record every author in `_tool_github_repo_accounts`. But `ConvertAccounts` sourced the domain `accounts` table `FROM _tool_github_accounts`, which is only populated for users we fetched full profiles for. So the convertors generate `creator_id` / `author_id` for everyone, while `accounts` only ever gets the committers.

## Change

- `ConvertAccounts` now reads `FROM _tool_github_repo_accounts LEFT JOIN _tool_github_accounts`, so every user the repo references becomes a domain account: enriched with name/email/avatar when we have the detail, login-only otherwise. The domain id uses the same `didgen` generator the issue/PR convertors use, so the foreign keys line up.
- `pr_extractor` also emits a repo_account for the `merged_by` user, which fixes `pull_requests.merged_by_id` (that user wasn't recorded anywhere before).
- The query is MySQL/PostgreSQL-agnostic (COALESCE not IFNULL, no backtick quoting, values parameterized via the dal). The join mirrors the existing one in `account_org_collector.go`.

## Testing

- The e2e fixture didn't exercise the bug (every referenced account already had a profile row), so I added the orphan case from the issue: `milichev`, referenced by the repo with no detail row. It now gets a login-only `accounts` row.
- Added a referential-integrity assertion to `TestAccountDataFlow`: every account the repo references must resolve to a domain `accounts` row, generated with the same id generator the convertors use. Verified it fails against the old converter and passes with the fix.
- Full `plugins/github/e2e` suite passes on both MySQL and PostgreSQL.

## Out of scope

Plural issue/PR assignees and PR requested reviewers aren't seeded into `_tool_github_repo_accounts`, so those FKs can still be unresolved. The `github_graphql` plugin likely shares the root cause. Happy to follow up on these separately.


backend/plugins/github/e2e/snapshot_tables/account.csv
```diff
@@ -1,8 +1,9 @@
...
-github:GithubAccount:1:3971390,cnliuyunpeng@gmail.com,ppmoon,ppmoon,https://avatars.githubusercontent.com/u/3971390?v=4,,"{""ConnectionId"":1,""Name"":""panjf2000/ants""}",_raw_github_api_accounts,14,
-github:GithubAccount:1:7496278,i@andypan.me,Andy Pan,panjf2000,https://avatars.githubusercontent.com/u/7496278?v=4,,"{""ConnectionId"":1,""Name"":""panjf2000/ants""}",_raw_github_api_accounts,5,
-github:GithubAccount:1:8518239,badger@gitter.im,The Gitter Badger,gitter-badger,https://avatars.githubusercontent.com/u/8518239?v=4,,"{""ConnectionId"":1,""Name"":""panjf2000/ants""}",_raw_github_api_accounts,9,
-github:GithubAccount:1:964542,sarath.sp06@gmail.com,Sarath Sadasivan Pillai,sarathsp06,https://avatars.githubusercontent.com/u/964542?v=4,"exotel,leadmrktr,shellagilehub,odysseyhack,boodltech","{""ConnectionId"":1,""Name"":""panjf2000/ants""}",_raw_github_api_accounts,1,
+github:GithubAccount:1:1052632,runner.mei@,runner,runner-mei,https://avatars.githubusercontent.com/u/1052632?v=4,,,,0,
```
klesh: _raw_data fields are missing, please fix it. Thanks.

JAORMX: Oops, will do when I'm back on my computer

JAORMX: Done in b67aad764! The converter now carries the _raw_data fields over: it prefers the enriched _tool_github_accounts row when we collected a profile, and falls back to the _tool_github_repo_accounts row for the non-committer case (that's why milichev shows _raw_github_api_issues provenance). The e2e fixture was missing the _raw_data columns too, so I added those and regenerated the snapshot... the enriched rows now carry exactly the same provenance as before this PR.

While at it, I also guarded pr_convertor.go against generating account ids for a zero AuthorId/MergedById. An unmerged PR would otherwise get merged_by_id pointing at github:GithubAccount:1:0, which is the same orphan shape this PR is fixing (issue_convertor already had that guard).

Verified the full github e2e suite on both MySQL and PostgreSQL.

klesh (approved):
LGTM.
Thanks for our contribution.

PR #8918, opened 2026-06-12
fix(ui): guard against empty plugin prop on connection creation page

### LINKED ISSUE #8914: [Bug][UI] Few connection creation page is broken
### Search before asking

### What happened

With the latest main branch code running in development environment, i am facing below error for few Data connection creation page.

Getting error as 

`TypeError: Cannot read properties of undefined (reading 'name')`

please refer to the video attached below

https://drive.google.com/file/d/15bjxQ8DcJScVCHw2g4cLbqsJySj-rr8V/view?usp=sharing

The Relevent model should be opened in order to configure the connection details.

### How to reproduce

Take the latest main branch and run it and try to create connection

### Anything else

I would like to contribute the fix for this.

### Version

latest main branch

### Code of Conduct

bujjibabukatta:
## Summary

Fixes #8914

## What happened

When clicking "New Connection" from the connection list modal, the `plugin`
state was still an empty string `''` when `ConnectionForm` rendered. This
caused `getPluginConfig('')` to return a fallback config with an empty `name`,
leading to:

TypeError: Cannot read properties of undefined (reading 'name')

## Root Cause

In `connections.tsx`, `handleShowFormDialog` set `type = 'form'` but did
not ensure `plugin` was set. The form modal only checked `pluginConfig` but
not `plugin` itself, so `ConnectionForm` received an empty string as the
`plugin` prop.

## Changes

**`config-ui/src/routes/connection/connections.tsx`**
- Added `plugin &&` guard to the form modal condition to prevent
  `ConnectionForm` from rendering with an empty plugin

**`config-ui/src/plugins/components/connection-form/index.tsx`**
- Added `?? {}` fallback on `getPluginConfig` destructuring
- Added `if (!plugin || !name) return null` guard after all hooks
  as a defensive measure

## How to Test

1. Take the latest main branch
2. Navigate to the Connections page
3. Click on any data source plugin
4. Click "New Connection" in the list modal
5. The connection creation form should open correctly without errors

### DIFF
--- a/config-ui/src/plugins/components/connection-form/index.tsx
+++ b/config-ui/src/plugins/components/connection-form/index.tsx
@@ -110,13 +110,17 @@ export const ConnectionForm = ({ plugin, connectionId, onSuccess }: Props) => {
   const {
     name,
     connection: { docLink, fields, initialValues },
-  } = getPluginConfig(plugin);
+  } = getPluginConfig(plugin) ?? {};
 
   const disabled = useMemo(() => {
     return Object.values(errors).some(Boolean);
   }, [errors]);
 
   const sanitizedCustomHeaders = useMemo(() => sanitizeCustomHeaders(values.customHeaders), [values.customHeaders]);
+
+  if (!plugin || !name) {
+    return null;
+  }
 
   const handleTest = async () => {
     const isUpdate = type === 'update' && !!connectionId;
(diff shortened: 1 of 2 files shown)


veetmoradiya3628 (approved):
Looks good to me

klesh (approved):

PR #8923, opened 2026-06-13
fix(jira): collect board issues via saved filter JQL

bujjibabukatta:
Fixes Jira Kanban boards silently missing issues when a board sub-filter is configured.

Previously, DevLake collected board issues via Jira Agile’s board issue endpoint, which applies the board saved filter and Kanban subQuery server-side. This caused resolved/released issues to disappear from DevLake when the subQuery excluded them.

This change collects Jira board issues through the Jira Search API using the saved board filter JQL stored on _tool_jira_boards.jql, bypassing the board-level subQuery while preserving the existing board issue association during extraction. Jira Server uses api/2/search, while Jira Cloud uses api/3/search/jql with token pagination.

Also stores the board configuration subQuery on _tool_jira_boards.sub_query for visibility/debugging.


klesh:
I don't think this is a good idea.

PR #8928, opened 2026-06-15
fix(gitlab): use diff_stats instead of changes_count for correct MR s…

### LINKED ISSUE #8888: [Bug][Gitlab] PR Size Incorrectly include previous PR
### Search before asking

### What happened

Image

PR with multiple part shows wrong size calculation, it should be 6523 not 9292

Image

It should calculate PR size based on specific PR from Gitlab 6523

### How to reproduce

Do multi-part PR on a single branch, the final pr size on latest PR is the total changes from previous PR.

Each PR should have separate size calculation not summation from previous PR. 

### Anything else

i'm running apache/devlake:latest
bae4e753959584b152fd2d66b2b13c4bc6c5ce02b9bd7c4a000e3e1296b09343

v1.0.2@41c047d

### Version

bae4e753959584b152fd2d66b2b13c4bc6c5ce02b9bd7c4a000e3e1296b09343

### Code of Conduct

bujjibabukatta:
## Problem
GitLab MR size is incorrectly inflated when a branch has multiple 
sequential PRs. The `changes_count` field from GitLab's API returns 
cumulative line changes across all commits on the branch including 
previous PRs, not just the current MR's diff.

## Root Cause
GitLab's `changes_count` sums all commits on the source branch. 
For multi-part PRs this means PR #3 shows 9292 lines changed 
instead of the actual 6523 lines in that PR alone.

## Fix
Use GitLab's `diff_stats` field (via `include_diff_stats=true` 
query param) which returns additions/deletions relative to the 
merge base — only what this PR itself introduces.

## Changes
- `backend/plugins/gitlab/tasks/mr_extractor.go` — added `DiffStats` 
  struct to `MergeRequestRes` and mapped `Additions`/`Deletions` 
  in `convertMergeRequest`
- `backend/plugins/gitlab/models/mr.go` — added `Additions` and 
  `Deletions` fields to `GitlabMergeRequest`
- `backend/plugins/gitlab/tasks/shared.go` — added 
  `include_diff_stats=true` query param to `GetQuery`
- `backend/plugins/gitlab/models/migrationscripts/` — added migration 
  script and registered it

Fixes #8888


backend/plugins/gitlab/tasks/mr_detail_collector.go
```diff
@@ -62,6 +62,7 @@ func CollectApiMergeRequestDetails(taskCtx plugin.SubTaskContext) errors.Error {
 		Query: func(reqData *helper.RequestData) (url.Values, errors.Error) {
 			query := url.Values{}
 			query.Set("with_stats", "true")
+			query.Set("include_diff_stats", "true")
```
klesh, line 65: When was it added to the GitLab API? We need to make sure this won't break older GitLab on-premise instances.

bujjibabukatta: Hey @klesh, good catch!
Looked it up — `diff_stats` has been around since GitLab 13.0 (June 2020), so most instances should be fine. For anything older, I've made `DiffStats` a pointer so we can tell when it's missing and fall back to `changes_count` instead of storing zeros. And `include_diff_stats=true` in the query is harmless on old versions — they just ignore params they don't know.

klesh (dismissed):
LGTM.
Thanks for your contribution.

bujjibabukatta:
Hi @klesh — I pushed 8ad6c93 to fix the build (corrected a } else if syntax error, added the missing strconv import, and added the Apache license header on the migration script). My earlier approval was auto-dismissed by the new commit. The CI checks don't seem to have triggered on this push — could you approve the workflow run and re-review when you get a chance? Thanks a lot for your patience!

klesh (approved):
LGTM
Thanks for your contribution.

PR #8927, opened 2026-06-15
feat(plugin): add CheckmarxOne data source plugin (#8869)

bujjibabukatta:
## Summary

This PR implements a complete CheckmarxOne data source plugin for Apache DevLake (resolves #8869).

The plugin enables security teams to integrate CheckmarxOne SAST scanning results into DevLake pipelines, allowing them to:
- Track and follow security issues discovered by CheckmarxOne
- Provide KPIs around findings, triage, and remediation
- Monitor vulnerability severity and remediation progress
- Support DevSecOps practices with security-focused metrics

## Motivation

With the rising importance of DevSecOps, organizations need to track security vulnerabilities alongside development metrics. This plugin provides seamless integration with CheckmarxOne, a leading code security scanning platform, enabling teams to establish meaningful performance indicators around security vulnerability remediation.

### Plugin Structure
- **Entry Point**: `checkmarxone.go` - Main plugin entry point with CLI support
- **Implementation**: `impl/impl.go` - Core plugin logic implementing all required interfaces
- **Models**: Connection, Project, and Finding data models
- **API Client**: OAuth-based authentication and API communication
- **Tasks**: Collector and extractor for findings data
- **API Endpoints**: REST endpoints for connection management
- **Migrations**: Database schema setup

#### CheckmarxoneConnection
- Server URL, Client ID/Secret for OAuth
- Secure credential storage (passwords hidden in JSON)

#### CheckmarxoneProject
- Project ID and metadata
- Connection association

#### CheckmarxoneFinding
- Vulnerability data: ID, name, severity, status
- First/Last found timestamps
- Finding type and count

### Features

✅ OAuth authentication with automatic token refresh
✅ Multi-project support with separate data collection
✅ Vulnerability severity tracking (Critical, High, Medium, Low)
✅ Remediation status monitoring
✅ REST API for connection CRUD operations
✅ Automatic database migrations
✅ Comprehensive error handling and logging
✅ Resource cleanup and connection management

PR #8929, opened 2026-06-15
fix(dora): ignore seed deployments in change lead time lookup

### LINKED ISSUE #8790: [Bug][DORA-change_lead_time_calculator] Wrong deployment reference
### Search before asking

### What happened

Hi so I did do a search to see if there were any issues similar to what I was facing and found the following:

[ #8188 ](https://github.com/apache/incubator-devlake/issues/8188)
[ #8249 ](https://github.com/apache/incubator-devlake/issues/8249)

and related 
[ #7193 ](https://github.com/apache/incubator-devlake/issues/7193)

I was wondering if these issues ever got resolved? I can see that there were two related pull requests but they were closed. 

https://github.com/apache/incubator-devlake/pull/8206
https://github.com/apache/incubator-devlake/pull/8552

This issue is first set of deployment commits for a project are being wrongly associated to old MRs leading to a massively skewed Median lead time for change. 

We had previously got around this using webhook deployments, the single registration of a deployment using webhooks did not lead to the behavior described above. 

However we are looking to transition to use the Gitlab connector plugin and annoyingly we are finding the very first merge requests of the project from over a year ago are being associated to the first recorded deployments in devlake.

It would be good to get an update on the above issues which seemed to stale out. 

We are using v1.0.2-beta6 and are looking to upgrade to the latest, however not sure if this issue was ever addressed? Please correct me if I am wrong. 

From my research I expect that the first deployment is a seed commit deployment in which there is no lead time calculation made. 

Then for the subsequent deployment the median lead time for change is calculated for MR commit hashes between the first deployment and the second deployment and so forth. 

### How to reproduce

**Reproduce:** 

1) Onboard a project with historical merge request data but without historical deployment data
2) Generate a number of new merge requests with associated deployments
3) Run the DORA pipeline on Devlake
4) View skewed Median lead time for change results 

### Anything else

This does not happen when registering deployments via webhooks one at a time. 

It does happen when using the Gitlab connector and there are already a group of deployments to be extracted for the first time. 

I have tried using a webhook deployment as a 'seed deploy commit' then followed by configuring the gitlab connector but I get the same behavior. 

### Version

v1.0.2-beta6

### Code of Conduct

bujjibabukatta:
Fixes #8790.

This updates the DORA change lead time deployment lookup so seed deployments are not used as valid deployment references for historical merge requests. The query now requires the deployment to have an actual previous deployment commit and joins `commits_diffs` against that previous commit SHA, instead of treating an empty previous commit as a valid range boundary.

This prevents the first recorded deployment for a project from being associated with old MRs when historical MR data exists but historical deployment data does not.

### DIFF
--- a/backend/plugins/checkmarxone/README.md
+++ b/backend/plugins/checkmarxone/README.md
@@ -1,0 +1,91 @@
+# CheckmarxOne Plugin
+
+## Summary
+
+This plugin collects security findings and vulnerabilities from [CheckmarxOne](https://checkmarx.com/checkmarx-one/) - a leading application security testing platform.
+
+## Features
+
+- Collect security findings/vulnerabilities from CheckmarxOne projects
+- Track vulnerability severity, status, and remediation progress
+- Support for multiple projects
+- Integration with DevLake's security domain layer
+
+## Requirements
+
+- CheckmarxOne account and API access
+- Server URL, Client ID, and Client Secret from CheckmarxOne
+
+## Configuration
+
+### Connection Setup
+
+Create a connection to CheckmarxOne using the following fields:
+
+- **Server URL**: The base URL of your CheckmarxOne instance (e.g., `https://checkmarx.mycompany.com`)
+- **Client ID**: OAuth client ID for API access
+- **Client Secret**: OAuth client secret for API access
+- **Username**: (Optional) Username for authentication
+- **Password**: (Optional) Password for authentication
+
+### Scope Configuration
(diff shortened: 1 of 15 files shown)


klesh:
Please, check the `Files changed` after you submitted the PR.
It contains code from #8927.
This is very confusing for me.

VaughanJ:
Why closed? If this fixes the Dora bug it will be great.

PR #8931, opened 2026-06-15
fix(jira): collect issues via search API to avoid kanban sub-filter exclusion

### LINKED ISSUE #8917: [Bug][Jira] Kanban board sub-filter silently excludes issues from collection
### Search before asking

### What happened

When a Jira kanban board is configured as a DevLake scope, resolved and completed issues can silently go missing from the collected data. Specifically:

- Issue counts in DevLake are significantly lower than what Jira reports for the same board filter
- DORA and cycle time metrics are underreported for kanban teams
- Resolved issues disappear from DevLake after being assigned a released `fixVersion` in Jira
- No errors or warnings are produced — syncs complete successfully with incomplete data

The root cause is that Jira kanban boards support a **sub-filter query** (`subQuery`) that is separate from the board's saved filter JQL. This sub-filter is configured in the board's settings and is applied server-side by Jira's Agile API before returning results. A common sub-filter is:

```
fixVersion in unreleasedVersions() OR fixVersion is EMPTY
```

DevLake collects board issues via `GET /rest/agile/1.0/board/{boardId}/issue` (`issue_collector.go`). Jira applies both the saved filter JQL **and** the sub-filter server-side, so DevLake receives only the filtered subset with no way to detect that issues were excluded.

The board configuration endpoint (`GET /rest/agile/1.0/board/{boardId}/configuration`) returns the sub-filter in a `subQuery` field. DevLake already calls this endpoint in `board_filter_begin_collector.go` to fetch the saved filter, but the `subQuery` field is not parsed into the `BoardConfiguration` struct — it is silently dropped.

```
Jira board config
├── filter.id → saved filter JQL (fetched and stored in _tool_jira_boards.jql)
└── subQuery  → sub-filter query (not parsed, not stored)

Collection: GET agile/1.0/board/{boardId}/issue
  → Jira applies saved filter JQL + subQuery server-side
  → DevLake receives only the filtered subset
  → _raw_jira_api_issues / _tool_jira_board_issues
```

This affects only kanban boards. Scrum boards are not affected because they do not use the sub-filter mechanism in the same way.

DevLake should collect all issues matching the board's saved filter JQL, regardless of any kanban sub-filter. At minimum, DevLake should parse and store the `subQuery` field so users can see when a sub-filter is active and understand why issue counts may differ from Jira.

### How to reproduce

1. Create or identify a Jira kanban board that has a sub-filter configured (Board Settings > General > Sub-filter Query), such as `fixVersion in unreleasedVersions() OR fixVersion is EMPTY`
2. Ensure the board's saved filter JQL matches issues that have a released `fixVersion`
3. Configure the board as a DevLake scope and run a full sync
4. Compare the issue count in `_tool_jira_board_issues` for that board against the count returned by running the board's saved filter JQL directly via the Jira search API (`/rest/api/2/search?jql=...`)
5. The DevLake count will be lower — issues with a released `fixVersion` are missing

You can confirm the sub-filter is present by checking the board configuration API:

```
GET /rest/agile/1.0/board/{boardId}/configuration
```

The response will include a `subQuery` field if a sub-filter is configured.

### Anything else

**Related issues:**
- #7826 — *"Issues disappearing from dataset"* — reports the same symptoms (issues silently vanishing from DevLake while still present in Jira under the board filter). The kanban sub-filter may be an undiscovered root cause for at least some of those reports.
- #8196 — *"Is there a way to allow DevLake to connect on Jira Projects instead of Jira Boards?"* — requesting project-based scoping, which would also bypass this problem since the search API does not apply board-level sub-filters.

**Potential solutions:**

**Option A: Use the search API instead of the board API.** Replace the board issue endpoint with the Jira search API (`api/2/search` or `api/3/search/jql`), using the board's saved filter JQL directly. The saved filter JQL is already fetched and stored in `_tool_jira_boards.jql` by `board_filter_begin_collector.go`. The epic collector (`epic_collector.go`) already uses this search-based approach as a reference pattern.

**Option B: Parse the sub-filter and allow override.** Read the `subQuery` field from the board configuration response (already fetched but not parsed), store it on `_tool_jira_boards` (new column), and add a scope config flag (e.g., `ignore_board_sub_filter`) that switches to the search API when enabled. This preserves backward compatibility while giving users control.

### Version

v1.0.3-beta12@1d21f19

### Code of Conduct

jbsmith7741:
### Summary

Jira boards configured with kanban sub-filters silently exclude issues when collected via the standard board-issue endpoint. This PR switches issue collection to use the Jira Search API (JQL) derived from the board's filter configuration, ensuring all matching issues are ingested regardless of sub-filter settings.

Additionally fixes:
- `SubQuery` struct shape to correctly unmarshal the nested API response
- Config-UI connection list passing the React `MouseEvent` into `onCreate` (which expects no arguments), causing the synthetic event object to be stored as plugin state. Fixed by wrapping the handler in an arrow function so `onCreate()` is called cleanly with no arguments.
- Config-UI connection form crashing when a plugin config omits `initialValues` — downstream code spreads/iterates this value, so destructuring without a default produced `undefined` and threw at runtime. Fixed by defaulting to `{}` during destructuring.

Closes #8917

### Screenshots

N/A — backend data-pipeline change with no UI impact beyond the crash fixes.

### Other Information

- New migration script `20260611_add_sub_query_to_boards.go` adds the `sub_query` column to `_tool_jira_boards`.
- Unit tests added for `BoardConfiguration` unmarshalling and the search-API issue collector.

### DIFF
--- a/backend/plugins/jira/models/board.go
+++ b/backend/plugins/jira/models/board.go
@@ -34,6 +34,7 @@ type JiraBoard struct {
 	Self         string `json:"self" mapstructure:"self" gorm:"type:varchar(255)"`
 	Type         string `json:"type" mapstructure:"type" gorm:"type:varchar(100)"`
 	Jql          string `json:"jql" mapstructure:"jql"`
+	SubQuery     string `json:"subQuery" mapstructure:"subQuery"`
 }
 
 func (b JiraBoard) ScopeId() string {
--- a/backend/plugins/jira/models/migrationscripts/20260611_add_sub_query_to_boards.go
+++ b/backend/plugins/jira/models/migrationscripts/20260611_add_sub_query_to_boards.go
@@ -1,0 +1,46 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package migrationscripts
+
+import (
+	"github.com/apache/incubator-devlake/core/context"
+	"github.com/apache/incubator-devlake/core/errors"
+	"github.com/apache/incubator-devlake/helpers/migrationhelper"
+)
+
+type JiraBoard20260611 struct {
+	SubQuery string
+}
+
+func (JiraBoard20260611) TableName() string {
+	return "_tool_jira_boards"
+}
+
+type addSubQueryToBoards struct{}
+
+func (script *addSubQueryToBoards) Up(basicRes context.BasicRes) errors.Error {
+	return migrationhelper.AutoMigrateTables(basicRes, &JiraBoard20260611{})
+}
+
+func (*addSubQueryToBoards) Version() uint64 {
+	return 20260611140000
+}
+
+func (*addSubQueryToBoards) Name() string {
+	return "add sub_query to _tool_jira_boards"
+}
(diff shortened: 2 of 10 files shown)


jbsmith7741:
@klesh @Startrekzky 
Is this something you could look at?

klesh:
Good job. Looks good to me overall. However, I would like to know if it was tested. If so, which versions (Jira Cloud, Server, Data Center) did you test against?

jbsmith7741:
This is the version I tested it against based on the /rest/api/2/serverInfo 

``` 
{
  "version": "1001.0.0-SNAPSHOT",
  "deploymentType": "Cloud",
  "buildNumber": 100292
}
```

klesh (approved):
LGTM.
Thanks for your contribution.

PR #8671, opened 2025-12-31
feat(azuredevops): add environment_pattern for extracting environment…

warren830:
… names from job/stage names

This enhancement addresses the issue where the Azure DevOps plugin was unable to correctly detect production deployments when the environment name is embedded in job/stage names rather than being in a standard format.

Changes:
- Add `environment_pattern` scope config field that supports regex capture groups to extract environment names from job/stage names
- Collect both Job and Stage records from the timeline API (previously only Job)
- Add `identifier`, `type`, and `parent_id` fields to the Job model
- Update environment detection logic:
  - If environment_pattern is configured, extract the environment name first
  - Apply production_pattern to the extracted environment name
  - Fall back to matching production_pattern against job name if no extraction
- Fix default environment behavior: only default to PRODUCTION when production_pattern is not configured (was defaulting PRODUCTION always)
- Add comprehensive tests for the new functionality

Example configuration for pipelines with jobs like 'deploy_xxxx-prod_helm':
- deployment_pattern: deploy
- production_pattern: prod
- environment_pattern: (?:deploy|predeploy)[_-](.+?)(?:[_-](?:helm|terraform))?$

This extracts 'xxxx-prod' from the job name and then applies production_pattern to correctly identify it as a production deployment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### ⚠️ Pre Checklist

> Please complete _ALL_ items in this checklist, and remove before submitting

- [ ] I have read through the [Contributing Documentation](https://devlake.apache.org/community/).
- [ ] I have added relevant tests.
- [ ] I have added relevant documentation.
- [ ] I will add labels to the PR, such as `pr-type/bug-fix`, `pr-type/feature-development`, etc.

### Summary
What does this PR do?

### Does this close any open issues?
Closes 8869

### Screenshots
Include any relevant screenshots here.

### Other Information
Any other information that is important to this PR.

PR #8932, opened 2026-06-16
fix(azuredevops): add missing migrations for Job model fields from #8671

magiskboy:
## Summary

Fixes a schema mismatch introduced by #8671 where the Azure DevOps Python plugin's `Job` model gained `identifier`, `type`, and `parent_id` fields, and `GitRepositoryConfig` gained `environment_pattern`, but no database migrations were added.

This caused the **Convert Jobs** subtask to fail on MySQL with:

```
Unknown column '_tool_azuredevops_jobs.identifier' in 'field list'
```

Affected releases: `v1.0.3-beta9` through `v1.0.3-beta12`.

## Changes

- Add migration `20251231120000` to add `identifier`, `type`, and `parent_id` columns to `_tool_azuredevops_jobs`
- Add migration `20251231120001` to add `environment_pattern` column to `_tool_azuredevops_gitrepositoryconfigs`

Uses `add_column` operations which skip columns that already exist (safe for users who applied manual SQL workarounds).

## Test plan

- [x] Start DevLake with an existing DB missing these columns and confirm migrations run on startup
- [x] Run Azure DevOps pipeline collect including **Convert Jobs** — should complete without SQL errors
- [x] Verify fresh install also gets columns via migrations (init schema + new migrations)
- [x] Confirm scope config with `environment_pattern` can be saved after migration

Made with [Cursor](https://cursor.com)

### DIFF
--- a/backend/python/plugins/azuredevops/azuredevops/migrations.py
+++ b/backend/python/plugins/azuredevops/azuredevops/migrations.py
@@ -247,3 +247,15 @@ def add_updated_url_column_length_in_raw_azuredevops_builds(b: MigrationScriptBu
     END $$;
     """, Dialect.POSTGRESQL)
 
+
+@migration(20251231120000, name="add job fields added in environment_pattern feature")
+def add_job_fields_for_environment_pattern(b: MigrationScriptBuilder):
+    table = '_tool_azuredevops_jobs'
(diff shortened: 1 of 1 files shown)


magiskboy:
Closing — developing this fix independently on magiskboy/devlake fork.

PR #8934, opened 2026-06-16
fix(migrations): remove hard-coded bigint types for postgres

### LINKED ISSUE #8867: [Bug][Migration][Postgres] testmo and q_dev plugins use MySQL-specific BIGINT UNSIGNED in raw SQL ALTER TABLE
### Search before asking

### What happened

Two migration scripts use `BIGINT UNSIGNED` and (in one case) MySQL backticks
in raw `db.Exec()` ALTER TABLE statements without dialect-aware branching.
Both fail on PostgreSQL with `ERROR: syntax error` (SQLSTATE 42601), blocking
any upgrade path that crosses these scripts on a Postgres backend.

**Affected migrations** (audited at v1.0.3-beta12):

1. `backend/plugins/testmo/models/migrationscripts/20250629_add_scope_config_id_to_projects.go`
   ```go
   return db.Exec("ALTER TABLE `_tool_testmo_projects` ADD COLUMN `scope_config_id` bigint unsigned NOT NULL DEFAULT 0")

— backticks + bigint unsigned.

2. `backend/plugins/q_dev/models/migrationscripts/20251123_add_scope_config_id_to_s3_slice.go`
```
err := db.Exec(`
    ALTER TABLE _tool_q_dev_s3_slices
    ADD COLUMN scope_config_id BIGINT UNSIGNED DEFAULT 0
`)
```
— BIGINT UNSIGNED (Postgres has no unsigned modifier).
This is the same class of bug as #8564 (closed by #8565), reproduced in
two new migrations.

Sibling migration `plugins/testmo/.../20250629_add_scope_config_id.go` in the
same plugin uses the portable `migrationhelper.AutoMigrateTables(...)`
pattern correctly — confirming this is a coding mistake, not a missing
infrastructure piece.

Migrations should either:
- Use `migrationhelper.AutoMigrateTables(&entity{})` with GORM tags (portable across MySQL/Postgres), or
- Branch on `db.Dialect()` for dialect-specific raw SQL (as `q_dev/20250320_modify_file_meta.go` does correctly after #8565).

### How to reproduce

1. Deploy DevLake v1.0.3-beta12 (or any beta from 2025-06+) with PostgreSQL 15+ as backend.
2. Run a fresh deployment, or upgrade from v1.0.1+.
3. Migration `20250629000001` (testmo) fails with `SQLSTATE 42601`.
4. Even if testmo is patched, the next blocker is `20251123000001` (q_dev) with the same root cause.

Reproduced on PostgreSQL 15 managed by the Zalando operator.

### Anything else

Recommendation: add a CI step that runs the full migration chain against
PostgreSQL on every PR. Most contributors use the portable helpers
correctly; these two slipped through specifically because they were
hand-written raw SQL bypassing the helpers.

Related: #8617 (teambition `ORDER BY` on `json` column — unresolved,
different root cause but blocks Postgres on the same upgrade path),
#8564 (q_dev MODIFY — fixed in #8565), #8350 (Postgres support parity
question, closed without resolution).

### Version

v1.0.3-beta12

### Code of Conduct

bujjibabukatta:
Fixes #8867.

Updated the Testmo and Q Dev scope_config_id migrations to avoid hard-coded BIGINT SQL type declarations in GORM tags. The migrations continue using migrationhelper.AutoMigrateTables, but now allow GORM to generate dialect-compatible column definitions for both MySQL and PostgreSQL.

Validated with git diff --check and targeted searches for the previous BIGINT UNSIGNED/raw ALTER TABLE patterns.

### DIFF
--- a/backend/plugins/q_dev/models/migrationscripts/20251123_add_scope_config_id_to_s3_slice.go
+++ b/backend/plugins/q_dev/models/migrationscripts/20251123_add_scope_config_id_to_s3_slice.go
@@ -26,7 +26,7 @@ import (
 type addScopeConfigIdToS3Slice struct{}
 
 type QDevS3Slice20251123 struct {
-	ScopeConfigId uint64 `gorm:"type:BIGINT DEFAULT 0"`
+	ScopeConfigId uint64 `gorm:"default:0"`
 }
 
 func (QDevS3Slice20251123) TableName() string {
--- a/backend/plugins/testmo/models/migrationscripts/20250629_add_scope_config_id_to_projects.go
+++ b/backend/plugins/testmo/models/migrationscripts/20250629_add_scope_config_id_to_projects.go
@@ -27,7 +27,7 @@ import (
 type addScopeConfigIdToProjects struct{}
 
 type TestmoProject20250629 struct {
-	ScopeConfigId uint64 `gorm:"type:BIGINT NOT NULL DEFAULT 0"`
+	ScopeConfigId uint64 `gorm:"not null;default:0"`
 }
 
 func (TestmoProject20250629) TableName() string {


klesh:
Are you using the project with all these plugins you are submitting PRs to?

bujjibabukatta:
@klesh I use DevLake with Postgres, and while I haven't personally set up Testmo or Q Dev specifically. I came across this issue #8867 and the fix was straightforward and removing the hardcoded dialect-specific types so GORM handles it correctly across databases. Happy to test it further if there's a specific scenario you'd like validated.Thanks

klesh:
Please make sure to test this before submitting. Just a quick reminder regarding the migration scripts: we shouldn't modify existing ones. Think of migrations as moving forward in time—altering past scripts breaks the sequence and can cause deployment issues. If we need changes, we should always roll forward with a new script!

PR #8939, opened 2026-06-18
feat(webhook): allow closing incidents via request body instead of url

### LINKED ISSUE #8910: [Feature][Webhooks] Move dinamic webhook url fields to body
### Search before asking

### Use case

Currently, Kibana v 8.19 webhook connectors require a fixed URL configuration per connector. This creates a limitation when integrating with devlake API webhook beacuse closing incident has dynamic path parameters like issueKey .

Kibana doen't support dynamically constructing the webhook URL using values from the alert context or previous action output.

As a result, it is not possible to reliably implement a “create + close” workflow using only Kibana alert with webhook actions when the external API requires the issueKey identifier in the URL path.

### Description

Allow both “create” and “close” operations to use a fixed webhook URL.
Move all dynamic identifiers (e.g., issueKey) into the request body.

**Workaround**

At the moment, this limitation forces users to implement external middleware to store and correlate issueKey values between alert firing and recovery events.

### Related issues

https://github.com/apache/devlake/discussions/8905

### Code of Conduct

bujjibabukatta:
### Summary

Added new webhook close issue endpoints that accept `issueKey` in the request body instead of the URL path. This enables Kibana v8.x webhook connectors to close incidents using a fixed URL.

### Changes

* Added `POST /connections/:connectionId/issue/close`
* Added `POST /connections/by-name/:connectionName/issue/close`
* Added request body support for `issueKey`
* Reused existing `closeIssue()` logic to avoid code duplication
* Regenerated Swagger documentation

### Notes

* No breaking changes
* Existing path-based endpoints remain fully supported
* Compatible with Kibana's fixed URL webhook configuration


klesh:
Looks good.
Did you test it on you local machine?
I would like some screenshots to confirm it was well tested since there is no test case and I am not familiar with this part..

bujjibabukatta:
Hey @klesh  These endpoints are basically just thin wrappers — they grab issueKey from the body and call the existing closeIssue() function, that's it. Nothing new that needs testing beyond what's already covered. Would you be okay merging this and I'll add tests in a follow-up?

klesh (approved):
LGTM.
Thanks for your contribution.

Issue #8249, opened 2024-12-18
[Question][DORA] Commits before first recorded deployments misattributed to random deployment


qpawelc:
## Question

Hey all!

I am using Devlake `v1.0.1` and I am currently ingesting the commits from a project with the Gitlab integration, and then ingesting the deployments for the project via webhook (from Spinnaker). I do not have the all-time deployment data for my project, only the last couple of months. After I run the collect data job, I noticed that in the `project_pr_metrics` table, all of the commits for the project that occurred before I onboarded to Spinnaker seem to get associated with a "seemingly random" deployment.

It may be easier to visualize this. Take a look at this screenshot. The bottom blue dots represent commits. The top dots represent deployments. The lines represent the commit-to-deployment relationship in the `project_pr_metrics` table. As you can see, all of my commits that occurred before the beginning of the my deployment data seem to ALL become associated with a single deployment, causing certain metrics to be very skewed.

Does anyone have any advice on how I can:

- Debug why devlake chooses to associated these commits to this deploy? For example, why wouldnt it be the first deploy? Or preferably, not create associations between commits and deployments where commits occurred before the first deployment?
- Work around this issue? One idea I have is to set the blueprint for each project only to ingest data after the second successful production deployment.

Thanks for your time and support!

## Screenshots
![image](https://github.com/user-attachments/assets/b6b0eb2c-007d-40e4-afa9-20cf471b9ee7)

## Additional context

Its worth noting that I have found a similar [issue](https://github.com/apache/incubator-devlake/issues/7193). In this [comment](https://github.com/apache/incubator-devlake/issues/7193#issuecomment-2085140726) @nicolavolpini mentioned that "still shows several PRs associated to the same deployment webhook" but no additional context was posted on how this issue could be resolved.

petkostas:
Looks similar to https://github.com/apache/incubator-devlake/issues/8188

Issue #8188, opened 2024-11-12
[Bug][DORA-change_lead_time_calculator] Wrong deployment reference


kostas-petrakis:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

While analyzing the data gathered from our GitHub integration, I stumbled upon an anomaly in a minor subset of deployments. Oddly enough, some deployments were reporting a deployment of incosistent time (in my example 30 weeks).
Given my familiarity with our GitHub operations, I was certain this wasn't accurate, prompting me to dig deeper. My investigation points towards a potential issue with the `getDeploymentCommit` function, particularly in relation to `EnrichPrevSuccessDeploymentCommit`.

To illustrate, I've attached an example PR
```
{
"pull_requests": [
	{
		"id" : "github:GithubPullRequest:4:1806504909",
		"created_at" : "2024-11-12 08:34:13.134",
		"updated_at" : "2024-11-12 08:34:13.134",
		"_raw_data_params" : "{\"ConnectionId\":4,\"Name\":\"********\"}",
		"_raw_data_table" : "_raw_github_api_pull_requests",
		"_raw_data_id" : 4499,
		"_raw_data_remark" : "",
		"base_repo_id" : "github:GithubRepo:4:***********",
		"base_ref" : "main",
		"base_commit_sha" : "a51af0dabd2d9e9bd7b23e432688842cf086c248",
		"head_repo_id" : "github:GithubRepo:4:*******",
		"head_ref" : "feature\***************",
		"head_commit_sha" : "e6a73f3095d1c97d4d4576e3c06719fb07938bd1",
		"merge_commit_sha" : "b297c73092685c4b1a0ee62c2326d77115a86e94",
		"status" : "MERGED",
		"original_status" : "closed",
		"type" : "",
		"component" : "",
		"title" : "****************************",
		"description" : "null",
		"url" : "https:\/\/github.com\/******\/*******\/pull\/1025",
		"author_name" : "***************",
		"author_id" : "github:GithubAccount:4:***********",
		"parent_pr_id" : "",
		"pull_request_key" : 1025,
		"created_date" : "2024-04-04 09:08:03",
		"merged_date" : "2024-04-05 09:00:16",
		"closed_date" : "2024-04-05 09:00:16",
		"additions" : 0,
		"deletions" : 0,
		"merged_by_name" : "",
[...]

PR #8942, opened 2026-06-19
fix(dora): use two-phase deployment lookup to prevent first-deploymen…

### LINKED ISSUE #8790: [Bug][DORA-change_lead_time_calculator] Wrong deployment reference

bujjibabukatta:
## Summary

Fixes the "wrong deployment reference" bug where PRs were incorrectly
associated with old deployments, causing massively skewed median lead
time for change metrics.

### Root cause

`batchFetchDeployments` used **only** diff-based mapping, which
deliberately excludes the very first deployment in a collected window
(via `prev_success_deployment_commit_id <> ''`). This guard is correct
for the diff path — without it, the first deployment would absorb every
historical PR via `commits_diffs`. However, it also blocked a legitimate
case: when a deployment's own `commit_sha` **directly equals** a PR's
`merge_commit_sha`, that association is precise and safe, even for the
first deployment.

This caused the scenario reported in #8790: projects using the GitLab
connector to onboard historical deployments for the first time would see
the first deployment incorrectly pull in all old MRs.

### Fix

Two-phase strategy in `batchFetchDeployments`:

**Phase 1 — direct match (new):**
Query all successful PRODUCTION deployments in the project, keyed by
`commit_sha`. If a deployment's `commit_sha` equals the PR's
`merge_commit_sha`, associate them. No `commits_diffs` join needed —
the match is exact, so there is no over-mapping risk.

**Phase 2 — diff-based fallback (unchanged):**
Retain the existing `commits_diffs` join with the
`prev_success_deployment_commit_id <> ''` guard. Only fills in PRs not
already resolved by Phase 1.

### Test coverage

All existing e2e test cases in `calculate_change_lead_time_test.go`
pass without change. In the existing fixtures, PR `merge_commit_sha`
values are distinct from deployment `commit_sha` values, so Phase 1
produces no matches and Phase 2 runs identically to before — verifying
no regression.

The bug scenario (Phase 1 actually matching) is the real-world case
described in #8790 and #8188 and is not yet covered by a fixture test.
A follow-up test case with `deployment.commit_sha == pr.merge_commit_sha`
would be a good addition.

### Related issues
Closes #8790
Related: #8188, #8249, #7193


klesh (commented):
The fix looks legit to me. Good work.
However, I think we should update the e2e test to include the exact matching case as well.

bujjibabukatta:
Thanks for the review, @klesh 

Agreed i have  added the e2e fixture in commit 4f14dee. It covers the Phase 1 direct-match scenario (`deployment.commit_sha == pr.merge_commit_sha`) that was the root cause in #8790, which wasn't exercised by the existing test data.

All existing cases still pass, and the new fixture confirms Phase 1 correctly associates the deployment without over-mapping. Let me know if you'd like any changes to the fixture structure!

klesh (commented):
Please fix the linting. Thanks.

klesh (approved):
LGTM.
Thanks for your contribution.

hera-a-glitch:
Amazing really looking forward to this any ETA for release?

PR #8946, opened 2026-06-20
fix(customize): skip orphaned _raw_data_id rows instead of aborting extraction

### LINKED ISSUE #8945: [Bug][customize] ExtractCustomizedFields aborts the whole scan on the first row with an orphaned _raw_data_id
### Search before asking

### What happened

The `customize` plugin's `ExtractCustomizedFields` subtask silently stops extracting custom fields the moment it encounters a single domain-layer row whose raw record no longer exists. Every remaining row in the cursor is left unprocessed, yet the subtask finishes as `TASK_COMPLETED`, so there is no error or warning.

**Root cause**: `backend/plugins/customize/tasks/customized_fields_extractor.go`

The extractor cursor uses a LEFT JOIN to the raw table:

```go
dal.Join(fmt.Sprintf(" LEFT JOIN %s ON %s._raw_data_id = %s.id", rawTable, table, rawTable)),

// When a domain row's _raw_data_id is orphaned (the raw record was cleaned up / replaced by a later collection), the joined data column is NULL. 
// The per-row type switch then falls through to its default branch:

switch blob := row["data"].(type) {
  case []byte:
      ...
  case string:
      ...
  default:
      return nil   // <-- exits the entire function, skipping all remaining rows
}
```

`return nil` returns from `extractCustomizedFields` entirely instead of skipping just that one row. Because cursor order is arbitrary, the first orphaned row encountered ends extraction for everything after it.

**Impact**
- Custom `x_*` columns silently stop being populated for a cursor-order-dependent subset of rows, some boards/projects populate fully while others stay entirely `NULL`.
- The effect persists across syncs because already-written custom columns are not cleared, which makes it look like only "new" data is affected.
- No error surfaces, the subtask reports success.

In our instance, a connection with 30,287 matched rows had 1,382 orphaned rows. Extraction stopped early and left ~28,000 rows unpopulated.

A domain row with no matching raw record should be skipped, and extraction should continue with the remaining rows. One orphaned `_raw_data_id` should never abort the entire subtask.

### How to reproduce

1. Configure a customize transformation rule that maps a JSON path into a custom column, e.g. `issues.x_defect_category` <- `fields.customfield_XXXXX.value`, applied either via a blueprint `afterPlan` or a standalone `customize` pipeline.
2. Ensure at least one matched domain row has a `_raw_data_id` with no corresponding record in the raw table. This happens naturally after re-collection orphans old raw records.
3. Run the customize subtask.
4. Observe: rows ordered after the first orphaned row are left `NULL`, despite having valid raw data and matching the filter, the subtask still completes successfully.

Confirming query (MySQL), counts orphaned rows within the rule's filter:
```sql
SELECT SUM(r.id IS NULL) AS orphaned, COUNT(*) AS total
FROM issues i
LEFT JOIN _raw_jira_api_issues r ON i._raw_data_id = r.id
WHERE i._raw_data_table = '_raw_jira_api_issues'
  AND i._raw_data_params LIKE '{"ConnectionId":1,%';
```

### Anything else

Happens every time once at least one orphaned `_raw_data_id` exists among the matched rows.
Which rows end up unpopulated depends on cursor order, so the symptom looks erratic across projects/boards even though the cause is deterministic.

Suggested fix: change the `default` branch in the value switch from `return nil` to `continue`, so orphaned rows are skipped instead of aborting the scan. Happy to submit this as a PR with a regression test for the orphaned-row case.

Searched existing issues, closest are #8173 (create-field, unrelated) and #7571 (wildcard rawDataParams, unrelated) but neither covers this extraction-abort.

v1.0.3-beta8 (also confirmed unchanged on main at v1.0.3-beta13)

### Version

v1.0.3-beta8

### Code of Conduct

danielemoraschi:
### Summary
ExtractCustomizedFields used a `LEFT JOIN` to the raw table, so a domain row whose `_raw_data_id` has no matching raw record yields `NULL` data. That hit the type switch default branch, which did `return nil` and exited the whole subtask, silently leaving every later row in the (unordered) cursor unprocessed while still reporting success.

Change the default branch to `continue` so an orphaned row is skipped and extraction proceeds with the remaining rows.

Closes #8945

### DIFF
--- a/backend/plugins/customize/tasks/customized_fields_extractor.go
+++ b/backend/plugins/customize/tasks/customized_fields_extractor.go
@@ -142,7 +142,9 @@ func extractCustomizedFields(ctx context.Context, d dal.Dal, table, rawTable, ra
 				}
 			}
 		default:
-			return nil
+			// orphaned/NULL raw data (LEFT JOIN found no matching raw record): skip this
+			// row and keep scanning, instead of aborting the whole subtask.
+			continue
 		}
 
 		if len(updates) > 0 {
--- a/backend/plugins/customize/tasks/customized_fields_extractor_test.go
+++ b/backend/plugins/customize/tasks/customized_fields_extractor_test.go
@@ -1,0 +1,83 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tasks
+
+import (
+	"context"
+	"testing"
+
+	"github.com/apache/incubator-devlake/core/dal"
+	mockdal "github.com/apache/incubator-devlake/mocks/core/dal"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+// TestExtractCustomizedFieldsSkipsOrphanedRows is a regression test for the case where a
+// domain-layer row's `_raw_data_id` has no matching raw record (LEFT JOIN yields NULL data).
+// Such a row must be skipped and extraction must continue with the remaining rows, instead of
+// aborting the whole scan on the first orphaned row.
+func TestExtractCustomizedFieldsSkipsOrphanedRows(t *testing.T) {
+	mockRows := new(mockdal.Rows)
+	mockRows.On("Next").Return(true).Times(2)
+	mockRows.On("Next").Return(false).Once()
+	mockRows.On("Close").Return(nil).Once()
+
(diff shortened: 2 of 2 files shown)


backend/plugins/customize/tasks/customized_fields_extractor.go
```diff
@@ -142,7 +142,9 @@ func extractCustomizedFields(ctx context.Context, d dal.Dal, table, rawTable, ra
 				}
 			}
 		default:
-			return nil
+			// orphaned/NULL raw data (LEFT JOIN found no matching raw record): skip this
+			// row and keep scanning, instead of aborting the whole subtask.
```
klesh, line 146: Could you please print a warning message here? Thanks.

danielemoraschi:
Closing this as superseded by #9027, which landed on `main` in 5eee167.

That change covers the same root cause. The LEFT JOIN against the raw table
yields a NULL `data` column for a domain row whose `_raw_data_id` no longer has
a matching raw record, and that value fell through to `default: return nil`,
ending the scan. #9027 also goes further than this PR did:

* @klesh's review note is addressed there. `extractCustomizedFields` returns the
  number of orphaned rows, and `ExtractCustomizedFields` logs a warning naming
  the table, the raw table, and the count.
* `rows.Err()` is now returned at the end of the scan, so an error raised while
  iterating the cursor is no longer discarded. That is the same class of silent
  partial processing this PR left in place.
* Test coverage is broader: skipping an orphan and continuing, every row
  orphaned, no orphans at all, and a cancelled context. This PR had one
  regression test.

Nothing in this branch is additive on top of that, so there is no point in
rebasing. Thanks @klesh for the review.

PR #8950, opened 2026-06-23
feat(jira): add JIRA_SKIP_UNPARSEABLE_ISSUES to skip unparseable issu…

### LINKED ISSUE #8949: [Feature][jira] Add JIRA_SKIP_UNPARSEABLE_ISSUES to continue collection when an issue page can't be parsed
### Search before asking

### Use case

### Description

  collection finish.

### Related issues

_No response_

### Code of Conduct

Murad-Suleymanov:
### Summary

Adds an opt-in config `JIRA_SKIP_UNPARSEABLE_ISSUES` to the Jira `collectIssues` subtask.
When an issue page returns a 2xx status but a body that can't be parsed as JSON (e.g. a
truncated/partial response), the subtask currently fails and aborts the whole Jira collection.
With this flag set to `true`, the unparseable page is logged and skipped so the rest of the
collection finishes. Default (`false`) keeps current behaviour.

Closes #8949

### DIFF
--- a/backend/plugins/jira/tasks/issue_collector.go
+++ b/backend/plugins/jira/tasks/issue_collector.go
@@ -135,6 +135,12 @@ func CollectIssues(taskCtx plugin.SubTaskContext) errors.Error {
 			}
 			err = json.Unmarshal(blob, &data)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


klesh:
This could work. But I think a better approach would be adding the option to ScopeConfig and allowing user to toggle on the config-ui.

PR #8951, opened 2026-06-23
fix(authentication): feature should be disabled by default

rbstp:
### Summary
Contrary to what the PR for the feature mentioned, the setting was enabled by default, instead of disabled.

Follow up to https://github.com/apache/devlake/pull/8854

### Screenshots
n/a


klesh (approved):
LGTM.
Thanks for your contribution.

PR #8952, opened 2026-06-24
fix(config-ui): Recover "Create a New Connection" action

kirillNevedrov:
### Summary
This PR fixes "Create a New Connection" action. The issue is that clicking the "Create a New Connection" button causes `Error: Objects are not valid as a React child` error. It happens because react event object is passed as a pluginName argument to handleShowFormDialog function on button click. So the fix calls onCreate function with no arguments.

No

### Screenshots
image


klesh (approved):
LGTM
Thanks for your contribution

PR #8957, opened 2026-06-25
fix(python): align PullRequestLabels and CommitParent composite primary keys with Go domain layer

### LINKED ISSUE #8954: [Bug][Pydevlake] Misalignement of Models between Go and Python - missing Primary Keys in PullRequestLabels
### Search before asking

### What happened

Go code defines PrimaryKeys like that:

```go
type PullRequestLabel struct {
	PullRequestId string `json:"id" gorm:"primaryKey;type:varchar(255);comment:This key is generated based on details from the original plugin"` // format: <Plugin>:<Entity>:<PK0>:<PK1>
	LabelName     string `gorm:"primaryKey;type:varchar(255)"`
	common.NoPKModel
}
```

and Python:

```python
class PullRequestLabels(NoPKModel, table=True):
    __tablename__ = 'pull_request_labels'
    pull_request_id: str = Field(primary_key=True)
    label_name: str
```

they should align

### How to reproduce

nothing to repro - this is problem in the code itself

### Anything else

workaround is to use extend_existing to internally in the code expand definition:

```python
class PullRequestLabels(NoPKModel, table=True):
    """
    Domain model for pull_request_labels table with the correct composite PK.

    Patches the vendored pydevlake.domain_layer.code.PullRequestLabels to use
    a composite primary key (pull_request_id, label_name) matching the Go
    DevLake domain layer definition.
    """

    __tablename__ = "pull_request_labels"
    __table_args__ = {"extend_existing": True}

    pull_request_id: str = Field(primary_key=True)
    label_name: str = Field(primary_key=True)
```

### Version

f2de2dcb19cdb2af5b2613a2b939f50a33bfd209

### Code of Conduct

bujjibabukatta:
## What does this PR do?

Fixes two missing composite primary keys in the Python domain layer that were misaligned with the Go definitions.

## Changes

- `PullRequestLabels.label_name`: added `Field(primary_key=True)`
- `CommitParent.parent_commit_sha`: added `Field(primary_key=True)`

## Why

Without the correct composite PKs, upserts match on fewer columns than intended,
causing rows that differ only in the missing PK column to silently overwrite
existing records instead of being treated as distinct entries.

## Related Issue

Closes #8954

### DIFF
--- a/backend/python/pydevlake/pydevlake/domain_layer/code.py
+++ b/backend/python/pydevlake/pydevlake/domain_layer/code.py
@@ -50,7 +50,7 @@ class PullRequest(DomainModel, table=True):
 class PullRequestLabels(NoPKModel, table=True):
     __tablename__ = 'pull_request_labels'
     pull_request_id: str = Field(primary_key=True)
-    label_name: str
+    label_name: str = Field(primary_key=True)
 
 
 class PullRequestCommit(NoPKModel, table=True):
@@ -94,7 +94,7 @@ class Commit(NoPKModel, table=True):
 class CommitParent(NoPKModel, table=True):
     __tablename__ = 'commit_parents'
     commit_sha: str = Field(primary_key=True)
-    parent_commit_sha: str
+    parent_commit_sha: str = Field(primary_key=True)
 
 
 class CommitsDiff(DomainModel, table=True):
@@ -111,6 +111,7 @@ class RefCommit(NoPKModel, table=True):
     old_ref_id: str = Field(primary_key=True)
     new_commit_sha: str
     old_commit_sha: str
+
 
 class Component(NoPKModel, table=True):
     __tablename__ = 'components'


bujjibabukatta:

klesh (commented):
This doesn't make sense to me. Why adding those extra columns to primary keys?

bujjibabukatta:
Hi @klesh 

Thanks for taking a look, and fair pushback — let me walk through why I made this change, since it wasn't obvious from the diff alone.

Domain records get saved via session.merge(domain_model) in Convertor._save(). SQLAlchemy's merge() decides insert vs update based on the primary key. With only pull_request_id (or commit_sha) marked as PK, two different labels on the same PR — or two different parents on the same commit — get treated as the same row, so the second merge silently overwrites the first and only one would ever survive.

It also matches Go — both fields are already gorm:"primaryKey" there for PullRequestLabel and CommitParent, and the actual DB tables already have this composite PK (Go owns migrations for domain-layer tables, pydevlake doesn't create or alter them). So this isn't a schema change at all — it's just correcting SQLAlchemy's mapper metadata to match what Go already declared and what's already in the database. No migration needed on either side.

This file already uses the same composite-PK pattern for PullRequestCommit, RefCommit, and CommitsDiff, so it's consistent with existing conventions too.

bujjibabukatta:
Hi @klesh can you please review and approve pull request?

klesh (approved):
You are right.
Thanks for the clarification.

PR #8972, opened 2026-07-02
feat(jira): Allow extra JQL when querying JIRA board

VaughanJ:
- [ x ] I have read through the [Contributing Documentation](https://devlake.apache.org/community/).
- [ x ] I have added relevant tests.
- [ x ] I have added relevant documentation.
- [ x ] I will add labels to the PR, such as `pr-type/bug-fix`, `pr-type/feature-development`, etc.

### Summary
When querying the JIRA board for issues, you can apply an optional JQL query to further filter the board, useful when a board has many components on it.

### Screenshots
image


backend/plugins/jira/impl/impl.go
```diff
@@ -249,6 +251,22 @@ func (p Jira) PrepareTaskData(taskCtx plugin.TaskContext, options map[string]int
 		Options:        &op,
 		ApiClient:      jiraApiClient,
 		JiraServerInfo: *info,
+		Board:          scope,
+	}
+
+	// Look up the DevLake project this board belongs to via project_mapping.
+	// This is best-effort: if the board is not yet mapped (e.g. first run or
+	// manual trigger outside a blueprint) we leave DevLakeProjectName empty.
+	if scope != nil {
```
klesh: I see why you did it – the `projectName` is not available for the data source plugin.
However, the solution seems too hacky and ad hoc.
We need a better solution:
1. adding the `projectName` to the `DataSourcePluginBlueprintV200` interface
2. adding a new user-definable column to `Board` and using it instead of the `projectName` variable.

klesh (changes requested):

backend/plugins/jira/impl/impl.go
VaughanJ: Can you expand on point 2, how do you see it working? Also is something similar seen elsewhere in the repo?

As what I'm trying to achieve is that the ExtraJql query has the Devlake project name as a template that can be used in the query.

Ideally I don't want to have to manually key project names as we have 1000s of projects and it seems wasteful when the Jira task and blueprint are already running under a project. So auto discovery of the project name to be used in the JQL is my angle.

VaughanJ: I have updated the code to give only the ExtraJQL functionality, removing the templating of DevLakeProjectName to make the PR cleaner. 

We can discuss options how/if it's possible to get the Dev Lake Project Name under a separate piece of work. 

As ExtraJQL is a beneficial standalone feature and can still work for my needs, I will just have one data scope per project.

klesh (approved):
LGTM

backend/plugins/jira/impl/impl.go
klesh: @VaughanJ Thanks for your support. I appreciate it.
Regarding the point 2: I checked and config-ui doesn't have an interface for editing the Board, so it would be quite an undertaking for the proposal to work. However, it would be a better solution IMO.

PR #8979, opened 2026-07-07
fix: docker file path in build-grafana-image target

kirillNevedrov:
### Summary
What does this PR do?

`build-grafana-image` target from Makefile file has wrong docker file path `./backend/Dockerfile`. So the docker image build fails with `file not found` error.
This MR changes the file path from `./backend/Dockerfile` to `./Dockerfile`, allowing the image to build successfully.

No

### Screenshots
N/A

### Other Information
N/A


klesh (approved):
Good catch.
Thanks for your contribution.

PR #9003, opened 2026-07-20
chore(docker): bump PostgreSQL to 18.1 and MySQL to 8.4.10

DoDiODev:
### Summary

Bump the database engine images used by the local dev, devcontainer, temporal deployment and remote e2e compose files:

- PostgreSQL `17.2` (dev) / `14.2-alpine` (e2e) -> `18.1` / `18.1-alpine`
- MySQL `8` / `8.0.26` -> `8.4.10`

### Details

- **PostgreSQL 18 volume mount (e2e):** `backend/test/e2e/remote/docker-compose.test.yml` now mounts the `postgres-test` volume at `/var/lib/postgresql` (the parent of PostgreSQL 18's new versioned data directory) instead of the legacy `/var/lib/postgresql/data`, matching the layout already used by `docker-compose-dev-postgresql.yml`.
- **Grafana datasource provisioning:** `grafana/scripts/entrypoint.sh` sets `postgresVersion` to `1800` so the provisioned PostgreSQL datasource targets PostgreSQL 18.

### Related / dependencies

- **Recommended before this on PostgreSQL: #9001** (`fix(dalgorm): make DropIndexes idempotent by skipping missing indexes`). It is a functionally independent fix, but without it a pre-existing migration (`change _tool_sonarqube_issue_code_blocks.component type to text`) aborts on PostgreSQL with `SQLSTATE 42704` when dropping an already-absent index. That failure is unrelated to the image bumps here, but it does surface during the PostgreSQL migration run. On MySQL this PR is unaffected by #9001.

### Out of scope

- Grafana database env parametrization / port changes.
- Any Go, Python, Node, or dashboard changes.

### Testing

- `docker compose config -q` passes for all changed compose files.
- `bash -n grafana/scripts/entrypoint.sh` passes.
- Manual gate — MySQL: fresh database starts, migrations run, smoke flow OK, data persists across restart.
- Manual gate — PostgreSQL 18: fresh database with `FORCE_MIGRATION=true`, container starts and accepts connections, migrations complete (with #9001 applied), `/ready` OK, Grafana datasource targets PG18; volume persists across restart.

### Rollback

Revert this commit to restore the previous image tags and the legacy e2e volume mount.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -24,7 +24,7 @@ services:
       - default
     command: sleep infinity
   mysql:
-    image: mysql:8
+    image: mysql:8.4.10
     volumes:
       - mysql-storage:/var/lib/mysql
     restart: always
--- a/backend/test/e2e/remote/docker-compose.test.yml
+++ b/backend/test/e2e/remote/docker-compose.test.yml
@@ -19,7 +19,7 @@ version: "3"
 services:
 
   mysql-test:
-    image: mysql:8.0.26
+    image: mysql:8.4.10
     platform: linux/x86_64
     volumes:
       - mysql-test-storage:/var/lib/mysql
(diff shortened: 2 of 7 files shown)


klesh (approved):
LGTM.
Thanks for your contribution.

PR #9008, opened 2026-07-21
Fix webhook authorization when AUTH_ENABLED=true

### LINKED ISSUE #9007: [Bug][Webhook] Requests via /rest/... alias return 401 when AUTH_ENABLED=true, even with a valid API key
### Search before asking

### What happened

When `AUTH_ENABLED=true`, requests to REST-style webhook endpoints (e.g. /rest/plugins/webhook/connections/:id/deployments) with a valid `Authorization: Bearer <key>` are rejected with `401 Unauthorized`, even though the key is valid and authorized for the path.

Root cause: RestAuthentication (backend/server/api/middlewares.go) validates the key and sets the user on the gin context via `c.Set(common.USER, ...)`, then reroutes /rest/... → /plugins/... via gin's HandleContext. That call resets c.Keys internally, wiping the just-set user, so the downstream RequireAuth middleware finds no user and returns 401.

A valid, authorized REST API key request should reach the handler and get 200, regardless of `AUTH_ENABLED`. The authenticated user should survive gin's context reset during the /rest/... → /plugins/... reroute.

### How to reproduce

1. Set `AUTH_ENABLED=true`, start DevLake.
2. Create an API key with AllowedPath covering a webhook endpoint, e.g. /plugins/webhook/connections/1/.*.
3. POST /rest/plugins/webhook/connections/1/deployments with Authorization: Bearer <api-key>.
4. Observe 401 instead of 200.

### Anything else

Only affects the /rest/... alias path (rerouted via HandleContext) when an auth-gating middleware like RequireAuth is active; direct /plugins/... requests are unaffected.

### Version

v1.0.3-beta15

### Code of Conduct

kamphaus:
### ⚠️ Pre Checklist

- I have added relevant documentation. → NA

### Summary

When AUTH_ENABLED=true, requests to REST-style webhook endpoints (e.g. /rest/plugins/webhook/connections/:id/deployments) were incorrectly rejected with 401 Unauthorized even when a valid, authorized API key was supplied. The root cause is in RestAuthentication: it validates the API key and stores the resulting user on the gin context, then reroutes the request from /rest/... to /plugins/... via gin's HandleContext. That call internally resets c.Keys, wiping the user that was just set, so the downstream RequireAuth middleware sees no authenticated user and rejects the request.

The fix stores the authenticated user in the underlying http.Request's context as well, since c.Request survives gin's context reset during the reroute. shared.GetUser now falls back to this request-context value when the gin context key is missing. A regression test spins up a real router with RestAuthentication + a RequireAuth-style gate + a webhook handler, and confirms a valid bearer-token request now reaches the handler (200) instead of being rejected (401).

Closes #9007

### Other Information

- [x] Disclose AI usage: AI-assisted
- [x] I have verified that the regression test fails without the fix and passes with the fix.

### DIFF
--- a/backend/server/api/middlewares.go
+++ b/backend/server/api/middlewares.go
@@ -33,6 +33,7 @@ import (
 	"github.com/apache/incubator-devlake/core/errors"
 	"github.com/apache/incubator-devlake/core/models/common"
 	"github.com/apache/incubator-devlake/helpers/apikeyhelper"
+	"github.com/apache/incubator-devlake/server/api/shared"
 	"github.com/gin-gonic/gin"
 )
 
@@ -272,9 +273,13 @@ func CheckAuthorizationHeader(c *gin.Context, logger log.Logger, db dal.Dal, api
 
 	logger.Info("redirect path: %s to: %s", c.Request.URL.Path, path)
 	c.Request.URL.Path = path
-	c.Set(common.USER, &common.User{
+	user := &common.User{
 		Name:  apiKey.Creator.Creator,
 		Email: apiKey.Creator.CreatorEmail,
-	})
+	}
+	c.Set(common.USER, user)
+	// Also store in the request context so the user survives gin's HandleContext
+	// resetting c.Keys when rerouting from /rest/... to /plugins/...
+	c.Request = shared.SetRestAuthUser(c.Request, user)
 	return true
 }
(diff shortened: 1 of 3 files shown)


klesh (approved):
LGTM.
Thanks for your contribution.

kamphaus:
I fixed the lint-commit-message check.

klesh (approved):

PR #9012, opened 2026-07-22
fix(sonarqube): tolerate missing migration indexes

### LINKED ISSUE #9011: [Bug][SonarQube] Migration fails when issue code block component index is absent
## Summary

A fresh MySQL installation can fail while applying the SonarQube and domain migrations at version `20260701000000` when the legacy component index is absent.

`Dalgorm.DropIndexes` unconditionally calls GORM's `DropIndex`, so MySQL aborts the migration with error 1091 instead of treating the already-absent index as the desired state.

## Reproduction

1. Start a fresh MySQL 8 database.
2. Create the legacy `cq_issue_code_blocks` and `_tool_sonarqube_issue_code_blocks` tables, then remove `idx_cq_issue_code_blocks_component` and `idx__tool_sonarqube_issue_code_blocks_component`.
3. Apply both version `20260701000000` migrations that change `component` to `text`.

Actual result: the first migration attempts to drop the absent index and fails with `Error 1091 (42000): Can't DROP 'idx_cq_issue_code_blocks_component'; check that column/key exists`.

Expected result: migration is idempotent when an index is already absent, while present indexes continue to be dropped before the column becomes `text`.

## Evidence

The targeted regression test `TestSonarqubeIssueCodeBlockLongComponent` reproduces the MySQL 1091 failure before the fix.

frederic-huebner:
### Summary

Make `Dalgorm.DropIndexes` idempotent by skipping absent indexes. This prevents MySQL error 1091 when the SonarQube and domain migrations at version `20260701000000` encounter legacy tables without their component indexes.

The regression test removes both legacy indexes before applying the two migrations, then verifies data preservation, `text` component columns, and absent indexes.

Closes #9011

### Other Information

Validated with the targeted MySQL-backed regression, the full SonarQube E2E package, `make lint`, and `make migration-script-lint`.

### DIFF
--- a/backend/impls/dalgorm/dalgorm.go
+++ b/backend/impls/dalgorm/dalgorm.go
@@ -453,6 +453,9 @@ func (d *Dalgorm) RenameTable(oldName, newName string) errors.Error {
 // DropIndexes drops indexes for specified table
 func (d *Dalgorm) DropIndexes(table string, indexNames ...string) errors.Error {
 	for _, indexName := range indexNames {
+		if !d.db.Migrator().HasIndex(table, indexName) {
+			continue
+		}
 		err := d.db.Migrator().DropIndex(table, indexName)
 		if err != nil {
 			return d.convertGormError(err)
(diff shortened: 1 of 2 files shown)


klesh (approved):
LGTM
Thanks for your contribution.

PR #9010, opened 2026-07-22
feat(jira): persist Sprint Report snapshot for accurate committed/com…

bramhanandlingala:
### Summary
Jira only tracked issues' current state, so committed/completed sprint velocity had to be approximated from resolution dates — which mis-attributes issues that span multiple sprints, causing errors in adjacent sprints.

### Changes Made
- New `_tool_jira_sprint_reports` table + collector/extractor — persists Jira's own frozen Sprint Report snapshot per closed sprint.
- Added `committed_story_point`/`completed_story_point` to the domain `sprints` table, populated by a new convertor.
- New "Sprint Velocity" panel on the Jira Grafana dashboard (mysql + postgresql).
- Updated `linear`/`zentao` e2e snapshot fixtures for the new shared domain columns.

### Does this close any open issues?
Closes #8967

Issue #8967, opened 2026-07-01
Feature: Persist Jira Sprint Report snapshot to enable accurate sprint velocity (committed & completed)


ashokobserve:
## Summary

DevLake ingests each Jira issue's **current** state (status, `resolution_date`, story points) plus whatever changelog it can collect, but it does **not** persist Jira's **Sprint Report snapshot** — the frozen record Jira takes at sprint close. As a result, per-sprint **committed** and **completed** velocity cannot be reproduced accurately from DevLake's domain tables, even though the numbers are correct in Jira's own Sprint/Velocity report.

I'd like to propose collecting and persisting the Greenhopper Sprint Report so velocity (commitment + completed) is exact and matches Jira's board.

## Why this matters

- **Velocity is a first-class Scrum metric.** Teams use DevLake dashboards to track committed vs completed story points per sprint. Today those numbers can only be *approximated*.
- **Commitment is not reconstructable at all** from current-state + changelogs. Commitment is the set of issues (and their SP) *as they stood when the sprint started* — a frozen snapshot DevLake never records.
- **Completed drifts per sprint due to carryover.** Jira attributes an issue's completion to the sprint where it was Done **at that sprint's close**. Reconstructing this from a single `resolution_date` mis-attributes carryover issues (issues living in ≥2 sprints), producing equal-and-opposite errors in adjacent sprints.

## Evidence

We validated a best-effort reconstruction (board-scoped `sprint_issues ∩ board_issues`, subtasks excluded, SP from the story-point custom field, completed = became-Done within the sprint window) against Jira's Sprint Report across 40 actively-used Scrum boards (208 closed sprints):

- **Aggregate accuracy: ~99.7%** (errors cancel across sprints).
- **Per-sprint accuracy varies sharply with carryover rate:**
  - Low-carryover boards (~0–4% of issues in multiple sprints): **100%** of sprints within 5%.
  - High-carryover boards (~25–29%): as low as **2 of 11** sprints within 5%.
- Root cause confirmed at the issue level: the same handful of carryover issues over-count sprint N and under-count sprint N+1, because their single `resolution_date` lands in the wrong sprint's window. Jira's Sprint Report places them correctly because it is a frozen per-sprint snapshot.

## How to reproduce the gap

1. Pick a board with meaningful carryover (issues that span >1 sprint).
2. For two adjacent closed sprints, compute completed SP from DevLake domain tables using `resolution_date` within the sprint window.
[...]

PR #9015, opened 2026-07-28
fix(jira): add missing _raw_data_* columns to _tool_jira_sprint_reports

DoDiODev:
> Starting point is the Jira Sprint Report bug below. The regression guard added
> for it turned out to catch the same class of bug in three more plugins
> (taiga, teambition, testmo), which are fixed here as well — see
> [§3](#3-additional-schema-drifts-found-by-the-new-cross-plugin-guard).

## Problem

Collecting Jira data fails in the `extractSprintReport` subtask with:

```
subtask extractSprintReport ended unexpectedly
  error getting batch from result (500)
  Error 1054 (42S22): Unknown column '_raw_data_table' in 'where clause'
```

### Root cause

The Sprint Report feature (PR #8967 / #9010) added the table
`_tool_jira_sprint_reports`. Its migration
(`20260722_add_sprint_report_table.go`) creates the table from a struct that
does **not** embed `common.NoPKModel`:

```go
type jiraSprintReport20260722 struct {
    ConnectionId uint64 `gorm:"primaryKey"`
    // ... no common.NoPKModel ...
}
```

The runtime model `models.JiraSprintReport` **does** embed `common.NoPKModel`
(→ `common.RawDataOrigin`), so it expects the columns `_raw_data_params`,
`_raw_data_table`, `_raw_data_id`, `_raw_data_remark` (plus `created_at`,
`updated_at`).

During extraction, `api.NewApiExtractor` deletes outdated rows via
`WHERE _raw_data_table = ? AND _raw_data_params = ?`
(`helpers/pluginhelper/api/batch_save_divider.go`). Because the migration never
created those columns, MySQL rejects the query with error 1054.

### 1. Fix for the reported bug (jira)
- **New migration** `plugins/jira/models/migrationscripts/20260727_add_raw_data_columns_to_sprint_report.go`
  re-runs `AutoMigrateTables` on a struct that embeds the raw-data columns,
  adding them to existing `_tool_jira_sprint_reports` tables **without data
  loss** (GORM `AutoMigrate` only adds missing columns). Registered in
  `models/migrationscripts/register.go`.
- The original `20260722_add_sprint_report_table.go` is left **unchanged**:
  migration scripts are append-only, and editing it would not repair databases
  that already ran it (its version is already recorded in
  `_devlake_migration_history`).

### 2. Regression tests (schema-drift guards)
- **`plugins/jira/e2e/migration_schema_test.go`** — runs the real Jira migration
  chain (framework + jira) and asserts every column each Jira model declares
  exists in the migrated table. Directly reproduces and guards the bug above.
- **`plugins/schema_e2e/migration_schema_test.go`** — cross-plugin generalization:
  applies framework + **all** plugin migrations and validates model-vs-table
  column parity for every built-in Go plugin. Includes `TestAllGoPluginsListed`,
  which fails if a new plugin directory with an `impl` package is added but not
  registered, so the guard stays complete automatically.

Both deliberately run the **real migration scripts** instead of `AutoMigrate`-ing
the runtime model — an `AutoMigrate`-based check could never detect this class of
drift. Both live in `e2e` packages, so they run under `make e2e-test-go-plugins`
(require `E2E_DB_URL`) and are excluded from the DB-less unit-test run.

Implementation details worth noting:
- Both tests call `dalgorm.Init(...)` to register the `encdec` GORM serializer.
  `runner.CreateBasicRes` does **not** do this (only `CreateAppBasicRes` does),
  so without it the migrations abort with `invalid serializer type encdec`.
- Both fall back to a deterministic `ENCRYPTION_SECRET` if none is configured;
  some migrations (e.g. jira `20220716`) refuse to run without one
  (`jira v0.11 invalid encKey`), and CI does not provide a value.
- Models that no migration materializes (pure API-response models such as
  `_tool_jira_server_infos`) are skipped — the check targets *drift* between an
  existing table and its model.

### 3. Additional schema drifts found by the new cross-plugin guard
The guard immediately uncovered three pre-existing bugs of exactly the same
class. Each is fixed with a new, additive migration (registered in the
respective `register.go`):

| Table | Missing column(s) |
| --- | --- |
| `_tool_taiga_scope_configs` | `type_mappings` |
| `_tool_teambition_scope_configs` | `id`, `created_at`, `updated_at` |
| `_tool_testmo_scope_configs` | `connection_id`, `name` |

Files: `plugins/{taiga,teambition,testmo}/models/migrationscripts/20260727_add_missing_scope_config_columns.go`.

All new migration scripts use `core/models/migrationscripts/archived` — importing
`core/models/common` from a migration script is rejected by
`make migration-script-lint`.

#### Why these repairs are safe on populated tables

The scope-config repairs re-add columns that carry constraints in
`common.ScopeConfig` (`name` has a `uniqueIndex`) and, for teambition, an
AUTO_INCREMENT primary key. Both were verified against live engines rather than
assumed:

- **`uniqueIndex` on `name`** — GORM adds new columns as *nullable*
  (`ALTER TABLE ... ADD COLUMN name VARCHAR(255)`, no `NOT NULL DEFAULT ''`), so
  pre-existing rows are backfilled with `NULL`, and both MySQL and PostgreSQL
  permit duplicate `NULL`s in a unique index. Creating the index on a table with
  two pre-existing rows succeeded on **MySQL 8.4.10** and **PostgreSQL 17.2**.
  The counter-test confirms the distinction: with an explicit
  `NOT NULL DEFAULT ''` column the same index fails
  (`Error 1062` / `pq: … (23505)`) — which is exactly the case that does *not*
  occur here.
- **AUTO_INCREMENT `id` on the primary-key-less `_tool_teambition_scope_configs`** —
  GORM issues a plain `ADD COLUMN`, and MySQL backfills consecutive ids for the
  existing rows (verified: two pre-existing rows received `id` 1 and 2). No
  `Error 1075` ("there can be only one auto column and it must be defined as a
  key").

### 4. Build script
`scripts/build-plugins.sh` builds every directory under `plugins/` with
`-buildmode=plugin`. `plugins/schema_e2e/` is not a plugin (it contains only the
cross-plugin test and has no `main` package), which made `make build-plugin` fail
with *"-buildmode=plugin requires exactly one main package"*. The directory is now
excluded, alongside the existing `core` / `helper` / `logs` exclusions.

## Why a new migration (not editing the old one)

- The buggy migration `20260722` is already on `main`/`upstream/main` and has
  already been applied to real databases; its version is recorded, so it will
  never re-run. Only a **new** migration can repair those databases.
- Append-only migrations keep the migration history reproducible across
  environments.

## Testing

- `make migration-script-lint` — OK.
- `gofmt -l` on all added/changed files — clean.
- `go vet ./plugins/{jira,taiga,teambition,testmo,schema_e2e}/...` — OK.
- **MySQL 8.4.10, fresh database:**
  - `go test ./plugins/schema_e2e/` — OK, **42/42 plugins PASS**
    (`TestAllGoPluginsListed` included).
  - `go test -run TestMigrationSchema ./plugins/jira/e2e/` — OK.
- **PostgreSQL 17.2, fresh database:** both tests OK; the repaired scope-config
  tables contain the added columns.
- **Upgrade path on a populated database** (not just a fresh one): all four
  migrations were applied to a real, long-running DevLake instance. They are
  recorded in `_devlake_migration_history` and the target tables carry the
  expected columns afterwards — e.g. `_tool_teambition_scope_configs` gained
  `id`, `created_at`, `updated_at`, and `_tool_jira_sprint_reports` (661 rows)
  gained the `_raw_data_*` columns without data loss.
- **Constraint safety** on populated tables (unique index / AUTO_INCREMENT PK)
  verified separately against MySQL 8.4.10 and PostgreSQL 17.2 — see
  [§3 "Why these repairs are safe on populated tables"](#why-these-repairs-are-safe-on-populated-tables).
- **Negative test:** dropping `_raw_data_table` from `_tool_jira_sprint_reports`
  makes the cross-plugin guard fail with
  `[jira] table "_tool_jira_sprint_reports" is missing column "_raw_data_table"`
  — i.e. the guard genuinely detects the original regression.

## Known limitations of the guard (intentional)

- It checks **column presence only** — not column type, length, nullability,
  primary keys or indexes.
- It runs against the shared E2E database. Other plugin E2E tests use
  `FlushTabler` (drop + `AutoMigrate` of the runtime model), so a table touched by
  an earlier test can appear "repaired". On a fresh database (as in CI) this does
  not apply.
- Tables that no migration creates are skipped, so "model listed in
  `GetTablesInfo()` but no table at all" is not flagged.


klesh:
Hi. Could you please take a look at the failing test cases. Thanks.

vbhanuchander-lang:
Nice piece of work — particularly leaving `20260722_add_sprint_report_table.go` untouched and repairing forward with an additive migration. That's the right call for anything already recorded in `_devlake_migration_history`, and it's the part people most often get wrong.

One gap worth considering in the new guard, since it's the most interesting part of this PR.

`migration_schema_test.go` runs the real migration chain against a fresh database and compares the resulting columns to the model. That catches drift in the *end state*, which is the bug you started from. But it runs against empty tables, so it cannot exercise the populated-table upgrade path — and that path is exactly where the teambition workaround lives:

> `id` is an auto-increment primary key, which GORM's AutoMigrate cannot append to an existing table ... The column is therefore added with explicit DDL

On an empty `_tool_teambition_scope_configs`, plain `AutoMigrate` and the explicit DDL produce an identical schema, so the guard stays green either way. If someone later simplifies that DDL back to `AutoMigrate` because it looks redundant, the guard still passes and only real upgrades break — the same failure shape as the bug this PR fixes, one layer up.

Would it be worth inserting a row into each table before the tail migrations run, so the guard also asserts the migration *applies* to a non-empty table? Even a single row would have made the `id` case fail loudly rather than silently.

Not a blocker — the fix itself looks correct on both engines to me. `Dialect()` returns `Dialector.Name()`, so the `"mysql"` comparison matches, and `BIGSERIAL PRIMARY KEY` is the right Postgres counterpart.

DoDiODev:
Thanks — I checked this against both engines rather than reasoning about it, and the specific example turns out to be already covered, for a slightly different reason than expected.

On **MySQL 8.4.10**, replacing the explicit DDL with plain `AutoMigrate` emits `ALTER TABLE ... ADD id bigint unsigned AUTO_INCREMENT` and fails with `Error 1075 (42000): Incorrect table definition; there can be only one auto column and it must be defined as a key` **regardless of row count** — it is a DDL-level rejection, so the guard fails on the empty table too (verified with 0 and with 2 pre-existing rows).

On **PostgreSQL 17.2** the same `AutoMigrate` succeeds in both cases (`ALTER TABLE ... ADD "id" bigserial`), just without the PRIMARY KEY constraint — so a seeded row would not surface it there either. What is invisible on Postgres is the missing primary key, and a column-presence guard cannot see that with or without data.

Your general point stands, though: data-dependent failures are not exercised — `NOT NULL` without a default, a `uniqueIndex` over pre-existing duplicates, type narrowing. A fully generic "seed every table before the tail migrations run" would need a two-phase migrator run with an arbitrary version cutoff plus valid rows for every tool table (JSON columns, NOT NULL fields), which ages badly.

What I am adding instead is a targeted upgrade-path guard for the tables this PR repairs: build the pre-repair table shape, insert rows, run only the new script's `Up()`, then assert the added columns exist, the rows survived, and the ids were backfilled — plus a primary-key assertion, so the `AutoMigrate`-instead-of-DDL case is caught on Postgres as well. That codifies exactly what is currently only documented as manual verification.

Thanks for the careful read — the `Dialect()` / `BIGSERIAL` confirmation is appreciated.

DoDiODev:
Implemented as announced, plus the CI question from @klesh is now answered — both in one update.

## 1. Targeted upgrade-path guard on **populated** tables

New: `backend/plugins/schema_e2e/migration_upgrade_path_test.go` →
`TestMigrationUpgradePathOnPopulatedTables`.

For every repair migration in this PR it

1. recreates the table **exactly as the buggy migration left it**,
2. inserts rows,
3. runs **only that repair script's** `Up()` (looked up through the plugin's own `MigrationScripts()` **by version**, so the test breaks if a script is removed or renumbered),
4. asserts: the columns were added, the rows survived, the table has a primary key, auto-increment ids were backfilled with distinct non-zero values, and a subsequent `INSERT` still works (i.e. the sequence/counter is in sync).

Covered (7 cases): `_tool_jira_sprint_reports`, `_tool_taiga_scope_configs`, `_tool_teambition_scope_configs`, `_tool_testmo_scope_configs` and the three `_tool_copilot_{enterprise,org,user}_ai_credit_usage` tables.

### Negative test — it really does catch the regression

Replacing the explicit `AUTO_INCREMENT` DDL in the teambition script with a plain `AutoMigrate` (exactly the "someone simplifies this later" scenario @vbhanuchander-lang described):

| engine | result |
| --- | --- |
| MySQL 8.4.10 | **FAIL** — `migration "add missing id/created_at/updated_at columns …" failed on a populated "_tool_teambition_scope_configs"` (Error 1075) |
| PostgreSQL 17.2 | **FAIL** — `table "_tool_teambition_scope_configs" has no primary key after …` |

The Postgres case is the one the column-presence guard cannot see: `AutoMigrate` happily adds `bigserial` *without* a key and reports success. That gap is now closed by the primary-key assertion.

## 2. Rebase + a fourth instance of the same bug class

Rebased onto current `main` (10 upstream commits). The guard immediately failed for `gh-copilot`: the three tables added by #9019 are missing all seven credit-breakdown columns.

| table | missing |
| --- | --- |
| `_tool_copilot_enterprise_ai_credit_usage` | `gross_quantity`, `discount_quantity`, `net_quantity`, `price_per_unit`, `gross_amount`, `discount_amount`, `net_amount` |
| `_tool_copilot_org_ai_credit_usage` | same 7 |
| `_tool_copilot_user_ai_credit_usage` | same 7 |

Root cause: `20260708_add_ai_credit_usage_metrics.go` declares them via an anonymous embedded struct whose **type name is unexported**

```go
creditUsageBreakdown20260708 `gorm:"embedded"`
```

GORM's schema parser skips anonymous fields of unexported types, so `AutoMigrate` never created the columns — while the runtime models declare them inline. The extractor would fail with `Unknown column 'gross_quantity' in 'field list'`: the Jira Sprint Report failure mode, one plugin over.

Fixed the same way as the other three — a new, additive migration (`20260731_fix_ai_credit_usage_breakdown_columns.go`); the original script is untouched. The two new plugins (`clickup`, `incidentio`) were registered in `allGoPlugins()`, as `TestAllGoPluginsListed` demanded — the self-completing property working as intended.

## 3. CI

The workflows on this PR sit at `action_required` (fork PR awaiting maintainer approval), so I mirrored the same checks in my fork on an identical tree — **all green**:

- `lint (go)`, `migration-script-lint`, `unit-test`, `config-ui`, ASF header check, Grafana dashboard check
- `e2e (mysql)` (`make e2e-test-go-plugins` + `make e2e-test`), including:
  - `--- PASS: TestMigrationSchema` (Jira-specific guard)
  - `--- PASS: TestAllGoPluginsListed`
  - `--- PASS: TestMigrationSchemaMatchesModels` — **44/44 plugins**
  - `--- PASS: TestMigrationUpgradePathOnPopulatedTables` — **7/7 cases**

Locally also verified against **MySQL 8.4.10** and **PostgreSQL 17.2**: 51/51 subtests pass on both.

@klesh this should address the failing checks — happy to have the workflows approved on the PR itself to confirm on your infrastructure.

klesh (approved):
LGTM
Thanks for your contribution.

Issue #8590, opened 2025-09-23
[Bug][Backend] Teams and Users API failing in 1.0.2


anshulverma46:
### Search before asking

- [ ] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

We are running Devlake using docker compose with image tags pinned to 1.0.2. It was on the older version before where the Teams and User API worked the first time. We updated the version few weeks back and this is the first time we are trying on 1.0.2 but the error is consistent as below:

```
2025/09/23 14:54:20 [Recovery] 2025/09/23 - 14:54:20 panic recovered:
PUT /rest/plugins/org/teams.csv HTTP/1.1
Host: devlake:8080
Accept: */*
Authorization: *
Content-Length: 566
Content-Type: multipart/form-data; boundary=------------------------8621d9a30c86a953
User-Agent: curl/7.81.0


runtime error: invalid memory address or nil pointer dereference
/usr/local/go/src/runtime/panic.go:260 (0x9b1ddc)
/usr/local/go/src/runtime/signal_unix.go:841 (0x9b1dac)
/app/plugins/org/api/team.go:88 (0x7dde2937baa4)
/app/server/api/router.go:141 (0x16179ce)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x16157e2)
/app/server/api/middlewares.go:95 (0x16157c6)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x1615c87)
/app/server/api/middlewares.go:117 (0x1615b22)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x1255301)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/recovery.go:102 (0x12552ec)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x1254426)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/logger.go:240 (0x1254409)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x12532aa)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/gin.go:620 (0x1252f31)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/gin.go:587 (0x1252c04)
[...]

Issue #8271, opened 2025-01-17
[Bug][API] Team API is not working in k8s deployment


AnkeshThakur:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

Team and user API is not working and not able to fill the teams and user data. My deployment is done on kubernetes and all the values filled up as per the documentation. All other functionalities are working as expected but teams and users data api is not working and responding 500. 

Here is the log from the container:

```
time="2025-01-17 09:34:28" level=info msg="redirect path: /rest/plugins/org/teams.csv to: /plugins/org/teams.csv"
2025/01/17 09:34:28 [Recovery] 2025/01/17 - 09:34:28 panic recovered:
PUT /rest/plugins/org/teams.csv HTTP/1.1
Host: devlake-app-lake.devlake.svc.cluster.local:8080
Accept: */*
Authorization: *
Content-Length: 680
Content-Type: multipart/form-data; boundary=------------------------nsDOmxfQmFYRaQW9NPt1y2
User-Agent: curl/8.7.1
X-Forwarded-For: 10.221.49.109
X-Forwarded-Host: devlake.development.********.com
X-Forwarded-Port: 80
X-Forwarded-Proto: http
X-Forwarded-Scheme: http
X-Real-Ip: 10.221.49.109
X-Request-Id: 7815a1e4196e533f483b93bed76ace61
X-Scheme: http
runtime error: invalid memory address or nil pointer dereference
/usr/local/go/src/runtime/panic.go:260 (0x9ae95c)
/usr/local/go/src/runtime/signal_unix.go:837 (0x9ae92c)
/app/plugins/org/api/team.go:88 (0x7f73a56c78c4)
/app/server/api/router.go:141 (0x161258e)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x16103a2)
/app/server/api/middlewares.go:95 (0x1610386)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x1610847)
/app/server/api/middlewares.go:117 (0x16106e2)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x1251da1)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/recovery.go:102 (0x1251d8c)
[...]

Issue #7957, opened 2024-08-26
[Bug][API] Team configuration fail to update using API


AlexPerkbox:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

Following [Team Configuration](https://devlake.apache.org/docs/Configuration/TeamConfiguration/) setup, `teams`, `users`, and `team_users` tables fail to populate when trying to upload CSV files using cURL/postman. No error is returned other than 500 internal error, managed to collect some logs from the AWS ECS instance, see below. Mysql is running inside AWS RDS

(older events at the bottom)
```
10.39.45.25 - - [26/Aug/2024:09:09:45 +0000] "PUT /api/rest/plugins/org/teams.csv HTTP/1.1" 500 0 "-" "curl/8.7.1" "10.1.170.32"
[31m2024/08/26 09:09:45 [Recovery] 2024/08/26 - 09:09:45 panic recovered:
runtime error: invalid memory address or nil pointer dereference
/usr/local/go/src/runtime/panic.go:260 (0x9a295c)
/usr/local/go/src/runtime/signal_unix.go:837 (0x9a292c)
/app/plugins/org/api/team.go:88 (0x7fc331b24584)
/app/server/api/router.go:141 (0x15ec88e)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x15ea6a2)
/app/server/api/middlewares.go:95 (0x15ea686)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x15eab47)
/app/server/api/middlewares.go:117 (0x15ea9e2)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x123c5a1)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/recovery.go:102 (0x123c58c)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x123b6c6)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/logger.go:240 (0x123b6a9)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x123a54a)
/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/gin.go:620 (0x123a1d1)
[...]

Issue #7219, opened 2024-03-26
[Bug][user.go] Uploading users.csv fails with run time error 


yacc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

Using curl to upload the users.csv file with the template users.csv file downloaded from Devlake is generating a run time error.

### What do you expect to happen

I expected to have the data from the users.csv file to be uploaded in the db

### How to reproduce

Running 
```
curl 'http://127.0.0.1:4000/api/rest/plugins/org/users.csv' -X 'PUT' -H "Authorization: Bearer $DEVLAKE_API_KEY" --form 'file=@"./users.csv"'
```
with 
```
➜  devlake cat users.csv
Id,Name,Email,TeamIds
1,Tyrone K. Cummings,TyroneKCummings@teleworm.us,1;2
2,Dorothy R. Updegraff,DorothyRUpdegraff@dayrep.com,3
```

### Anything else

Here's the logs:
```log
2024-03-26 10:23:09 [GIN] 2024/03/26 - 09:23:09 | 500 |    8.590084ms |      172.18.0.5 | PUT      "/rest/plugins/org/users.csv"
2024-03-26 10:23:09 
2024-03-26 10:23:09 
2024-03-26 10:23:09 2024/03/26 09:23:09 [Recovery] 2024/03/26 - 09:23:09 panic recovered:
2024-03-26 10:23:09 runtime error: invalid memory address or nil pointer dereference
2024-03-26 10:23:09 /usr/local/go/src/runtime/panic.go:260 (0x9955e7)
2024-03-26 10:23:09 /usr/local/go/src/runtime/signal_unix.go:837 (0x9ad803)
2024-03-26 10:23:09 /app/plugins/org/api/user.go:91 (0xffff46fa2328)
2024-03-26 10:23:09 /app/server/api/router.go:135 (0x149d547)
2024-03-26 10:23:09 /go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x149b8cb)
2024-03-26 10:23:09 /app/server/api/middlewares.go:95 (0x149b8a8)
2024-03-26 10:23:09 /go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x149bcfb)
2024-03-26 10:23:09 /app/server/api/middlewares.go:117 (0x149bbd4)
[...]

PR #9022, opened 2026-07-29
fix(org): resolve API handlers at request time to survive late plugin init

### LINKED ISSUE #9021: [Bug][org] All org plugin API endpoints panic with nil pointer — root cause: routes bind handlers before InitPlugins runs
## What happened

Every `plugins/org` API endpoint (`teams.csv`, `users.csv`, `user_account_mapping.csv`, GET and PUT) returns HTTP 500 with a nil-pointer panic. Verified on v1.0.2 and v1.0.3-beta15 (helm deployment, MySQL backend). GET with `?fake_data=true` works (it never touches the store).

Panic (v1.0.3-beta15, `PUT /plugins/org/teams.csv`):

```
runtime error: invalid memory address or nil pointer dereference
/app/plugins/org/api/team.go:87   <- h.store.deleteAll(&crossdomain.Team{})
/app/server/api/router.go:149
```

GET without fake_data panics at `team.go:47` (`h.store.findAllTeams()`).

## Root cause (with evidence)

It's an initialization-order race between route registration and plugin initialization:

1. `runner.LoadGoPlugins` only does `Lookup("PluginEntry")` + `plugin.RegisterPlugin(...)` — it does **not** call `PluginInit.Init()` ([loader.go](https://github.com/apache/devlake/blob/v1.0.3-beta15/backend/core/runner/loader.go#L62-L83)).
2. `plugin.InitPlugins(basicRes)` — the only place `Init()` runs in server mode — is invoked from `pipelineServiceInit()` ([pipeline.go:83](https://github.com/apache/devlake/blob/v1.0.3-beta15/backend/server/services/pipeline.go#L83)).
3. `registerPluginEndpoints` builds the router from each plugin's `ApiResources()`. For `org`, that returns **bound method values** on instance state: `p.handlers.GetTeam`, etc. ([impl.go](https://github.com/apache/devlake/blob/v1.0.3-beta15/backend/plugins/org/impl/impl.go#L104-L120)).
4. In Go, a method value captures its receiver **at evaluation time**. If `ApiResources()` is evaluated before `Init()` set `p.handlers`, the router permanently holds handlers bound to a nil `*Handlers` — running `Init()` later doesn't rebind them.

Startup timeline from our pod logs proving the order (a DB migration was pending at boot, which delays `pipelineServiceInit` while the HTTP server must start to serve `/proceed-db-migration`):

```
12:13:58  plugin loaded org
12:14:11  endpoint PUT /plugins/org/teams.csv ... registerPluginEndpoints   <- ApiResources() evaluated, p.handlers == nil
12:14:23  plugin: <X> doesn't implement 'PluginInit', it will be skipped.   <- InitPlugins runs, 12s too late
```

`org` is the only plugin whose `ApiResources()` binds per-instance state created in `Init()` — other plugins read package-level vars at call time — which is why only `org` breaks, and why it breaks non-deterministically across deployments (whoever wins the race). This likely explains the long trail of closed-as-stale reports: #8590, #8271, #7957, #7219 (different symptoms of the same race — reporters couldn't pin it because it depends on startup timing; a pending DB migration at boot makes it deterministic).

Note the v1.0.3-beta fix `(&team{}).toDomainLayer(tt)` (team.go:84) resolved a *second, independent* nil deref, which made the remaining one harder to spot.

`plugins/org` CSV endpoints work regardless of startup timing.

## How to reproduce

1. Deploy v1.0.2 or v1.0.3-beta15 (helm chart, MySQL) with a pending DB migration at boot (e.g. any version upgrade), or any startup where route registration beats `pipelineServiceInit`.
2. Confirm the migration, then `curl -X PUT <endpoint>/plugins/org/teams.csv -F 'file=@teams.csv'` → HTTP 500, panic above. `GET /plugins/org/teams.csv` → same at team.go:47.

## Proposed fix (happy to submit a PR)

Two options, smallest first:

**A (org-scoped, minimal):** make `org`'s `ApiResources()` resolve handlers at call time instead of bind time — closures over `p` (pointer receiver) rather than method values over `p.handlers`:

```go
func (p *Org) ApiResources() map[string]map[string]plugin.ApiResourceHandler {
    wrap := func(f func(*api.Handlers) plugin.ApiResourceHandler) plugin.ApiResourceHandler {
        return func(in *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
            if p.handlers == nil {
                return nil, errors.Internal.New("org plugin not initialized yet")
            }
            return f(p.handlers)(in)
        }
    }
    ...
}
```

**B (framework):** guarantee `plugin.InitPlugins()` runs before `registerPluginEndpoints` (e.g. right after `runner.LoadPlugins` in `services.Init`), decoupling it from `pipelineServiceInit`. Fixes the class of bug for any future plugin with instance state, but touches startup ordering for everything.

## Version

v1.0.3-beta15 (also reproduced on v1.0.2). Helm chart on Kubernetes (EKS), external MySQL 8.0.

Yes.

pballester:
### Summary

All `plugins/org` API endpoints (`teams.csv`, `users.csv`, `user_account_mapping.csv`, `project_mapping.csv`) panic with a nil pointer dereference whenever route registration wins the race against plugin initialization — full diagnosis with startup-log evidence in #9021.

`InitPlugins()` (the only thing that calls `PluginInit.Init()` in server mode) runs inside `pipelineServiceInit()`, while `registerPluginEndpoints` evaluates each plugin's `ApiResources()` independently. The org plugin returned **bound method values** (`p.handlers.GetTeam`), and Go captures the receiver at evaluation time — if `Init()` hasn't run yet, the router permanently holds handlers bound to a nil `*Handlers`. A pending DB migration at boot makes the bad ordering deterministic, which is why this bit several users after upgrades (#8590, #8271, #7957, #7219).

This fix resolves `p.handlers` at **request time** via method expressions, returning a clean 500 with "org plugin is not initialized yet, please retry later" during the (brief) pre-init window instead of panicking forever after it.

- `go build` / `go vet` / `gofmt` clean on the touched packages
- New unit test `TestApiResourcesBeforeInitFailsGracefully` covers every org endpoint pre-init (panicked before this change, clean error now)
- e2e tests unchanged (require `E2E_DB_URL`)

A broader alternative would be moving `InitPlugins()` ahead of route registration in server startup — deliberately not done here to keep the change scoped to the affected plugin; happy to follow up if maintainers prefer that direction.

Closes #9021

### DIFF
--- a/backend/plugins/org/impl/impl.go
+++ b/backend/plugins/org/impl/impl.go
@@ -102,24 +102,41 @@ func (p Org) RootPkgPath() string {
 	return "github.com/apache/incubator-devlake/plugins/org"
 }
 
-func (p Org) ApiResources() map[string]map[string]plugin.ApiResourceHandler {
+// wrapHandler defers the resolution of p.handlers to request time.
+// ApiResources() may be evaluated during route registration, which can happen
+// before InitPlugins() has called Init() (InitPlugins runs inside
+// pipelineServiceInit); a bound method value like p.handlers.GetTeam would
+// capture a nil receiver permanently, making every org endpoint panic with a
+// nil pointer dereference. See https://github.com/apache/devlake/issues/9021.
+func (p *Org) wrapHandler(
+	method func(*api.Handlers, *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error),
+) plugin.ApiResourceHandler {
+	return func(input *plugin.ApiResourceInput) (*plugin.ApiResourceOutput, errors.Error) {
+		if p.handlers == nil {
+			return nil, errors.Internal.New("org plugin is not initialized yet, please retry later")
+		}
+		return method(p.handlers, input)
+	}
+}
+
+func (p *Org) ApiResources() map[string]map[string]plugin.ApiResourceHandler {
 	return map[string]map[string]plugin.ApiResourceHandler{
 		"teams.csv": {
-			"GET": p.handlers.GetTeam,
-			"PUT": p.handlers.CreateTeam,
+			"GET": p.wrapHandler((*api.Handlers).GetTeam),
+			"PUT": p.wrapHandler((*api.Handlers).CreateTeam),
 		},
 		"users.csv": {
-			"GET": p.handlers.GetUser,
-			"PUT": p.handlers.CreateUser,
+			"GET": p.wrapHandler((*api.Handlers).GetUser),
+			"PUT": p.wrapHandler((*api.Handlers).CreateUser),
 		},
 
 		"user_account_mapping.csv": {
-			"GET": p.handlers.GetUserAccountMapping,
-			"PUT": p.handlers.CreateUserAccountMapping,
+			"GET": p.wrapHandler((*api.Handlers).GetUserAccountMapping),
+			"PUT": p.wrapHandler((*api.Handlers).CreateUserAccountMapping),
 		},
 		"project_mapping.csv": {
-			"GET": p.handlers.GetProjectMapping,
-			"PUT": p.handlers.CreateProjectMapping,
+			"GET": p.wrapHandler((*api.Handlers).GetProjectMapping),
+			"PUT": p.wrapHandler((*api.Handlers).CreateProjectMapping),
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


klesh (approved):
LGTM.
Thanks for your contribution.

PR #9026, opened 2026-08-01
feat(jira): add JIRA_SKIP_UNPARSEABLE_ISSUES to survive unparseable issue pages

### LINKED ISSUE #8949: [Feature][jira] Add JIRA_SKIP_UNPARSEABLE_ISSUES to continue collection when an issue page can't be parsed

vbhanuchander-lang:
### Summary

A Jira search page that comes back with a successful status but a body that is not the expected JSON — a truncated payload, or an HTML error page substituted by a proxy — currently fails the whole `collectIssues` subtask. On a large Jira instance that throws away an otherwise complete sync because of a single bad page, typically during deep pagination with changelog expansion.

This adds `JIRA_SKIP_UNPARSEABLE_ISSUES`:

- **Default `false`** — unchanged behaviour: the error surfaces and the subtask fails.
- **`true`** — the offending page is logged with its URL and body size, then skipped, and collection continues.

The flag is read through the same `taskCtx.GetConfigReader().GetBool(...)` pattern that `JIRA_JQL_AUTO_FULL_REFRESH` already uses in this plugin, and is logged once at the start of collection when enabled.

Only parse failures on otherwise-successful responses are affected. Non-2xx statuses never reach the response parser and keep their existing retry behaviour, as the issue requested.

The V2 (`api/2/search`) and V3 (`api/3/search/jql`) response parsers were byte-identical, so rather than duplicating the change they now share one `parseIssuesResponse` helper — a net reduction in duplicated code. A skipped page returns an empty non-nil slice, so the collector records "this page had no rows" instead of treating the result as absent.

Closes #8949

### Screenshots

N/A — behavioural change with no UI surface.

### Other Information

**Tests.** Nine cases added to `issue_collector_test.go`, following the existing table-driven style: valid pages under both flag states, the HTML-error-page and truncated-body variants under both flag states, a valid page with no `issues` key, and the empty-vs-nil distinction on the skip path. Plus a small `responseUrl` test for nil-safety.

One wrinkle worth flagging: `unithelper.DummyLogger()` stubs `Warn` with two arguments, but the real signature is `Warn(err, format, a ...interface{})`, which mockery records as three. The test adds the variadic stub locally rather than changing the shared helper or making the production call less idiomatic — variadic `logger.Warn` is used widely across the plugins.

**Verification.** `go test ./plugins/jira/...` passes (`plugins/jira/tasks` ok), `gofmt` clean, `go vet` clean for the changed files. Across the repo's unit gate, the only failures I see locally are the `gitextractor` packages failing to *build* because `pkg-config`/libgit2 isn't installed on my machine — unrelated to this change. `e2e` is excluded from the unit gate and needs `E2E_DB_URL`.

**Documentation.** I have not added docs here, matching the existing precedent: `JIRA_JQL_AUTO_FULL_REFRESH` and the other plugin-level flags are not documented in this repo, and `env.example` carries only global settings. Happy to send a companion PR to the website repo if that is the preferred home for it.

**Scope question for maintainers.** I have implemented this in the Jira plugin as the issue asks, but the same "2xx with an unparseable body" failure can affect any collector. If you would rather see it handled generically in the shared collector helper, I am glad to rework it.

**Coordination.** @Murad-Suleymanov opened the issue and mentioned they were willing to submit a PR. I commented on the issue before opening this and will happily close it if they already have work in progress.

### DIFF
--- a/backend/plugins/jira/tasks/issue_collector.go
+++ b/backend/plugins/jira/tasks/issue_collector.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/apache/incubator-devlake/core/dal"
 	"github.com/apache/incubator-devlake/core/errors"
+	"github.com/apache/incubator-devlake/core/log"
 	"github.com/apache/incubator-devlake/core/plugin"
 	"github.com/apache/incubator-devlake/helpers/pluginhelper/api"
 	"github.com/apache/incubator-devlake/plugins/jira/models"
@@ -95,12 +96,17 @@ func CollectIssues(taskCtx plugin.SubTaskContext) errors.Error {
 		pageSize = 100
 	}
 
+	skipUnparseable := taskCtx.GetConfigReader().GetBool("JIRA_SKIP_UNPARSEABLE_ISSUES")
+	if skipUnparseable {
+		logger.Info("JIRA_SKIP_UNPARSEABLE_ISSUES is enabled, unparseable issue pages will be skipped")
+	}
+
 	if strings.EqualFold(string(data.JiraServerInfo.DeploymentType), string(models.DeploymentServer)) {
 		logger.Info("Using api/2/search for JIRA Server issue collection")
-		err = setupIssueV2Collector(apiCollector, data, filterJql, pageSize)
+		err = setupIssueV2Collector(apiCollector, data, filterJql, pageSize, skipUnparseable, logger)
 	} else {
 		logger.Info("Using api/3/search/jql for JIRA Cloud issue collection")
-		err = setupIssueV3Collector(apiCollector, data, filterJql, pageSize)
+		err = setupIssueV3Collector(apiCollector, data, filterJql, pageSize, skipUnparseable, logger)
 	}
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


klesh:
It makes sense to me. I wonder if the `http code` could be used for the purpose? What status code does it return?
Please update the `env.example` to add the variable with comment to explain how it works.

vbhanuchander-lang:
Thanks @klesh — I've pushed the `env.example` entry with a comment explaining the behaviour.

On the status code: unfortunately it can't distinguish these, because the failing pages arrive with a *successful* status. `HttpMinStatusRetryCode` is `http.StatusBadRequest` (`helpers/pluginhelper/api/api_async_client.go`), so anything >= 400 is retried and then surfaced as an error before `ResponseParser` is ever called. By the time the parser runs, the response is already < 400 — typically a plain 200 whose body is a truncated payload or an HTML error page substituted by a proxy. The status line says success; only the body reveals the problem, which is why the check has to sit at the parse step rather than on the status code.

That is also why the flag is deliberately narrow: HTTP error statuses keep their existing retry behaviour and are unaffected by it.

klesh:
> Thanks @klesh — I've pushed the `env.example` entry with a comment explaining the behaviour.
> 
> On the status code: unfortunately it can't distinguish these, because the failing pages arrive with a _successful_ status. `HttpMinStatusRetryCode` is `http.StatusBadRequest` (`helpers/pluginhelper/api/api_async_client.go`), so anything >= 400 is retried and then surfaced as an error before `ResponseParser` is ever called. By the time the parser runs, the response is already < 400 — typically a plain 200 whose body is a truncated payload or an HTML error page substituted by a proxy. The status line says success; only the body reveals the problem, which is why the check has to sit at the parse step rather than on the status code.
> 
> That is also why the flag is deliberately narrow: HTTP error statuses keep their existing retry behaviour and are unaffected by it.

Thanks for the explanation.  I appreciate it.

klesh (approved):
LGTM

PR #8875, opened 2026-05-13
chore: upgrade go to 1.26

dncrews:
### Summary

Upgrade the backend Go toolchain to `1.26` and update related CI/tooling compatibility.

- Update Go versions in `go.mod`, Dockerfiles, devcontainer, and lake-builder image.
- Migrate `golangci-lint` to v2 and update the GitHub Action.
- Update `mockery` for Go 1.26 compatibility.
- Replace deprecated Go APIs flagged by the updated lint setup.
- Add local E2E setup notes, including the `libgit2` rpath requirement.
- Update stale `actions/*` workflow versions.

### Does this close any open issues?

No, but there are some related issues that are closed. See below.

### Screenshots

N/A

### Other Information

Follows up on the earlier Go upgrade work in #8599 by @petkostas.

Verified locally:

- `make go-dep`
- `make mock`
- `make lint`
- targeted Go tests for changed packages
- `go run ./test/init.go`
- E2E remote/services tests
- `make unit-test-go`
- workflow YAML parse check
- `git diff --check`

Suggested label: `pr-type/feature-development` or `pr-type/maintenance`

PR #9028, opened 2026-08-03
build(deps): bump libgit2 1.3 → 1.5.0 and git2go v33 → v34

DoDiODev:
## Summary

Bumps the Go binding `libgit2/git2go` from **v33.0.6** to **v34.0.0** and the
native **libgit2** library from **1.3.x** to **1.5.0** at every install site.

`git2go` majors are bound to exactly one libgit2 minor via the C ABI — there is
no supported version range. The pair must therefore be pinned in lockstep
everywhere, which is what this PR does.

Current pinning (before this PR) is 3.5 years old: libgit2 1.3.0 was released in
Sep 2021, 1.5.0 in Jul 2022 and carries bug/security fixes of the 1.4 and 1.5
series (including several `CVE`-relevant fixes in the smart-transport/packfile
code paths).

## Why v34 / 1.5.0 and not something newer?

| libgit2 | git2go | Import path | Status |
|---------|--------|-------------|--------|
| 1.5.0   | v34    | `github.com/libgit2/git2go/v34` | **target — highest importable pair** |
| 1.3.x   | v33    | `github.com/libgit2/git2go/v33` | current / documented rollback |
| 1.5.0   | ~~v35~~ | — | tag exists, but its `go.mod` still declares the module path `.../git2go/v34` → **not `go get`-able** |
| ~~1.6–1.9~~ | — | — | **no** git2go release exists for these |

So `v34.0.0` + libgit2 `1.5.0` is the ceiling that the canonical, upstream
`libgit2/git2go` module allows today (verified again on 2026-08-03: the tag list
is unchanged, `v35.0.0` still points at the same broken commit as `v34.0.0`).

**No Go code changes are required:** the git2go Go API is identical between
v33.0.6 and v34.0.0 for every call used by `gitextractor`
(`DiffOptions.NotifyCallback`, `Odb.ForEach`, `NewBranchIterator`, `Branch.*`,
`Reference.Target`, `DiffLine.Origin`, `DefaultDiffOptions`). The major bump only
reflects the C ABI/libgit2 version.

## What changed

| File | Change |
|---|---|
| `backend/go.mod` / `backend/go.sum` | `git2go/v33 v33.0.6` → `git2go/v34 v34.0.0` (`go mod tidy`) |
| `backend/plugins/gitextractor/parser/repo_libgit2.go` | import path `/v33` → `/v34` |
| `backend/plugins/gitextractor/models/diffline.go` | import path `/v33` → `/v34` |
| `backend/Dockerfile` | libgit2 tarball `v1.3.0` → `v1.5.0` (both cross-build arches) |
| `backend/Dockerfile.local` | libgit2 tarball `v1.3.0` → `v1.5.0` |
| `devops/docker/lake-builder/Dockerfile` | libgit2 tarball `v1.3.2` → `v1.5.0`; **plus two build fixes**, see below |
| `.devcontainer/Dockerfile` | `git clone -b v1.3.0` → `-b v1.5.0` |
| `backend/plugins/gitextractor/README.md` | documented local build version |
| `backend/scripts/install-libgit2.sh` + Go CI workflows | checksum-pinned libgit2 1.5.0 bootstrap for the currently published builder image |

Out of scope on purpose: `devops/docker/alpine-dbt/Dockerfile` (`apk add
libgit2-dev`) feeds the dbt image, not the CGO `gitextractor` build.

## CI compatibility and builder-image follow-up

The Go CI workflows (`test.yml`, `golangci-lint.yml`, `test-e2e.yml`, and
`migration-script-lint.yml`) run inside the prebuilt container
`mericodev/lake-builder:latest`. That image currently ships **libgit2 1.3.2**
(and Go 1.20.4), so git2go v34 cannot compile in it:

```
/go/pkg/mod/github.com/libgit2/git2go/v34@v34.0.0/Build_system_dynamic.go:12:3:
  error: #error "Invalid libgit2 version; this git2go supports libgit2 between v1.5.0 and v1.5.0"
```

To keep the PR and post-merge `main` CI runnable before that image is
republished, this PR adds a checksum-pinned, idempotent
`backend/scripts/install-libgit2.sh` bootstrap to the lint, unit-test, and both
E2E jobs. It installs libgit2 1.5.0 in the old builder container before Go
compilation and becomes a no-op once the image already provides libgit2 1.5.x.
`migration-script-lint` does not compile git2go and was already green, so it
does not need the bootstrap.

An image rebuild is therefore **not a prerequisite for green PR checks**.
After merge, maintainers should still push a `builder-*` tag so
`build-builder.yml` republishes `mericodev/lake-builder:latest` with the updated
toolchain. That publishing workflow needs the repository's Docker Hub
credentials and can only be run by maintainers.

GitHub's separate `action_required` conclusion on newly pushed fork commits is
the normal fork-workflow approval gate; it is unrelated to the builder-image
compatibility issue described above.

Two extra commits are included so that the image *can* actually be rebuilt —
`devops/docker/lake-builder/Dockerfile` is currently **unbuildable on `main`**,
independently of this change:

1. `golang:1.26-bullseye` (introduced by #8875) was never published — Go dropped
   the bullseye variants. The stage does not use Go at all, so it now uses plain
   `debian:bullseye` with `wget`/`ca-certificates` installed explicitly.
2. The libgit2 cmake step failed with *"Could not find a python interpreter,
   which is needed to build the tests"* on that base → `-DBUILD_TESTS=OFF`
   (libgit2's own test suite was never built here anyway).

Suggested order: merge after the PR workflows have run and passed, then push a
`builder-*` tag so `mericodev/lake-builder:latest` is rebuilt with libgit2
1.5.0. Happy to split the two Dockerfile fixes into a separate, earlier PR if
that is preferred.

## Validation

**Fork preflight (all jobs included there green):** since the published builder
image could not compile git2go v34 directly, a fork workflow built
`devops/docker/lake-builder/Dockerfile` itself and ran the following jobs inside
it —
[run 30828741519](https://github.com/DoDiODev/devlake/actions/runs/30828741519),
**8/8 jobs successful**: builder image, lint (go), unit-test, e2e (mysql),
migration-script-lint, config-ui, ASF license header, grafana dashboards. This
earlier fork workflow did **not** include the upstream PostgreSQL E2E job.

**Published-builder bootstrap smoke test (green):**
[run 31404675667](https://github.com/DoDiODev/devlake/actions/runs/31404675667)
started in the unchanged `mericodev/lake-builder:latest` image (Go 1.20.4,
libgit2 1.3.2), installed libgit2 1.5.0 with the new script, successfully ran
`go build ./plugins/gitextractor/...`, and confirmed that a second bootstrap
invocation is a no-op.

**Full affected Go CI against the published builder (all green):**
[run 31405140967](https://github.com/DoDiODev/devlake/actions/runs/31405140967)
used the PR head plus only a temporary fork workflow, ran the bootstrap in each
job, and passed **4/4 jobs**: golangci-lint, unit-test, E2E (MySQL), and E2E
(PostgreSQL).

Additionally, locally on macOS (arm64) against a self-built libgit2 1.5.0 prefix:

- ✅ `go build ./plugins/gitextractor/...` — compiles and links against libgit2 1.5.0
- ✅ standalone git2go v34 smoke program (open repo, iterate ODB/branches/tags, diff lines)
- ✅ `make mock`, plugin build (`gitextractor.so` builds as a Go plugin)
- ✅ `make unit-test` — no failures
- ✅ `docker build -f backend/Dockerfile.local` — green
- ⚠️ `docker build -f backend/Dockerfile` (multi-arch cross build) fails locally
  **with and without this change** — the identical failure reproduces on
  unmodified `main` with libgit2 1.3.0, i.e. it is a pre-existing limitation of
  the local (Rancher Desktop/emulated) environment, not a regression of this PR.
  Left to CI to verify.
- ⚠️ `.devcontainer/Dockerfile` cannot be built locally either: its base image
  `mcr.microsoft.com/devcontainers/go:1-1.26-bookworm` does not exist (pre-existing,
  unrelated to this change).

## Note for contributors (post-merge)

Anyone building `gitextractor` locally must rebuild/reinstall **libgit2 1.5.0** —
an existing 1.3.x prefix is ABI-incompatible with git2go v34 and will fail at link
or, worse, at runtime. Devcontainer users need a "Rebuild Container".

## Question to the maintainers

`libgit2/git2go` has been **inactive since Oct 2022**; its highest importable
module is `v34` (libgit2 1.5), and the `v35.0.0` tag is broken. That means
`apache/devlake` is permanently cut off from libgit2 ≥ 1.6 security and bug fixes
as long as the canonical module is used.

There is an actively maintained fork, **`coveord/git2go`** (last commit
2025-11-25), which publishes correctly tagged majors: `v35` → libgit2 1.6,
`v36` → libgit2 1.7. It keeps the canonical module path, so using it requires a
Go `replace` directive (or vendoring) — it is not a drop-in.

**Would the maintainers be open to switching to `coveord/git2go` in a follow-up
PR?** This PR deliberately stays on the canonical module (`v34` / libgit2 1.5.0)
and does not introduce the fork.

## Rollback

Revert to the status quo `git2go v33.0.6` + libgit2 1.3.x — same Dockerfile
structure, only the version numbers differ.


klesh (approved):
LGTM.
Good work.

PR #9032, opened 2026-08-04
fix(core): replace constraints.Unsigned with local interface in archived base

mfrancisc:
## Summary
- Replace `golang.org/x/exp/constraints.Unsigned` with a local `unsignedInteger` interface in archived `GenericModel`
- Drops an unnecessary `x/exp` import that can break golangci-lint typecheck on older Go CI


klesh (approved):
LGTM.
Thanks for your contribution.

PR #9034, opened 2026-08-06
feat(github): warn when adding a repo already scoped on another connection

### LINKED ISSUE #8947: [Bug][Config-UI] No warning when the same repository is added under multiple connections, causing duplicate domain-layer records
### Search before asking

### What happened

When the same physical repository is added to DevLake under more than one connection - which the UI currently allows without any warning - every entity collected from that repository (pull requests, issues, board associations, repo-commit links) is stored as a separate record for each connection.
For example a repository connected via four connections, every pull request appears four times in the pull_requests table with four different primary keys. Any metric computed over these tables - PR count, cycle time, throughput, DORA lead time - is inflated by the number of connections pointing at the same repo.
**The UI gives no indication that this configuration will produce duplicate data**. Users following the suggested multi-connection workaround [issue#7684](https://github.com/apache/devlake/issues/7684#issuecomment-2198868619) are silently creating corrupted metrics.

**Verification**: All duplicate records share the same url field (e.g., https://github.com/owner/repo/pull/123). Running the following query confirms the problem:

```
SELECT url, COUNT(*) as copies
FROM pull_requests
GROUP BY url
HAVING COUNT(*) > 1
ORDER BY copies DESC;
```

When a user adds a repository scope that is already registered under a different connection (detected by matching _html_url_ / _clone_url_ across connections), the UI should display a clear warning before the user saves, for example:
"_This repository is already connected via Connection 'GitHub Production'. Collecting it here will create duplicate pull requests and issue records, which will inflate all metrics for this repository._"

The warning should not block the action - there are legitimate reasons to have the same repository under multiple connections (different scope configs, different team tokens). But the user should be able to make an informed choice.

Additionally, a backend diagnostics endpoint would help existing installations detect the problem:
```
GET /api/scope-duplicates
```

Returns a list of repository URLs that appear under more than one connection, along with the affected connection IDs, so administrators can audit and clean up existing configurations.

### How to reproduce

1. Add the same GitHub repository to DevLake under two different connections.
2. Run blueprints for both connections.
3. Query pull_requests grouped by url - every PR will appear twice.
4. Note that at no point during configuration does the UI warn about this.

### Proposed implementation

**Backend** - one new API handler that queries _tool_github_repos (and equivalent tables for other plugins) grouped by html_url, returning repos that appear under more than one connection:

`GET /api/plugins/github/scope-duplicates`

**Config-UI** - when a user selects a repository scope in the blueprint or connection wizard, call the endpoint and render a dismissible warning banner if the selected repo URL is already registered elsewhere.

**Additional context**
This issue affects all data-source plugins that support multiple connections to the same platform instance (GitHub, GitLab, Bitbucket, etc.).
A related workaround exists: deduplicating views over the domain tables using url as a natural key. We are willing to contribute that as a stopgap alongside the UI fix if it would be useful to the project ( see : https://github.com/konflux-ci/devlake/pull/106 )

### Version

main

### Code of Conduct

mfrancisc:
## Summary
Fixes #8947
When the same GitHub repository is added under more than one connection, DevLake stores duplicate domain-layer records and metrics get inflated. The Config UI previously allowed this with no warning.
This PR adds:
- `GET /plugins/github/scope-duplicates` — diagnostics (repos already on multiple connections) and a pre-add check (`connectionId` + `githubIds`)
- A dismissible, non-blocking warning in Config-UI when selecting a GitHub repo that already exists under another connection
Save is still allowed; users can choose to proceed intentionally.

## Screenshot
image

## Follow-ups
- Same API + Config-UI warning for GitLab and Bitbucket
- Optional global `GET /api/scope-duplicates` aggregator
- Domain-layer dedup / metric fixes remain out of scope

### DIFF
--- a/backend/plugins/github/api/scope_duplicates_api.go
+++ b/backend/plugins/github/api/scope_duplicates_api.go
@@ -1,0 +1,205 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/apache/incubator-devlake/core/dal"
+	"github.com/apache/incubator-devlake/core/errors"
+	"github.com/apache/incubator-devlake/core/plugin"
+)
+
+// ScopeDuplicateConnection is a connection that shares a repository scope.
+type ScopeDuplicateConnection struct {
+	ConnectionId   uint64 `json:"connectionId"`
+	ConnectionName string `json:"connectionName"`
+}
+
+// ScopeDuplicateGroup is one repository that appears under multiple connections
+// (diagnostics) or already exists under another connection (pre-add check).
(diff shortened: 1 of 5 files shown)


klesh (approved):
LGTM

PR #9042, opened 2026-08-10
fix(jira): stop the changelog convertor hiding unconverted items forever

### LINKED ISSUE #8834: Jira changelog items collected but not converted to domain layer (issue_changelogs)
### Search before asking

### What happened

Jira status/resolution changelog items are collected and stored in `_tool_jira_issue_changelog_items` but a significant percentage are never promoted to the domain-level `issue_changelogs` table. This happens silently — no errors, no warnings.                                                                                                                                                                    
                                                                                                                                                                                                                
The affected changelogs are from the same sync run, same board, same connection as the ones that ARE converted. The converter processed approximately 79% and silently skipped approximately 21% from the same sync run.                                                
                                                                                                                                                                                                                
Across multiple Jira projects on the same instance, the percentage of resolved issues with no status changelogs in the domain table ranges from 4% to 99%.                                                    
  
For a specific issue, the tool layer has the data:                                                                                                                                                                                                                        
                                                                                                                                                                                                     
```
  SELECT ci.field, ci.from_string, ci.to_string             
  FROM _tool_jira_issue_changelog_items ci                                                                                                                                                                      
  JOIN _tool_jira_issue_changelogs cl
    ON ci.connection_id = cl.connection_id AND ci.changelog_id = cl.changelog_id                                                                                                                                
  WHERE cl.issue_id = <ISSUE_ID> AND cl.connection_id = <CONN_ID> AND ci.field = 'status';  
```
                                                                                                                 
  -- Returns: e.g. New → In Progress, In Progress → Closed                                                                                                                                                      
                                                                                                                                                                                                                
But the domain layer is empty:                                                                                                                                                                                
                                                                                                                                                                                                                
```
  SELECT field_name, original_from_value, original_to_value                                                                                                                                                     
  FROM issue_changelogs                                                                                                                                                                                         
  WHERE issue_id = 'jira:JiraIssue:<CONN_ID>:<ISSUE_ID>' AND field_name = 'status'; 
```       
                                                                                                              
  -- Returns: (empty)                                                                                                                                                                                           
                                                                                                                                                                                                                
  Jira UI (History tab) confirms the status transitions exist.                                                                                                                                                  
                                                            
  What we've ruled out:                                                                                                                                                                                         
  - Not board scoping — all affected issues exist in _tool_jira_board_issues for the same board as the working issues.
  - Not incremental since filter — _devlake_collector_latest_state shows time_after far in the past, meaning all changelogs should be converted.                                                                
  - Not a collection gap — 64% of the affected issues have tool-level changelogs that were synced on the same date as the working ones.         
  - Not a specific field — the affected issues have changelog items for status, resolution, labels, Story Points, and other fields. None were converted.                                                        
                                                                                                                                                                                                                
  Related issues: #7911, #7826, #8196, #7394     

Expected behavior:
All changelog items in _tool_jira_issue_changelog_items that have a matching _tool_jira_board_issues entry should be converted to issue_changelogs. The converter should not silently skip records — if a changelog item cannot be converted, it should log a warning. 

Actual behavior:                                                                                                                                                                                              
The converter silently skips a subset of changelogs. No error is logged. The skipped items remain in the tool layer but never appear in the domain layer. 

### How to reproduce

  1. Configure a Jira Cloud connection with a board                                                                                                                                                             
  2. Run a full sync (not incremental)
  3. Compare counts:                                                                                                                                                                                            
                                                                                                                                                                                                                
  -- Tool layer count                                                                                                                                                                                           

```
  SELECT COUNT(*) FROM _tool_jira_issue_changelog_items ci                                                                                                                                                      
  JOIN _tool_jira_issue_changelogs cl                                                                                                                                                                           
    ON ci.connection_id = cl.connection_id AND ci.changelog_id = cl.changelog_id
  WHERE cl.connection_id = <CONN_ID> AND ci.field = 'status';   
```                                                                                                                                   
                                                            
  -- Domain layer count                                                                                                                                                                                         

```
  SELECT COUNT(*) FROM issue_changelogs                     
  WHERE issue_id LIKE 'jira:JiraIssue:<CONN_ID>:%' AND field_name = 'status';
```
                                                                                                                                                                                                                
  4. If domain count is significantly lower than tool count, the bug is present.                                                                                                                                
  5. Verify affected issues have status transitions in Jira UI (History tab).          

### Anything else

  Related issues                                                                                                                                                                                                
                                  
  - #7911 — field_id empty in changelog items (fixed, PR #7912)                                                                                                                                                 
  - #7826 — Issues disappearing from dataset (raw has data, domain doesn't — same symptom pattern)
  - #8196 — Request to use Jira Projects instead of Boards
  - #7394 — Incremental mode for changelog converter  

### Version

Latest as of March 2026. Managed instance — exact version/commit SHA not accessible.

### Code of Conduct

vbhanuchander-lang:
Closes #8834

### The mechanism is the incremental filter, not the LEFT JOIN

The thread converged on the `LEFT JOIN` + `WHERE board_id` anti-pattern, and @klesh reasonably rejected the proposed fix because moving `board_id` into the `ON` clause makes every board task convert the whole connection. I think the actual defect is somewhere else, and the fix costs nothing in scope or performance.

The convertor's incremental filter is:

```go
dal.Where("_tool_jira_issue_changelog_items.created_at >= ? ", since)
```

`created_at` comes from `common.NoPKModel` and is stamped when the row is **first inserted**. It never moves again. So a changelog item collected during one window but not converted in that window can never be selected by any later incremental run — the column it is filtered on is permanently in the past. Nothing errors. The rows just stay in the tool layer.

That fits every symptom in the report:

| Reported | Explained by |
|---|---|
| Partial, ~79% converted / ~21% not, same sync run | window boundary, not board membership |
| Persists across runs | `created_at` cannot move, so the item is permanently excluded |
| No errors, no warnings | nothing failed — the rows were never selected |
| 4%–99% across projects | depends on how much of each project's history predates the window |

It also explains why the "not the incremental filter" check came out clean: `_devlake_collector_latest_state` is the **collector's** state. The convertor keeps its own subtask state, and that is what gates this query.

### The fix

```diff
-dal.Where("_tool_jira_issue_changelog_items.created_at >= ? ", since)
+dal.Where("_tool_jira_issue_changelog_items.updated_at >= ? ", since)
```

`updated_at` is refreshed by the extractor's upsert (`CreateOrUpdate` → `OnConflict{UpdateAll: true}`), so a re-collected item is reconsidered and a re-collection actually repairs the gap.

Consistency check across the code base — of 36 convertors, this was **the only one** filtering on `created_at`:

```
$ grep -rh "dal.Where(\"" --include="*convertor*.go" plugins/ | grep -E "(created|updated)_at >= "
  35 updated_at
   1 created_at   <- plugins/jira/tasks/issue_changelog_convertor.go
```

**Crucially, this does not widen the board filter.** No board task converts anything outside its own board, so the cost @klesh objected to does not arise.

### Making the board filter non-silent

The board filter is deliberate and I have left it alone. But the issue also asks that the convertor not skip records silently, and that part is fair independently of the bug above.

After a successful conversion the subtask now counts collected changelog items whose issue is not on this board, and logs the number with the reason. One aggregate query per board task, negligible next to the conversion. It turns "21% of my changelogs are missing" into a logged figure with an explanation. A failure in the diagnostic is logged, never propagated — it cannot turn a good run into a failed one.

@ciaramulligan — for overlapping boards this means the shortfall becomes visible and attributable rather than mysterious, and with the filter fix a re-collection now repairs previously stranded items.

### Tests

`TestIssueChangelogBoardScopeDataFlow` (e2e, ran against MySQL 8.4) drives the convertor with a changelog for an issue on no board and asserts: the out-of-scope item is not converted, nothing is attached to `jira:JiraIssue:2:0`, and in-scope changelogs still convert — so the exclusion is scoping, not a regression that dropped everything. The diagnostic is observably firing in that run:

```
level=warning msg="1 collected changelog item(s) are not associated with board 8 and were not
converted; they belong to issues outside this board's scope. ..."
```

`TestIncrementalFilterUsesUpdatedAtNotCreatedAt` guards the column choice. Reverting it is a one-token change that silently restores permanent data loss and nothing else in the suite would catch it.

Full plugin, with a database attached:

```
ok  github.com/apache/incubator-devlake/plugins/jira/api               0.221s
ok  github.com/apache/incubator-devlake/plugins/jira/e2e              11.106s
ok  github.com/apache/incubator-devlake/plugins/jira/tasks            0.254s
ok  github.com/apache/incubator-devlake/plugins/jira/tasks/apiv2models 0.418s
```

### What I could not verify

I have no access to the reporter's Jira instance, so I have **not** reproduced the original 21% shortfall end to end. The mechanism above is established from the code, the model definitions and the upsert behaviour, and it is consistent with every reported symptom — but the confirmation that would close it beyond doubt is @ciaramulligan running a sync on this branch and comparing tool-layer and domain-layer counts again.

If the shortfall persists after this, the remaining suspect is genuine board non-membership, and the new warning will say exactly how many items that accounts for — which turns the next round of diagnosis into reading a log line instead of writing SQL.

### DIFF
--- a/backend/plugins/jira/e2e/issue_changelog_board_scope_test.go
+++ b/backend/plugins/jira/e2e/issue_changelog_board_scope_test.go
@@ -1,0 +1,105 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"testing"
+	"time"
+
+	"github.com/apache/incubator-devlake/core/dal"
+	"github.com/apache/incubator-devlake/core/models/domainlayer/ticket"
+	"github.com/apache/incubator-devlake/helpers/e2ehelper"
+	"github.com/apache/incubator-devlake/plugins/jira/impl"
+	"github.com/apache/incubator-devlake/plugins/jira/models"
+	"github.com/apache/incubator-devlake/plugins/jira/tasks"
+	"github.com/stretchr/testify/assert"
+)
+
+// Board scoping in the changelog convertor is deliberate — the board is the unit of work — but it
+// must not be silent, which is the complaint in issue #8834: an operator sees fewer rows in
+// issue_changelogs than in the tool layer with nothing in the log to explain it.
+//
+// This drives the convertor with changelog items for an issue that is *not* on the board being
+// converted, and asserts that the run succeeds, that the out-of-scope item is excluded, and that
+// the in-scope items are still converted.
+func TestIssueChangelogBoardScopeDataFlow(t *testing.T) {
+	var plugin impl.Jira
+	dataflowTester := e2ehelper.NewDataFlowTester(t, "jira", plugin)
+
+	taskData := &tasks.JiraTaskData{
+		Options: &tasks.JiraOptions{
+			ConnectionId: 2,
+			BoardId:      8,
+		},
+	}
+
+	dataflowTester.ImportCsvIntoTabler("./snapshot_tables/_tool_jira_issue_changelogs.csv", &models.JiraIssueChangelogs{})
+	dataflowTester.ImportCsvIntoTabler("./snapshot_tables/_tool_jira_issue_changelog_items.csv", &models.JiraIssueChangelogItems{})
+	dataflowTester.ImportCsvIntoTabler("./snapshot_tables/_tool_jira_statuses_for_changelog.csv", &models.JiraStatus{})
+	dataflowTester.ImportCsvIntoTabler("./snapshot_tables/_tool_jira_board_issues_for_changelog.csv", &models.JiraBoardIssue{})
+	dataflowTester.ImportCsvIntoTabler("./snapshot_tables/_tool_jira_issue_fields.csv", &models.JiraIssueField{})
+
+	db := dataflowTester.Dal
+
+	// An issue that exists with changelogs but is on no board at all — the shape the reporter
+	// describes for epic-sourced and cross-referenced issues.
+	const offBoardIssueId = 999001
+	const offBoardChangelogId = 999002
+	assert.NoError(t, db.CreateOrUpdate(&models.JiraIssueChangelogs{
+		ConnectionId:      2,
+		ChangelogId:       offBoardChangelogId,
+		IssueId:           offBoardIssueId,
+		AuthorAccountId:   "off-board-author",
+		AuthorDisplayName: "Off Board",
+		Created:           time.Date(2026, 3, 1, 12, 0, 0, 0, time.UTC),
+	}))
+	assert.NoError(t, db.CreateOrUpdate(&models.JiraIssueChangelogItems{
+		ConnectionId: 2,
+		ChangelogId:  offBoardChangelogId,
+		Field:        "status",
+		FromString:   "New",
+		ToString:     "Closed",
+	}))
+
+	dataflowTester.FlushTabler(&ticket.IssueChangelogs{})
+	dataflowTester.Subtask(tasks.ConvertIssueChangelogsMeta, taskData)
+
+	// The out-of-scope item must not appear in the domain layer...
+	offBoard, err := db.Count(
+		dal.From(&ticket.IssueChangelogs{}),
+		dal.Where("issue_id = ?", "jira:JiraIssue:2:999001"),
+	)
+	assert.NoError(t, err)
+	assert.Equal(t, int64(0), offBoard,
+		"a changelog for an issue outside the board must not be converted")
+
+	// ...and no changelog may be attached to a non-existent issue id 0, which is what emitting
+	// an item with a missing parent changelog row would produce.
+	orphans, err := db.Count(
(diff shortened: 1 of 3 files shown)


klesh (approved):
LGTM
Thanks for your contribution.

ciaramulligan:
Thanks @vbhanuchander-lang  — We've been working around the bug by querying the tool tables directly instead of issue_changelogs, and the pattern you identified matches what we were seeing. Smaller boards were barely affected but some of our larger multi-project boards had significant data loss.

I checked the missing items and they consistently have created_at stuck at first-insert time while updated_at was refreshed by re-extraction — like you describe.

I'll validate the fix end-to-end once we upgrade our Devlake version and report back with the results. Thank you!

vbhanuchander-lang:
Thanks for confirming it against real data — `created_at` stuck at first-insert while
`updated_at` moves with re-extraction is exactly the signature, and it's useful to know it
scaled with board size for you.

One thing worth planning for before you validate: **upgrading alone will not backfill the
rows you have already lost.** The new filter is still only applied on incremental runs:

```go
if stateManager.IsIncremental() {
    since := stateManager.GetSince()
    if since != nil {
        clauses = append(clauses, dal.Where("..._changelog_items.updated_at >= ? ", since))
    }
}
```

Your historical gaps have an `updated_at` older than the `since` of the next incremental
run, so they stay outside the window. Two ways to recover them, either is fine:

- Run a **full-refresh sync** on the affected boards. `IsIncremental()` is false, no time
  filter is added, and every collected changelog item in board scope is converted.
- Or re-collect the issues first — the extractor upserts with `OnConflict{UpdateAll: true}`,
  which refreshes `updated_at`, after which a normal incremental run reconsiders them.

After that, if anything still looks missing, the PR also adds a diagnostic that should tell
you whether it is scope rather than a bug. It logs, once per board:

> `N collected changelog item(s) are not associated with board <id> and were not converted;
> they belong to issues outside this board's scope.`

Previously those items were silently dropped, so "missing" and "out of scope" were
indistinguishable in the logs. If that warning accounts for your remaining delta, the data
is being excluded by board scope by design and you would need a board covering those issues.

Please do report back — if the numbers don't line up after a full refresh I would like to
know, since that would point at something beyond this fix.

PR #9043, opened 2026-08-11
refactor(api): rename GetURIStringPointer to ResolveRequestURI, reject absolute relativePath

klesh:
## Summary
- Renamed `GetURIStringPointer` to `ResolveRequestURI` and added a doc comment explaining its resolution semantics.
- Added validation that `relativePath` must be a relative reference: `url.URL.ResolveReference` silently discards the base URL when the reference is absolute (`scheme://...`) or protocol-relative (`//host/...`), which would send the request to the wrong host with no error. Now this returns a `BadInput` error instead.
- Updated both call sites (`ApiClient.Do` and the GitHub plugin's `refreshApiClient.do`).

## Test plan
- [x] `go build ./helpers/pluginhelper/api/... ./plugins/github/token/...` passes
- [x] No other callers or tests reference the old function name (verified via repo-wide search)


Startrekzky (approved):
LGTM

PR #9045, opened 2026-08-12
fix(framework): first-ever converter run should full-sync, not silent…

### LINKED ISSUE #9016: [Bug][framework] First ever converter run bootstraps into incremental mode and silently skips the entire collected backlog
### Search before asking

### What happened

Fresh install, connected github with appriopriate scopes to pull CI history, ran the blueprint & CI result sets were empty.

Initial run of a converter that has never ran before should perform a full sync of collected data, regardless of the collectors state.

### How to reproduce

1. Fresh install of `v1.0.3-beta15` (empty database).
2. Add a GitHub connection + scope with CICD entities enabled, on a repo with GitHub Actions history
3. Run the blueprint once, let it finish
4. Observe: `_tool_github_runs` is populated, task log shows `Convert Workflow Runs ... records processed: 0`, and `select count(*) from cicd_pipelines` returns 0. Same for  `_tool_github_releases` vs `cicd_releases`.

### Anything else

Only on first install

### Version

v1.0.3-beta15

### Code of Conduct

bramhanandlingala:
### Summary
Fixes a bug where a subtask running for the very first time (fresh install, or
a newly-added converter subtask) is silently bootstrapped into incremental
mode, causing it to filter out the entire already-collected backlog and
produce zero rows in the domain layer — even though the tool-layer table is
fully populated.

Repro: fresh DB → GitHub connection + CICD scope → run blueprint once →
`_tool_github_runs` is populated, but `cicd_pipelines` returns 0 rows.

### Root Cause
`bootstrapStateFromCollectorStateIfNeeded` (in
`backend/helpers/pluginhelper/api/subtask_state_manager.go`) copied the
sibling collector's `LatestSuccessStart` into the subtask's own
`PrevStartedAt` whenever the subtask had no prior state of its own.

`PrevStartedAt == nil` is the framework's signal that a subtask has never run
before, and it's what triggers a full sync in
`calculateStateManagerIncrementalMode`. On a fresh install, the collector
finishes moments before the converter runs — in the same pipeline execution —
so `LatestSuccessStart` is effectively "now". Bootstrapping `PrevStartedAt`
from that value made the converter's genuinely-first-ever run look like a
*returning* run, flipping it into incremental mode and filtering rows by
`since` against historical timestamps (e.g. `github_updated_at`), which
silently dropped the entire backlog.

### Fix
Stop bootstrapping `PrevStartedAt` from the collector state. Only `TimeAfter`
is still inherited from the collector, since it's just a fallback lower bound
and has no effect on the incremental-vs-full-sync decision.

This is a shared framework fix (`SubtaskStateManager`, used by
`StatefulDataConverter`/`StatefulDataExtractor`), so it applies to every
plugin using stateful subtasks — not just GitHub's CICD converters.

Closes #9016

### DIFF
--- a/backend/helpers/pluginhelper/api/subtask_state_manager.go
+++ b/backend/helpers/pluginhelper/api/subtask_state_manager.go
@@ -160,9 +160,6 @@ func bootstrapStateFromCollectorStateIfNeeded(db dal.Dal, preState *models.Subta
 		return nil, errors.Default.Wrap(err, "failed to load collector state for subtask bootstrap")
 	}
 
-	if collectorState.LatestSuccessStart != nil {
-		preState.PrevStartedAt = collectorState.LatestSuccessStart
-	}
 	if preState.TimeAfter == nil && collectorState.TimeAfter != nil {
 		preState.TimeAfter = collectorState.TimeAfter
 	}
--- a/backend/helpers/pluginhelper/api/subtask_state_manager_test.go
+++ b/backend/helpers/pluginhelper/api/subtask_state_manager_test.go
@@ -194,7 +194,7 @@ func TestBootstrapStateFromCollectorStateIfNeeded(t *testing.T) {
 	latest := errors.Must1(time.Parse(time.RFC3339, "2026-04-27T18:05:31Z"))
 	timeAfter := errors.Must1(time.Parse(time.RFC3339, "2024-12-20T00:00:00Z"))
 
-	t.Run("bootstrap prev_started_at and time_after from collector state", func(t *testing.T) {
+	t.Run("bootstrap time_after but not prev_started_at from collector state", func(t *testing.T) {
 		mockDal := new(mockdal.Dal)
 		mockDal.On("HasTable", mock.Anything).Return(true).Once()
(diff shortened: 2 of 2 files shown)


klesh:
The diagnosis looks legitimate to me, but the fix does not; it effectively disables the incremental state completely. I think a better way is to migrate to `subtask_state_manager` and use the tool layer table (`_tool_github_runs`) for the `Table` field so that we can track the incremental state correctly.

bramhanandlingala:
Hi @klesh, thanks so much for reviewing this!

Just to clarify the scope of the change — the diff only removes the two lines that copied collectorState.LatestSuccessStart into PrevStartedAt. Everything else stays the same, so incremental mode still works normally after the first run: Close() saves PrevStartedAt, and the next run picks it up and goes incremental as before. It's only the very first run that behaves differently now — it does a proper full sync instead of being mistaken for a run that already happened.

On the _tool_github_runs idea — I think we might run into the same issue there. The extractor writes to that table just before the converter runs, in the same pipeline, so on a genuine first run it would already have data too, and we could end up dropping the backlog again the same way.

Let me know your thoughts, and happy to adjust if I'm missing something on your end!

PR #9046, opened 2026-08-12
fix(gitlab): add missing commit_updated_at column used by the MR incremental filter

DoDiODev:
# fix(gitlab): add missing `commit_updated_at` column used by the MR incremental filter

## Summary

#8959 changed `GetMergeRequestsIterator` (`backend/plugins/gitlab/tasks/shared.go`) to
filter merge requests by

```sql
GREATEST(gmr.gitlab_updated_at, COALESCE(gmr.commit_updated_at, gmr.gitlab_updated_at)) > ?
```

but `commit_updated_at` does not exist on `_tool_gitlab_merge_requests`. There is no field
on the `GitlabMergeRequest` model, no migration script, and no code writing the value —
the only occurrences of the identifier in the whole repository are the three lines added
by that PR.

Consequently every **incremental** run of `Collect MR Notes` — the only caller passing a
stateful collector, `mr_commit_collector.go` passes `nil` — fails with:

```
subtask Collect MR Notes ended unexpectedly
  Error 1054 (42S22): Unknown column 'gmr.commit_updated_at' in 'where clause'
```

## Approach

Rather than reverting #8959, this PR makes the intended behaviour actually work:

* add `CommitUpdatedAt *time.Time` to `GitlabMergeRequest`
* add migration script `20260812_add_mr_commit_updated_at.go` (registered in `register.go`)
* maintain the column in `ExtractApiMergeRequestsCommits`: after extraction it is set to
  the latest `commit_authored_date` of the MR's commits, scoped to the current
  connection/project
* the filter in `GetMergeRequestsIterator` is left as introduced by #8959 and now resolves
  against a real column, so MRs that received new commits without their own `updated_at`
  being bumped (e.g. force-pushes) are picked up again

## Testing

* `make migration-script-lint`
* `go test ./plugins/gitlab/e2e/ -run TestGitlabMrCommitDataFlow` — green on **MySQL 8**
  and **PostgreSQL 14**; the extractor's `UPDATE ... SET ... = (SELECT MAX(...))` and the
  `GREATEST`/`COALESCE` filter behave identically on both engines
* `go test ./plugins/ -run TestTableInfo`

## Notes

The `GREATEST(...)` expression returns `NULL` on MySQL when `gitlab_updated_at` is `NULL`,
so such MRs stay excluded — same as before #8959, where `gitlab_updated_at > ?` also
evaluated to `NULL`. No behaviour change there.

### DIFF
--- a/backend/plugins/gitlab/models/migrationscripts/20260812_add_mr_commit_updated_at.go
+++ b/backend/plugins/gitlab/models/migrationscripts/20260812_add_mr_commit_updated_at.go
@@ -1,0 +1,50 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package migrationscripts
+
+import (
+	"time"
+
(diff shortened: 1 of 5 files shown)


klesh (approved):
Good catch!
Thanks for the fix.

PR #9047, opened 2026-08-12
feat(gitlab,bitbucket): warn when adding a repo already scoped on another connection

mfrancisc:
## Summary
Follow-up to #9034, which added the duplicate-scope warning for GitHub. This PR
extends the same feature to **GitLab** and **Bitbucket**, per the follow-up noted
in that PR's description (originally tracked in #8947).
When the same repository/project is added under more than one connection, DevLake
stores duplicate domain-layer records and metrics get inflated. The Config UI
previously allowed this with no warning for any plugin; #9034 fixed it for GitHub
only. This PR brings GitLab and Bitbucket to parity.
This PR adds:
- `GET /plugins/gitlab/scope-duplicates` and `GET /plugins/bitbucket/scope-duplicates`
   diagnostics (repos already on multiple connections) and a pre-add check
  (`connectionId` + `gitlabIds`/`bitbucketIds`)
- Generalization of the Config-UI `data-scope-remote` component and the scope API
  client so the same dismissible, non-blocking warning now renders for GitHub,
  GitLab, and Bitbucket alike, instead of being hardcoded to GitHub
  
Save is still allowed; users can choose to proceed intentionally, same as before.

### Bitbucket-specific note
Unlike GitHub/GitLab (numeric scope IDs), Bitbucket identifies repositories with a
string `owner/repo` id. The new endpoint and query parsing handle this as a string
rather than assuming a numeric ID, including comma-separated `bitbucketIds` query
values (e.g. `bitbucketIds=owner/repo-a,owner/repo-b`).
## Testing
- Added unit tests for both new endpoints' query parsing and row-grouping logic
  (`scope_duplicates_api_test.go` for each plugin), following the same test
  structure as the existing GitHub tests.
- Verified locally end-to-end via `docker-compose-dev-mysql.yml`: confirmed
  `GET /plugins/{github,gitlab,bitbucket}/scope-duplicates` all return `200` with
  the expected `{"duplicates": [...]}` shape, and that the Config UI bundle
  includes the generalized `gitlabIds`/`bitbucketIds` query-param logic.
- Manually tried to add the same repo to multiple connections, see :
Screenshot 2026-08-12 at 15 44 48

  
## Follow-ups 
- global `GET /api/scope-duplicates` aggregator across all plugins

### DIFF
--- a/backend/plugins/bitbucket/api/scope_duplicates_api.go
+++ b/backend/plugins/bitbucket/api/scope_duplicates_api.go
@@ -1,0 +1,200 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"net/http"
+	"strconv"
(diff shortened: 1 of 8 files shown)


klesh (approved):
LGTM
Thanks for your contribution.

PR #9048, opened 2026-08-12
feat(jira): add JIRA_SKIP_UNPARSEABLE_ISSUES to continue collectIssues on parse failure (#8949)

### LINKED ISSUE #8949: [Feature][jira] Add JIRA_SKIP_UNPARSEABLE_ISSUES to continue collection when an issue page can't be parsed

kotwal-itpro:
Closes #8949.

## Problem

On large Jira instances, `jira:collectIssues` occasionally receives an otherwise-successful (2xx) response whose body cannot be decoded as JSON — typically a truncated body from a dropped connection, or (in rare misconfigured-proxy cases) an HTML body served with a 200. The current `ResponseParser` returns an error, which aborts the entire subtask and kills the whole Jira collection even though thousands of other pages succeeded.

This is distinct from HTTP error statuses (`502/503/504`), which are already handled by the collector's retry logic — the request is specifically about parse failures on successful responses.

## Fix

Introduce a `JIRA_SKIP_UNPARSEABLE_ISSUES` env var:

- **default** (unset or anything other than `"true"`): existing behaviour — surface the parse error and fail the subtask.
- **`"true"`** (case-insensitive): log a warning and skip the unparseable page so the rest of the collection completes. The warning includes body length and a 256-byte body prefix so operators can still investigate without hiding the problem.

The V2 (Server) and V3 (Cloud) collectors previously duplicated the response-parsing closure. Factored the shared logic into a `parseIssuesResponse(logger)` helper so the new behaviour applies to both.

## Test

Adds `Test_parseIssuesResponse` in `issue_collector_test.go` covering:

- Happy path returns the expected issues, no warning.
- Default (env var unset) surfaces the parse error, no warning.
- `JIRA_SKIP_UNPARSEABLE_ISSUES=true` returns an empty non-nil slice and logs exactly one warning.
- Case-insensitivity (`TRUE`).
- Explicit `"false"` behaves like default.

Uses `t.Setenv` for isolation. Full `plugins/jira/tasks` suite still passes.

### DIFF
--- a/backend/plugins/jira/tasks/issue_collector.go
+++ b/backend/plugins/jira/tasks/issue_collector.go
@@ -23,15 +23,26 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
 	"github.com/apache/incubator-devlake/core/dal"
 	"github.com/apache/incubator-devlake/core/errors"
+	"github.com/apache/incubator-devlake/core/log"
 	"github.com/apache/incubator-devlake/core/plugin"
 	"github.com/apache/incubator-devlake/helpers/pluginhelper/api"
 	"github.com/apache/incubator-devlake/plugins/jira/models"
 )
+
+// skipUnparseableIssuesEnvVar controls whether jira:collectIssues aborts on a
+// JSON parse failure of an otherwise-successful (2xx) response body (default),
+// or logs a warning and skips the bad page so the rest of the sync completes.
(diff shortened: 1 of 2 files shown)


kotwal-itpro:
Superseded by #9026 (which uses `taskCtx.GetConfigReader().GetBool(...)` instead of `os.Getenv`, a cleaner integration with DevLake's config plumbing) — the same feature has now landed on `main`. Closing this. Thanks for the reviewer bandwidth.

PR #9049, opened 2026-08-13
Fix Copilot seat organization attribution causing wrong DORA adoption denominator

### LINKED ISSUE #9041: [Bug][gh-copilot] DORA dashboard uses seat-assigning organization as organization adoption denominator
### Search before asking

- [x] I searched the issues and found no issue covering this organization attribution and denominator problem.

Related, but not duplicates:

- #8964 updates Copilot API endpoints and schemas.
- #8960 adds AI adoption phases/cohorts.
- #9002 implements parts of those changes.

### What happened

The bundled `github-copilot-dora-correlation` dashboards calculate organization adoption as:

```sql
daily_active_users /
COUNT(_tool_copilot_seats WHERE organization = selected_scope.organization)
```

This assumes that Copilot seats assigned to members of an organization are also assigned by that organization.

GitHub explicitly supports assigning licenses through a central or "shell" organization. Its current attribution rules state that once a user has a Copilot seat anywhere in the enterprise, their usage is attributed to every organization in that enterprise where they are a member.

Therefore, an organization can have valid organization-scoped usage metrics while having zero seats assigned directly by that organization.

Observed result:

- `_tool_copilot_enterprise_daily_metrics` contained daily active-user data for the selected organization scope.
- `_tool_copilot_seats` contained no seats attributed to that organization.
- The dashboard denominator was zero, so adoption was `NULL`.
- Adoption trend, current adoption, correlation, and tier-based panels showed no or misleading data.
- `NULL` adoption was classified as `>75%` by tier queries because their `CASE` expressions use `ELSE '>75%'`.

The MySQL DORA dashboard currently contains this seat-denominator pattern in 27 query locations. The PostgreSQL version appears to use the same assumption.

There is also a plugin-level enterprise seat attribution concern:

- `CollectCopilotSeatAssignments` correctly calls the enterprise seats endpoint when an enterprise is configured.
- `copilotSeatResponse` does not parse the response's `organization` field.
- `ExtractSeats` sets every row's `Organization` to `connection.Organization`.
- The enterprise endpoint can return multiple seat rows for the same user when access is granted through multiple organizations or enterprise teams.

This means enterprise seat records cannot reliably represent organization membership or organization-level license eligibility.

GitHub documentation:

https://docs.github.com/en/copilot/concepts/copilot-usage-metrics/copilot-metrics#how-are-metrics-attributed-across-organizations

Relevant behavior from the documentation:

> If licenses are assigned in a dedicated "shell" organization for administrative purposes within the enterprise, users still appear in the metrics for all other organizations in the enterprise they belong to.

Organization-scoped dashboards should use an organization-scoped licensed-user population rather than filtering enterprise seats by their assigning organization.

### How to reproduce

1. Create an enterprise containing:
   - a central organization used to assign Copilot seats
   - a separate product organization
2. Assign Copilot seats to users through the central organization.
3. Make those users active members of the product organization.
4. Enable GitHub Copilot usage metrics for the enterprise.
5. Configure the DevLake `gh-copilot` plugin with the enterprise and product organization.
6. Collect Copilot reports and seat assignments.
7. Select the product organization in the bundled GitHub Copilot + DORA Correlation dashboard.
8. Confirm that organization-scoped active-user metrics exist.
9. Observe that the dashboard denominator is zero or otherwise incorrect because it filters seats by assigning organization.
10. Observe that `NULL` adoption weeks can be labeled `>75%`.

Diagnostic query:

```sql
SELECT
  day,
  daily_active_users
FROM _tool_copilot_enterprise_daily_metrics
WHERE connection_id = ?
  AND scope_id = ?
ORDER BY day DESC;
```

Current denominator pattern:

```sql
SELECT COUNT(*)
FROM _tool_copilot_seats s
WHERE s.connection_id = ?
  AND s.organization = ?;
```

The first query returns organization usage while the second returns no organization-assigned seats.

### Anything else

Counting rows from `_tool_copilot_user_daily_metrics` is not a valid replacement denominator. Those reports represent daily user activity and produce weekday/weekend fluctuations rather than the complete licensed population.

The current-state seat table also cannot reconstruct historical weekly eligibility. A historical adoption trend needs either:

- a daily licensed-population snapshot, or
- an effective-dated license/membership source.

The dashboard should avoid silently applying today's seat state to historical dates.

### Version

Observed with the `gh-copilot` plugin and bundled dashboard from the DevLake `v1.0.3-beta10` generation. The same denominator pattern is present on the current `main` branch.

- [ ] Yes, I am willing to submit a PR.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct).

bramhanandlingala:
## Summary
Fixes #9041. Copilot seats from GitHub's enterprise API were being
tagged with the connection's "shell" org instead of the org that
actually granted the seat, which made the DORA dashboard's adoption
percentage wrong (often zero) for any org besides the shell org. Two
dashboard queries also mislabeled NULL adoption as the highest tier
instead of excluding it.

## Root cause
- `seat_extractor.go` hardcoded every seat's organization to
  `connection.Organization`, ignoring the per-seat `organization`
  field the enterprise seats endpoint actually returns.
- Both DORA dashboards' adoption-tier `CASE` statements used `ELSE` as
  a catch-all, so NULL adoption silently got labeled `>75%` / `high`
  instead of being excluded.

## Fix
- Parse the API's per-seat `organization` field
  (`metrics_extractor.go`) and prefer it over the connection's
  configured org, falling back only when the API doesn't supply one
  (`seat_extractor.go`).
- Replace `ELSE` with explicit `WHEN` + `WHERE ... IS NOT NULL` in
  both dashboards' tier queries (MySQL and PostgreSQL).
- Refactored seat parsing into a standalone function and added unit
  tests covering org-scoped, enterprise-with-org, and
  enterprise-with-empty-org cases.

Closes #9041


anthonyhull-ford:
Thank you for creating a PR so quickly and for helping me investigate this issue.

Does this fix seats assigned through a central licensing organization?

The change stores `seat.organization.login`, which appears to identify the organization assigning the seat. If **Central Licensing Org** assigns a seat to a user who belongs to **Product Org A** and **Product Org B**, GitHub attributes usage to both product organizations, but the seat may remain attributed only to Central Licensing Org.

Would Product Org A and Product Org B still have zero denominators? Should licensed users be joined with organization membership, and could this scenario be added as a test?

bramhanandlingala:
@anthonyhull-ford Thanks for looking into this closely — good catch.

You're right that this doesn't cover centrally-licensed seats. I checked GitHub's docs, and when a seat is licensed through multiple orgs in the same enterprise, GitHub bills (and reports) only one org per seat, chosen by GitHub — so the seats API itself never exposes the user's other org memberships. That means Product Org A and B could still show zero denominators in that case, but it's not fixable from this endpoint alone; it'd need a separate join against org membership data.

I'd like to keep that as a follow-up rather than block this fix, since it's a different data source and a bigger scope.

Also — noticed the CI checks aren't triggering on this PR. Could someone approve the workflow run?

Issue #8698, opened 2026-02-05
[Bug][GitHub] Activities not present in worklogs


furlangiacomo:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

After adding a new user via the /api/rest/plugins/org/users.csv API, we are seeing only `Finish a commit` activities in the work logs.

<img width="639" height="604" alt="Image" src="https://github.com/user-attachments/assets/e1d5d128-6726-423e-8efe-15c292d39385" />

### What do you expect to happen

We should also be able to see activities like `Review PR` or `Open a PR`, as we see from the GitHub profile

<img width="740" height="270" alt="Image" src="https://github.com/user-attachments/assets/72b7fa50-53b3-49ad-80c2-e774e5a7ca75" />

### How to reproduce

I haven’t found a way to reproduce the issue yet, and I’m not sure if you have any insights.

### Anything else

For users who were already present, it's working fine.

### Version

v1.0.3-beta9

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

furlangiacomo:
@dosu the user is present in the `user_accounts` table with entries in `_raw_data_table` set to `gitextractor`. I’ve noticed that for users who have both `gitextractor` and `_raw_github_graphql_accounts`, the work logs work as expected.

furlangiacomo:
@klesh, sorry for tagging you directly. Do you have any clue, or am I just doing something wrong?

furlangiacomo:
Can anyone help with this? Thanks!

klesh:
> [@klesh](https://github.com/klesh), sorry for tagging you directly. Do you have any clue, or am I just doing something wrong?

Sorry for the late response, and unfortunately I have no clue either.

vbhanuchander-lang:
I traced this through the code. It is not something you are doing wrong — the CSV upload simply
cannot link a GitHub account in your situation, and your own observation about `gitextractor` vs
`_raw_github_graphql_accounts` is exactly the tell.

**Why only commit activities appear**

There is exactly one subtask that links users to accounts, `ConnectUserAccountsExact` in
[`backend/plugins/org/tasks/user_account.go`](https://github.com/apache/devlake/blob/main/backend/plugins/org/tasks/user_account.go),
[...]

PR #9051, opened 2026-08-13
fix(org): match user and account emails case-insensitively

### LINKED ISSUE #9050: [Bug][org] User/account linking compares emails case-sensitively and silently drops matches
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

`ConnectUserAccountsExact` links users to accounts by exact map lookup, so the email comparison is
case-sensitive. A user imported from `users.csv` with the address `Tyrone.Cummings@corp.com` is not
linked to an account whose email is `tyrone.cummings@corp.com`, and the account is silently left
unlinked.

In [`backend/plugins/org/tasks/user_account.go`](https://github.com/apache/devlake/blob/main/backend/plugins/org/tasks/user_account.go)
the map is keyed on the raw address (line 49):

```go
emails[user.Email] = user.Id
```

and looked up with the raw account address (line 79):

```go
if userId, ok := emails[account.Email]; account.Email != "" && ok {
```

Email addresses are treated case-insensitively in practice, and the two sides of this comparison
come from different systems: a corporate git config, a provider profile, and a hand-maintained CSV
routinely disagree on capitalisation of the same address. When they do, the link is dropped with no
warning — the subtask reports success and the account simply never appears in `user_accounts`.

The consequence is not cosmetic. Once an account is unlinked, everything that attributes through it
stops being counted for that user, so the account's activity is missing from work logs and from any
team-level metric built on `user_accounts`.

There is also no second chance: the scan is restricted to
`id NOT IN (SELECT account_id FROM user_accounts)` (line 58), so a later run will not revisit an
account that failed to match, and the name-based fallbacks (lines 87 and 95) only help when the
display name or login happens to equal the user's name exactly.

Two addresses that differ only in capitalisation should be treated as the same address and linked.

### How to reproduce

1. Import a user via `PUT /plugins/org/users.csv` whose email is `E6` (any mixed case value).
2. Have an account whose email is `e6`, with a name and login that do not match the user's name.
3. Run the `connectUserAccountsExact` subtask.
4. No `user_accounts` row is created for that account.

I added exactly this case to the existing `org` e2e fixtures; it fails before the change and passes
after.

### Anything else

Found while investigating #8698, where a reporter's GitHub account was never linked and only commit
activities appeared in work logs. Case sensitivity is **not** the cause of that issue — I posted the
root cause there separately — but it is the same class of silent linkage failure and it turned up
while reading this code.

### Version

main

### Code of Conduct

vbhanuchander-lang:
### Summary

- Closes #9050
- `ConnectUserAccountsExact` compared emails by exact map lookup, so addresses differing only in
  capitalisation were never linked and the account was silently left out of `user_accounts`.

`user_account.go` keyed the map on the raw address and looked it up with the raw account address:

```go
emails[user.Email] = user.Id                                              // line 49
...
if userId, ok := emails[account.Email]; account.Email != "" && ok {       // line 79
```

The two sides come from different systems — a git config, a provider profile, and a
hand-maintained CSV — which routinely disagree on the case of the same address. When they do, the
subtask still reports success, so the failure is invisible: the account just never appears in
`user_accounts`, and everything attributing through it stops counting towards that user.

There is no second chance either. The scan is restricted to
`id NOT IN (SELECT account_id FROM user_accounts)` (line 58), so a later run will not revisit an
account that failed to match.

### What this changes

Case is folded on both sides of the **email** comparison only.

**Name matching is deliberately untouched.** Emails are case-insensitive by convention; personal
names and logins are not, and loosening those comparisons could manufacture links that are simply
wrong. That felt like a separate decision rather than something to slip into this fix.

### Test Plan

The `org` e2e fixtures gain one account, `a11`, whose email (`E6`) differs from user `U006`'s (`e6`)
only in case, with `full_name` and `user_name` set to `zz` so neither name fallback can match. That
isolates the email path.

Verified against a MySQL 8 instance with `E2E_DB_URL` set:

| | Result |
| --- | --- |
| Without the change | `FAIL` — `records in user_accounts count not match expectation count,[expected(CSV):11][actual(DB):10]` |
| With the change | `ok github.com/apache/incubator-devlake/plugins/org/e2e` |

`go test ./plugins/org/... -count=1` → e2e and impl both pass. `go vet ./plugins/org/...` is clean,
and `go build ./plugins/org/tasks/` is clean.

Two pre-existing build complaints are unrelated to this change and reproduce identically on an
unmodified tree, which I checked rather than assumed: `plugins/org` reports
`function main is undeclared in the main package` because it builds as a Go plugin, and
`git2go/v33` needs `pkg-config`, which is absent from my machine.

### Context

Found while investigating #8698, where a reporter's GitHub account was never linked so only commit
activities showed in work logs. To be clear, case sensitivity is **not** the cause of that issue — I
posted the root cause there separately (GitHub returns an empty email for users who have not made it
public, and `users.csv` has no column for a provider login, so no branch can match). This is a
different instance of the same silent-linkage class that surfaced while reading the code.

🤖 AI-assisted changes — generated with Claude Code (Opus 5) and reviewed by me before submitting.


klesh (approved):
LGTM.
Thanks for your contribution.

PR #9053, opened 2026-08-13
[Jira] Fix false-positive "field name is duplicated" warnings when multiple boards share a connection

### LINKED ISSUE #9052: [Bug][Jira] getIssueFieldMap() emits false-positive "field name is duplicated" warnings when multiple boards share a connection
### Search before asking

### What happened

When a Jira connection is configured with multiple boards (scopes), `getIssueFieldMap()` in `backend/plugins/jira/tasks/epic_extractor.go` logs hundreds of false-positive warnings per pipeline run:

[convertIssueChangelogs] filed name Time Spent is duplicated
[convertIssueChangelogs] filed name Original Estimate is duplicated
[convertIssueChangelogs] filed name Remaining Estimate is duplicated

**Root cause:** The table `_tool_jira_issue_fields` stores fields per board. With 10 boards, the same field (identical field ID) appears 10 times. `getIssueFieldMap()` builds a name-based map and logs "duplicated" on the second occurrence — without checking whether the field ID is actually different.

**SQL evidence from our instance (1 connection, 10 boards):**

| Connection | True name duplicates (different field IDs) | Board duplicates (same ID, multiple boards) | Total |
|---|---|---|---|
| 3 | 1 | 852 | 853 |

99.9% of warnings are false positives caused by board-level storage, not actual field name collisions.

- DevLake version: v1.0.3-beta15 (also confirmed present in v1.0.3-beta10)
- 1 Jira connection with 10 boards

The warning should only be emitted when two fields with DIFFERENT field IDs share the same display name — i.e., a genuine name collision. Same field ID appearing multiple times (due to multi-board storage) should not trigger a warning.

Proposed fix in `getIssueFieldMap()`:

```go
// Current:
if _, exists := fieldMap[field.Name]; exists {
    logger.Warn("filed name %s is duplicated", field.Name)
}

// Proposed:
if existing, exists := fieldMap[field.Name]; exists {
    if existing.FieldId != field.FieldId {
        logger.Warn("field name %s is duplicated (IDs: %s, %s)", field.Name, existing.FieldId, field.FieldId)
    }
}
This would eliminate ~99.9% of false warnings while preserving the alert for genuine ambiguity.

### How to reproduce

1. Create a Jira connection in DevLake
2. Add multiple boards (≥2) as scopes to that connection
3. Run a data collection (blueprint or manual trigger)
4. Observe logs during convertIssueChangelogs task
5. Every Jira field that exists on more than one board will produce a "filed name ... is duplicated" warning — multiplied by (number_of_boards - 1).

### Anything else

- The log message contains a typo: "filed name" should be "field name" — could be fixed in the same PR.
- The only genuine name collision in our instance is "Progress" (built-in progress field vs. customfield_36295 both named "Progress"). This is a real case where the warning is valuable.
- An alternative fix: deduplicate _tool_jira_issue_fields by (connection_id, field_id) before building the map, ignoring the board dimension.
- The warning has no functional impact on data import — it's purely log noise — but it degrades operator confidence and pollutes alerting systems.


### Version

v1.0.3-beta15

### Code of Conduct

bramhanandlingala:
closes: #9052

Summary
getIssueFieldMap() logged a "field name is duplicated" warning every time
a field name appeared more than once in _tool_jira_issue_fields, even when
it was the exact same field just present on multiple boards. On a
connection with several boards this produced hundreds of false-positive
warnings per pipeline run.

Root cause
_tool_jira_issue_fields stores one row per (connection_id, board_id, field
id), so a field present on N boards produces N rows with the same ID but
the same Name. getIssueFieldMap() warned on any second occurrence of a
Name without checking whether the ID actually differed, so board
recurrence was misread as a name collision.

Fix
Only warn when two different field IDs share the same display name, which
is the actual ambiguous case. The existing SchemaType == "user" override
is unchanged. Also fixed a typo in the log message: "filed name" -> "field
name".

### DIFF
--- a/backend/plugins/jira/tasks/epic_extractor.go
+++ b/backend/plugins/jira/tasks/epic_extractor.go
@@ -103,8 +103,11 @@ func getIssueFieldMap(db dal.Dal, connectionId uint64, logger log.Logger) (map[s
 	}
 	issueFieldMap := make(map[string]models.JiraIssueField)
 	for _, v := range allIssueFields {
-		if _, ok := issueFieldMap[v.Name]; ok {
-			logger.Warn(nil, "filed name %s is duplicated", v.Name)
+		if existing, ok := issueFieldMap[v.Name]; ok {
+			// Same field can appear on multiple boards; only warn on a real name collision.
+			if existing.ID != v.ID {
+				logger.Warn(nil, "field name %s is duplicated (IDs: %s, %s)", v.Name, existing.ID, v.ID)
+			}
 			if v.SchemaType == "user" {
 				issueFieldMap[v.Name] = v
 			}


klesh (approved):
LGTM.
Good catch.
Thanks for your contribution.

PR #9067, opened 2026-08-20
fix(jira): honor EpicKeyField when extracting issues

mfrancisc:
### Summary
`EpicKeyField` is already stored on Jira scope config and exposed in Config UI as the Epic Link field, but `extractIssues` never applied it. Epic keys were only taken from the legacy `fields.epic` object, so Jira Cloud / Parent Link custom fields stayed empty at the tool and domain layers.
This follows the same pattern as `StoryPointField` / `DueDateField` in the same extractor.

### Changes
- Map `ScopeConfig.EpicKeyField` from `AllFields`
- Include `epicKeyField` in `SubtaskConfig` so changing the transformation setting triggers a full re-extract
- Unit tests for string, object, legacy fallback, and empty-config (no accidental overwrite)


klesh (approved):
LGTM
Thanks for your contribution.

PR #9068, opened 2026-08-21
fix(server,asana): return real error from /proceed-db-migration instead of bare 500

### LINKED ISSUE #9066: [Bug][Database Migration] Operation Failed for devlake:v1.0.3-beta15
### Search before asking

### What happened

I currently have devlake version beta12, i can't upgrade to the newer version i always get 500 error when i migrate. 
August 20, 2026, 15:10 | [GIN] 2026/08/20 - 13:10:59 \| 500 \| 201.216µs \| 10.129.163.208 \| GET "/proceed-db-migration"
-- | --

Successfully upgrade and pass migration 

### How to reproduce

Image

### Anything else

_No response_

### Version

v1.0.3-beta15

### Code of Conduct

bujjibabukatta:
### Summary
Fixes #9066. `/proceed-db-migration` was returning a 500 with no body on upgrade, so there was no way to tell what actually failed without going into the container logs.

### Root cause
The endpoint wrapped `services.ExecuteMigration()` in `errors.Must()`, which panics on any error. Gin's recovery middleware turns that into an empty 500, so the real failure never made it back to the client.

While looking for what's actually breaking on the beta12 -> beta15 upgrade path, I found `20260509000001_encrypt_connection_token.go` in the asana plugin. It's new in beta13 and it's the first migration here that touches `ENCRYPTION_SECRET` during the migration itself. If that key is missing or doesn't match what was used before, it fails - and because of the bug above, that failure was invisible.

### Fix
- `api.go`: check the error from `ExecuteMigration()` and return it with `shared.ApiOutputError` instead of panicking, same as the rest of the file already does.
- asana migration: clearer error message when `ENCRYPTION_SECRET` isn't set, skip tokens that are already encrypted (so re-running after a failure doesn't double-encrypt them), and include the connection id if encryption fails on a row.

Note: I can't confirm the asana part is the exact thing the reporter hit since there's no log/stack trace in the issue, but with the api.go fix the next attempt will actually show the real error, so we'll know for sure.

### DIFF
--- a/backend/plugins/asana/models/migrationscripts/20260509000001_encrypt_connection_token.go
+++ b/backend/plugins/asana/models/migrationscripts/20260509000001_encrypt_connection_token.go
@@ -18,6 +18,8 @@ limitations under the License.
 package migrationscripts
 
 import (
+	"fmt"
+
 	"github.com/apache/incubator-devlake/core/context"
 	"github.com/apache/incubator-devlake/core/dal"
 	"github.com/apache/incubator-devlake/core/errors"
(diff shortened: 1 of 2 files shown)


klesh (approved):
LGTM

PR #9070, opened 2026-08-25
fix(gitlab): use stateful extractors for users and tags

mfrancisc:
## Summary

GitLab Extract Users and Extract Tags were the last GitLab extractors still using the deprecated `NewApiExtractor`. That helper always deletes tool-layer rows by `_raw_data_table` + `_raw_data_params` before insert.

On self-hosted GitLab, Collect Users can stop early when a page is shorter than `pageSize` (`fetchPagesUndetermined`) and still return success. The following Extract Users run then wipes accounts that project last wrote and keeps only the partial dump.

This switches both extractors to `NewStatefulApiExtractor`, matching MRs, notes, pipelines, and jobs:
- incremental runs upsert instead of delete-then-insert
- full delete still happens on first run, full refresh, or subtask config change (`skipCreatedUserAt`)


klesh (approved):

klesh:
LGTM, Thanks for your contribution.

PR #9074, opened 2026-08-27
feat: add Kiro commit attribution helper

warren830:
## Summary

- add an opt-in Kiro Stop hook recorder and Git `prepare-commit-msg` hook that append deterministic `Kiro-Session-Id` trailers
- add a one-command installer that respects `core.hooksPath`, preserves an existing hook, and handles managed unwritable hook directories
- register the current Kiro `v1` hook schema with the permitted `bash` command form
- document coverage limitations so an absent trailer is not misinterpreted when the hook was never installed

This helper is independent of the Kiro S3 collection plugin series and can be reviewed or landed separately.

## Validation

- `sh -n` on all three shell scripts
- `shellcheck` on all three shell scripts
- isolated install and `--check` smoke test in a temporary Git repository
- installer self-test verified both trailer creation after a recorded session and no inherited trailer without a new session

Signed-off-by: warren <warren.chen830@gmail.com>

### DIFF
--- a/tools/kiro-attribution/README.md
+++ b/tools/kiro-attribution/README.md
@@ -1,0 +1,132 @@
+# Binding Kiro sessions to git commits
+
+## What this solves
+
+Usage data alone cannot answer whether AI helped a given person: high credit
+consumption is compatible with both "the agent did a lot of work" and "the agent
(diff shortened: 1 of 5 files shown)


klesh (approved):
LGTM

PR #9083, opened 2026-08-30
server/services: add optional DeleteProject hook for plugin-owned project state

### LINKED ISSUE #9081: [Bug][DevLake] Project deletion lifecycle lacks extension hook for plugin-owned state
### Search before asking

### What happened

The generic `DELETE /projects/:projectName` endpoint deletes a project through `services.DeleteProject()` without exposing a lifecycle hook for plugins or features that maintain project-scoped state.

Core DevLake records such as the project, blueprint, mappings and metric settings are cleaned up, but feature-owned records can be left stale if their cleanup normally happens through a feature-specific or Config UI deletion flow.

There is already a project lifecycle extension for renames through `ProjectService.RenameProject(...)`, but there is no equivalent hook for deletion.

Project deletion should provide a lifecycle extension point that allows plugins or features with project-scoped state to validate and clean up their own records before the project is deleted.

Ideally this should integrate with the core deletion transaction where possible so project deletion and feature-owned cleanup cannot leave inconsistent state.

For example, the existing `ProjectService` boundary could expose a `BeforeDeleteProject` or `DeleteProject` hook similar to `RenameProject`.

### How to reproduce

1. Add or use a feature/plugin that persists a record associated with a DevLake project and normally removes or validates that record during its own project-deletion flow.
2. Create a DevLake project and associate feature-owned state with it.
3. Delete the project directly through:

   DELETE /projects/:projectName

4. Observe that `server/api/project.DeleteProject` calls `services.DeleteProject(projectName)` directly.
5. Observe that core project records are removed, but no plugin/feature deletion hook is invoked.
6. The feature-owned project association can therefore remain stale or orphaned.

This is reproducible on current `main` at commit:

f110addad86930c4c792a3d22051e922346f3656

### Anything else

_No response_

### Version

f110addad86930c4c792a3d22051e922346f3656

### Code of Conduct

bujjibabukatta:
### Summary
Fixes #9081. The generic `DELETE /projects/:projectName` flow
(`services.DeleteProject`) has no lifecycle hook for plugins/features
with project-scoped state, unlike `PatchProject`'s rename flow, which
already calls `projectService.RenameProject(...)`. Feature-owned
records can be left stale or orphaned when a project is deleted
directly through this endpoint.

### Root Cause
`ProjectService` only defines `RenameProject`. `DeleteProject` deletes
core records (project, mappings, metric settings, blueprint) with no
equivalent extension point, so any plugin/feature relying on its own
project-deletion flow to clean up state never gets invoked when
deletion happens through the core API.

### Fix
- Added `ProjectDeleteHook`, a separate optional interface (not a new
  required method on `ProjectService`), so existing downstream
  implementations that only support renaming aren't broken.
- `DeleteProject` now checks via type assertion whether the injected
  `projectService` also implements `ProjectDeleteHook`, and if so
  invokes it inside the same deletion transaction, before core rows
  are deleted — a hook error rolls back the whole deletion, and any
  cleanup it performs commits atomically with core deletion.

**Known limitation:** `deleteProjectBlueprint(name)` runs before the
transaction/hook, so a hook veto can't stop the blueprint from already
being deleted. Not addressed in this PR — flagging for a possible
follow-up.

### DIFF
--- a/backend/server/services/project.go
+++ b/backend/server/services/project.go
@@ -19,10 +19,11 @@ package services
 
 import (
 	"fmt"
-	"golang.org/x/exp/slices"
-	"golang.org/x/sync/errgroup"
 	"strings"
 	"time"
+
+	"golang.org/x/exp/slices"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/apache/incubator-devlake/core/dal"
 	"github.com/apache/incubator-devlake/core/errors"
@@ -33,6 +34,11 @@ import (
 
 type ProjectService interface {
 	RenameProject(db dal.Transaction, oldProjectName, newProjectName string) errors.Error
+}
+
+// ProjectDeleteHook is an optional hook; kept separate from ProjectService so existing implementers aren't broken.
+type ProjectDeleteHook interface {
+	DeleteProject(db dal.Transaction, projectName string) errors.Error
 }
 
 var projectService ProjectService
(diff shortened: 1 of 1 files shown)


klesh:
This doesn't make sense. You want plugins to depend on the backend/server package? Do you even understand the code you submitted? Please don't pick an issue you don't understand and generate random code, especially since the author of #9081 stated they will submit a PR for this matter.

PR #9000, opened 2026-07-20
feat(dora): exclude bot/automation accounts from PR Pickup Time calcu…

bramhanandlingala:
## Summary

Fixes #8974.

AI-powered code review bots (such as GitHub Copilot, Dependabot, and GitHub Actions) often post comments immediately after a pull request is created. DevLake currently considers the first PR comment as the start of the review process, regardless of whether the comment is from a human or an automated account. As a result, the `PR Pickup Time` metric can become artificially close to zero and no longer represent the actual time taken for a human reviewer to respond.

This PR introduces bot account detection and excludes bot-generated comments when calculating the first review event.

## Changes

- Added an `IsBot` field to the cross-domain `Account` model.
- Added a database migration to introduce the `is_bot` column in the `accounts` table with a default value of `false`.
- Updated the GitHub account converter to populate `IsBot` using:
  - GitHub account `type == "Bot"` when available.
  - The `[bot]` login suffix as a fallback.
- Updated the DORA `batchFetchFirstReviews` logic to ignore comments from bot accounts when determining the first review, ensuring `PR Pickup Time` and `PR Review Time` reflect the first human review.

## Scope

This change currently supports GitHub accounts only.

For other integrations (GitLab, Bitbucket, Azure DevOps, etc.), the `is_bot` field defaults to `false`, resulting in no behavioral changes until bot detection is implemented for those providers.

Fixes #8974

PR #9086, opened 2026-08-31
fix(github_graphql): resolve Bot actors so bot-authored PRs keep their author

### LINKED ISSUE #9085: [Bug][github_graphql] PR/issue author is empty for Bot actors (GitHub Apps, Dependabot) — GraphQL query only spreads "... on User"
### Search before asking

### What happened

When the GitHub connection has `Use GraphQL APIs` enabled, every pull request, issue and review opened by an actor of GraphQL type `Bot` (GitHub Apps, Dependabot, Renovate, GitHub Actions, any AI code-review or automation app) is stored with an **empty author**: `author_name = ''` and `author_id = 0` in `_tool_github_pull_requests`, and no `pull_requests.author_id` in the domain layer.

The data is already empty in the raw layer, so it cannot be recovered by re-running the extractor:

```json
"Author": {"Login": "", "Id": 0, "Name": "", "Company": "", "Email": "", ...}
```

Root cause is in `backend/plugins/github_graphql/tasks/account_graphql_pre_extractor.go`:

```go
type GraphqlInlineAccountQuery struct {
	GithubAccountEdge `graphql:"... on User"`
}

func extractGraphqlPreAccount(result *[]interface{}, res *GraphqlInlineAccountQuery, repoId int, connId uint64) {
	if res == nil || res.Id == 0 {
		return
	}
```

`author`, `mergedBy`, `assignees` and review authors are all typed as `Actor` in the GitHub schema. `Actor` is implemented by `User`, **`Bot`**, `Organization`, `Mannequin` and `EnterpriseUserAccount`. The query spreads only `... on User`, so for any non-`User` actor GitHub returns an empty selection set, and the `res.Id == 0` guard then drops the account entirely.

The REST collector is not affected — it resolves bot authors correctly — which is why a connection using GraphQL shows almost no bot-authored PRs while a REST one does.

PRs, issues and reviews opened by GitHub Apps / bots should carry their author, with the same identity the REST collector produces (`dependabot[bot]`, id `49699333`), so that `accounts` rows, `pull_requests.author_id` and any bot-aware metric work the same regardless of which collector is used.

### How to reproduce

The behaviour can be shown against the GitHub API directly, no DevLake instance needed. This query mirrors what the plugin sends today (`userFragmentOnly`) next to what it should send (`withBotFragment`):

```bash
gh api graphql -f query='
{
  search(query: "repo:grafana/grafana is:pr author:app/dependabot", type: ISSUE, first: 2) {
    nodes {
      ... on PullRequest {
        number
        userFragmentOnly: author { ... on User { login databaseId } }
        withBotFragment:  author { __typename ... on User { login databaseId } ... on Bot { login databaseId } }
      }
    }
  }
}'
```

Result:

```json
{"number":131544,
 "userFragmentOnly":{},
 "withBotFragment":{"__typename":"Bot","login":"dependabot","databaseId":49699333}}
```

End to end:

1. Create a GitHub connection with `Use GraphQL APIs` enabled, scoped to a repo that receives Dependabot/Renovate/GitHub App PRs.
2. Run the blueprint.
3. `SELECT number, author_name, author_id FROM _tool_github_pull_requests WHERE author_name = '';` — every bot-authored PR is listed.
4. Repeat with `Use GraphQL APIs` disabled: the same PRs come back with `dependabot[bot]` / `49699333`.

### Anything else

Happens every time, for every bot-authored PR, on every repo. In our deployment roughly 1,000 PRs opened by our own GitHub App over the last eight months are unattributed, and all 18 bot accounts in `_tool_github_accounts` (collected from comments and reviews) have zero PRs linked to them.

Three details worth knowing before writing the fix:

**1. `... on Bot` may only be spread where the schema says `Actor`.** `GraphqlInlineAccountQuery` is also used for fields typed `User` — PR and issue assignees, and `commit.author.user` — and spreading the fragment there makes GitHub reject the *entire* query, so PR and issue collection would stop working:

```
gh api graphql -f query='
{ repository(owner:"grafana", name:"grafana") { pullRequest(number:131544) {
  assignees(first:1){ nodes { ... on User { login } ... on Bot { login } } } } } }'

{"errors":[{"extensions":{"code":"cannotSpreadFragment","typeName":"Bot","parentName":"User"},
  "message":"Fragment on Bot can't be spread inside User"}]}
```

The `Actor` fields need their own type, leaving the existing one `User`-only:

```go
type GithubBotEdge struct {
	Login     string
	Id        int `graphql:"databaseId"`
	AvatarUrl string
	HtmlUrl   string `graphql:"url"`
}

// for PullRequest.author, PullRequest.mergedBy, PullRequestReview.author, Issue.author
type GraphqlInlineActorQuery struct {
	GithubAccountEdge `graphql:"... on User"`
	Bot               GithubBotEdge `graphql:"... on Bot"`
}
```

The named-field-with-inline-fragment-tag pattern is already used in this codebase for `RequestedReviewer` (#7716), so the client handles it.

**2. `Bot` has a different field set than `User`.** It exposes `login`, `databaseId`, `avatarUrl`, `url`, `createdAt`, `updatedAt` — but no `name`, `company` or `email`, hence the separate struct above. Leaving name, company and email empty matches what the REST collector stores for a bot: `GET /users/dependabot[bot]` returns `null` for all three.

**3. `Bot.login` omits the `[bot]` suffix.** GraphQL returns `dependabot`, REST returns `dependabot[bot]`; the `databaseId` is the same `49699333` in both, so there is no risk of duplicate accounts, but the logins would diverge between collectors.

This matters for the `is_bot` flag added in #9000: the fallback is `strings.HasSuffix(githubUser.Login, "[bot]")`, and `_tool_github_accounts.type` is never populated on the GraphQL path (`GithubAccountEdge` has no `Type` field, and `extractGraphqlPreAccount` only writes `GithubRepoAccount`). So a naive fix would produce bot accounts that `is_bot` cannot detect. The normalisation should append `[bot]` to match REST, and ideally set `Type: "Bot"` on the `GithubAccount` row.

It also interacts with #8886 / #8894: `pr_convertor.go` now guards `if pr.AuthorId != 0` before generating the account id, so as long as the author stays empty these PRs will keep having no `author_id` by design — the fix has to happen at collection time.

Existing data cannot be repaired by re-running extract or convert; a **Full Refresh** re-collect is required, since the author is already empty in `_raw_github_graphql_prs`.

Same fix applies to `mergedBy`, which is currently empty for merges performed by a bot.

### Version

Verified on `main` @ `79ef9f4f`; the code is unchanged since #8583 (2025-09-24) and is present in `v1.0.3-beta16`.

### Code of Conduct

zyv4yk:
### ⚠️ Pre Checklist

### Summary

When the GitHub connection runs on GraphQL, every pull request, issue and review opened by an actor of type `Bot` — GitHub Apps, Dependabot, Renovate, GitHub Actions — is stored with no author: `author_name = ''`, `author_id = 0`, and no `pull_requests.author_id` in the domain layer.

`author`, `mergedBy` and review authors are typed `Actor` in the GitHub schema, and `Actor` is implemented by `User`, `Bot`, `Organization`, `Mannequin` and `EnterpriseUserAccount`. `GraphqlInlineAccountQuery` spread only `... on User`, so a `Bot` author came back as an empty selection set and the `Id == 0` guard in `extractGraphqlPreAccount` then dropped the account. The REST collector is unaffected, which is why the same repo shows bot PRs on REST and none on GraphQL.

**The fragment cannot just be added to the existing struct.** `GraphqlInlineAccountQuery` is reused for fields the schema types as `User` — PR and issue assignees, and `commit.author.user` — and GitHub rejects the *entire* query when `... on Bot` is spread there:

```
Fragment on Bot can't be spread inside User
```

So this PR splits the two cases:

- `GraphqlInlineAccountQuery` — unchanged shape, for `User`-typed fields (assignees, commit authors).
- `GraphqlInlineActorQuery` — `... on User` + `... on Bot`, for `Actor`-typed fields (`PullRequest.author`, `PullRequest.mergedBy`, `PullRequestReview.author`, `Issue.author`).

Both expose `Account() GithubAccountEdge`, so the extractors read one normalized account and stop caring which type the actor resolved to. The accessor is nil-safe, which also removes the `!= nil` checks at the call sites.

Two details in the normalization:

1. **`Bot.login` has no `[bot]` suffix.** GraphQL returns `dependabot`, REST returns `dependabot[bot]`, both with `databaseId` `49699333`. The suffix is appended so one actor does not end up with two logins depending on the collector. This also keeps the `is_bot` flag from #9000 working: its fallback is `strings.HasSuffix(login, "[bot]")`, and `_tool_github_accounts.type` is never populated on the GraphQL path.
2. **`Bot` exposes a narrower field set** than `User` — no `name`, `company` or `email` — hence a separate `GithubBotEdge`. `Name`, `Company` and `Email` are left empty, which is exactly what the REST collector stores for a bot (`GET /users/dependabot[bot]` returns `null` for all three).

One follow-on: `Collect Users` looks bot logins up through `user(login:)`, which resolves Users only and answers `Could not resolve to a User with the login of 'dependabot[bot]'`. Errors there are already ignored, but now that bot logins reach `_tool_github_repo_accounts` that would be one guaranteed `NOT_FOUND` per bot on every sync, so the cursor skips them. Bots still become accounts — `ConvertAccounts` reads `_tool_github_repo_accounts` and only enriches from `_tool_github_accounts` when a row exists.

Closes #9085

### Screenshots

Running the query this PR generates against the live API, `nodes { author { login databaseId } }` on `grafana/grafana`:

| | before | after |
|---|---|---|
| `#131864` | `""` / `0` | `cursor` / `206951365` |
| `#131868` review authors | `""`, `""` | `github-actions`, `cursor` |
| `#131851` review authors | `""`, `""`, `""` | `copilot-pull-request-reviewer`, `cursor`, `PranshulSoni` |

The single-field version of the same check, which needs nothing but a token:

```
gh api graphql -f query='
{
  search(query: "repo:grafana/grafana is:pr author:app/dependabot", type: ISSUE, first: 1) {
    nodes { ... on PullRequest {
      number
      userFragmentOnly: author { ... on User { login databaseId } }
      withBotFragment:  author { __typename ... on User { login databaseId } ... on Bot { login databaseId } }
    } }
  }
}'

{"number":131544,
 "userFragmentOnly":{},
 "withBotFragment":{"__typename":"Bot","login":"dependabot","databaseId":49699333}}
```

### Other Information

**Tests.** `account_graphql_pre_extractor_test.go` covers the accessor for both query types (user, bot, bot login that already carries the suffix, unresolved actor, nil), the raw-layer JSON round trip in both the new and the pre-fix shape, and — the part that actually guards the regression — `graphql.ConstructQuery` output, asserting `... on Bot` reaches the actor fields and never the `User`-typed ones. `go build`, `go vet` and `golangci-lint run ./plugins/github_graphql/...` are clean. The `e2e` package needs `E2E_DB_URL` and was not run locally; the only `github_graphql` e2e fixture covers deployments, which this PR does not touch.

**Existing data.** The author is already empty in `_raw_github_graphql_prs`, so rows collected before this fix are not repaired by re-running extract or convert — they need a re-collect in Full Refresh mode.

**Labels.** I do not have permission to set them; this is `pr-type/bug-fix`.

### DIFF
--- a/backend/plugins/github_graphql/tasks/account_collector.go
+++ b/backend/plugins/github_graphql/tasks/account_collector.go
@@ -75,10 +75,17 @@ func CollectAccount(taskCtx plugin.SubTaskContext) errors.Error {
 	db := taskCtx.GetDal()
 	data := taskCtx.GetData().(*githubTasks.GithubTaskData)
 
+	// `user(login:)` resolves Users only, a bot login always comes back as
+	// "Could not resolve to a User", so bots are left out of this lookup. They are
+	// still converted to accounts: ConvertAccounts reads _tool_github_repo_accounts
+	// and only enriches it from _tool_github_accounts when a row is there.
 	cursor, err := db.Cursor(
 		dal.Select("login"),
 		dal.From(models.GithubRepoAccount{}.TableName()),
-		dal.Where("repo_github_id = ? and connection_id=?", data.Options.GithubId, data.Options.ConnectionId),
+		dal.Where(
+			"repo_github_id = ? and connection_id = ? and login not like ?",
+			data.Options.GithubId, data.Options.ConnectionId, "%"+botLoginSuffix,
+		),
 	)
 	if err != nil {
 		return err
--- a/backend/plugins/github_graphql/tasks/account_graphql_pre_extractor.go
+++ b/backend/plugins/github_graphql/tasks/account_graphql_pre_extractor.go
@@ -18,8 +18,15 @@ limitations under the License.
 package tasks
 
 import (
+	"strings"
+
 	"github.com/apache/incubator-devlake/plugins/github/models"
 )
+
+// botLoginSuffix is how the REST API reports a bot login: `dependabot[bot]`.
+// GraphQL's `Bot.login` omits it, so it is appended back to give a bot the same
+// identity no matter which collector produced it.
+const botLoginSuffix = `[bot]`
 
 type GithubAccountEdge struct {
 	Login     string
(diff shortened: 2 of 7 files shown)


klesh (approved):
LGTM

PR #8573, opened 2025-09-13
Fix/GitHub

GoSimplicity:
enhance pull request handling by including merged state and additional metrics GitHub API pull request model. #8571

Issue #8571, opened 2025-09-12
[Bug][GitHub] Merged PRs incorrectly stored as closed with merged = false since March 2025


jcg-juan:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-devlake/issues?q=is%3Aissue) and found no similar issues.

### What happened

We’ve identified a data inconsistency with how DevLake ingests GitHub Pull Requests. Since around March 2025, all PRs that are merged in GitHub are being stored in DevLake as state = closed and merged = false. This results in:

- Merged PRs not being reflected correctly in queries (state = 'MERGED' returns nothing after February).
- Metrics such as additions/deletions for merged PRs show up as 0.
- Dashboards and reports undercounting merged PR activity.

Impact:
- Any downstream metrics that rely on merged PRs (e.g., code change size, lead time, DORA metrics) are currently inaccurate.
- Dashboards show zero additions/deletions after Feb 2025.

### What do you expect to happen

- DevLake should correctly recognise merged PRs (state = 'MERGED') or, at a minimum, set the merged column to true when GitHub API reports it.
- Additions/deletions should be populated for merged PRs the same way as open ones.

### How to reproduce

Steps to Reproduce:
- Ingest PR data from GitHub into DevLake.
- Run the following query to check the latest PRs:
```
SELECT *
FROM public._tool_github_pull_requests
ORDER BY github_updated_at DESC
LIMIT 1000;
```

- Observe that since March 2025, all PRs are either open or closed, never MERGED.
- Additionally, **merged** column is always **false**, even when GitHub API reports merged = true.

### Anything else

1. Example GitHub API response for a merged PR (PR #1 before March working as expected):
```
curl -s -H "Accept: application/vnd.github+json" \
   -H "Authorization: Bearer <token>" \
   https://api.github.com/repos/<org>/<repo>/pulls/1 \
   | jq '{number, state, merged, additions, deletions}'
```
Response:
```
{
  "number": 1,
  "state": "closed",
  "merged": true,
  "additions": 203,
  "deletions": 0
}

From Devlake DB (table _tool_github_pull_requests): 
```
<html><head></head><body>
[...]

PR #9087, opened 2026-08-31
fix(github_graphql): set the merged flag on pull requests

zyv4yk:
### ⚠️ Pre Checklist

### Summary

`_tool_github_pull_requests.merged` is `false` for every pull request collected through GraphQL, merged ones included.

`convertGithubPullRequest` in `backend/plugins/github_graphql/tasks/pr_extractor.go` fills `MergedAt`, `MergedBy` and `MergeCommitSha`, but never `Merged`, so the column keeps its zero value:

```go
ClosedAt:        pull.ClosedAt,
MergedAt:        pull.MergedAt,
Body:            pull.Body,
```

#8573 closed the same gap on the REST side — it added `Merged: pull.Merged` to `plugins/github/tasks/pr_extractor.go` — but the GraphQL extractor was not part of that change, so a connection running on GraphQL still stores the flag as `false`.

The flag is derived from `mergedAt` rather than collected as a separate `merged` field, for a practical reason: `mergedAt` is already present in `_raw_github_graphql_prs` for existing installs, so re-running the extractor repairs rows that are already collected. Querying `merged` would only fix data going forward and would need a full re-collect for the rest.

The two are equivalent by construction — GitHub sets `mergedAt` exactly when a pull request is merged. Checked against the API over the 50 most recent pull requests in `grafana/grafana` (18 merged): zero rows where `bool(merged) != (mergedAt != null)`.

```
gh api graphql -f query='{ repository(owner:"grafana",name:"grafana"){
  pullRequests(first:50, states:[MERGED,CLOSED,OPEN], orderBy:{field:CREATED_AT,direction:DESC}){
    nodes { number state merged mergedAt } } } }'
```

No open issue. It is the GraphQL half of #8571, which was closed by #8573 for the REST collector only.

### Other Information

Impact is limited today because the domain layer does not read this column for status: `pr_convertor.go` computes `Status` from `pr.State == "MERGED" || (pr.State == "closed" && (pr.Merged || pr.MergedAt != nil))`, and the GraphQL state is already `MERGED`. Any consumer that reads `_tool_github_pull_requests.merged` directly, or any future query that keys on `merged = true`, gets nothing for GraphQL-collected data.

`go build`, `go vet` and `golangci-lint run ./plugins/github_graphql/...` are clean. `pr_extractor_test.go` covers merged, closed-unmerged and open. The `e2e` package needs `E2E_DB_URL` and was not run locally; the only `github_graphql` e2e fixture covers deployments, which this PR does not touch.

I do not have permission to set labels; this is `pr-type/bug-fix`.

### DIFF
--- a/backend/plugins/github_graphql/tasks/pr_extractor.go
+++ b/backend/plugins/github_graphql/tasks/pr_extractor.go
@@ -171,6 +171,7 @@ func convertGithubPullRequest(pull *GraphqlQueryPr, connId uint64, repoId int) (
 		GithubUpdatedAt: pull.UpdatedAt,
 		ClosedAt:        pull.ClosedAt,
 		MergedAt:        pull.MergedAt,
+		Merged:          pull.MergedAt != nil, // mergedAt is set exactly when a pull request is merged
 		Body:            pull.Body,
 		BaseRef:         pull.BaseRefName,
 		BaseCommitSha:   pull.BaseRefOid,
--- a/backend/plugins/github_graphql/tasks/pr_extractor_test.go
+++ b/backend/plugins/github_graphql/tasks/pr_extractor_test.go
@@ -1,0 +1,56 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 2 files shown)


zyv4yk:
Closing this one — the domain layer does not read this column today (`pr_convertor.go` derives status from the state and `mergedAt`), so the practical impact is limited, and I would rather not split review attention away from #9086. Happy to reopen if a maintainer wants the flag stored for parity with the REST extractor.

PR #9093, opened 2026-09-02
chore: remove the incubator prefix

klesh:
### Summary
Remove the `incubator-` prefix as the project graduated already.


Startrekzky (approved):
lgtm

PR #9056, opened 2026-08-17
chore(deps): bump CI Go bootstrap to 1.26.6

DoDiODev:
### Summary

Upgrade the job-local Go bootstrap from 1.26.2 to 1.26.6 in the Apache CI
workflows. Go 1.26.6 contains the fixes for the standard-library advisories
reported against earlier 1.26 patch releases.

The change intentionally keeps `backend/go.mod` and all `golang:1.26-*` image
tags on the floating 1.26 minor line. It only changes:

- `backend/scripts/install-go.sh`: version and official Linux amd64/arm64
  SHA-256 checksums;
- the four visible bootstrap step names in the lint, unit-test, MySQL E2E and
  PostgreSQL E2E jobs.

### Security verification

`govulncheck` v1.7.0 was run with Go 1.26.6 against the complete backend package
graph after mock generation. It reports **zero reachable standard-library
findings**. Remaining findings belong to Go modules and are handled separately;
this PR does not mix module upgrades into the toolchain patch.

Both Linux archive checksums were verified against the official Go download
catalog and the downloaded archives. In the unchanged
`mericodev/lake-builder:latest` image, the installer was tested from the image's
Go 1.20.4 baseline:

1. install Go 1.26.6 into `/opt/go/1.26.6`;
2. verify `go version go1.26.6 linux/amd64`;
3. run the installer again and verify the no-op path.

### Verification

- `sh -n backend/scripts/install-go.sh`
- `actionlint` on all three changed workflows
- installer checksum/install/idempotency test in
  `mericodev/lake-builder:latest`
- `make mock` — 64 generated mock files
- `make unit-test` — green, including Azure DevOps 8 passed/1 skipped and
  pydevlake 10 passed
- `make e2e-test-go-plugins` and `make e2e-test` against isolated MySQL 8 — green
- `make e2e-test-go-plugins` and `make e2e-test` against isolated PostgreSQL
  14.2 — green
- `govulncheck ./...` with Go 1.26.6 — 0 reachable standard-library findings

### Builder image

The published builder image still contains Go 1.20.4. The job-local,
checksum-verified bootstrap keeps the affected jobs independent of a builder
republish; refreshing the image remains a separate maintainer follow-up.

PR #9096, opened 2026-09-02
build(ci): pin Go in migration-script-lint like the other container jobs

DoDiODev:
### Summary

`migration-script-lint` is the only job that runs inside
`mericodev/lake-builder:latest` **without** installing the pinned Go toolchain.
The image still ships **Go 1.20.4**, while `backend/go.mod` declares `go 1.26`.

`test.yml`, `test-e2e.yml` and `golangci-lint.yml` all bootstrap the toolchain
via `backend/scripts/install-go.sh` (added in #9031 / #9056). This PR adds the
same step to the remaining job, using the identical step name and command:

```yaml
      - uses: actions/checkout@v6
      - name: Install Go 1.26.6
        run: backend/scripts/install-go.sh
      - name: migration script linting
```

### Why it matters even though the job is green today

Go ≤ 1.20 does not treat a higher `go` directive as a hard error — it only
reports it when compilation fails for another reason. The job therefore passes,
but it builds and runs `core/migration/linter/main.go` under a language version
six major releases behind what the project declares. Any migration script that
uses post-1.20 language or standard-library features would fail here for a
reason unrelated to the lint itself, and the failure message would point at the
wrong thing.

This also removes the last place where the mutable `:latest` builder image
silently determines the Go version used to run project code.

### Cost

`install-go.sh` is idempotent: it checks for an already-installed toolchain
under `/opt/go/$GO_VERSION` before downloading. Measured on a verification run
(worst case, nothing pre-installed): **2.8 s** for download, checksum
verification and extraction, in a job that takes about **45 s** in total.

`install-libgit2.sh` and `install-mockery.sh` are deliberately **not** added —
the migration linter needs neither cgo bindings nor generated mocks.

### Validation

Verified on a fork branch that mirrors the upstream workflow step lists:
[run 33620580891](https://github.com/DoDiODev/devlake/actions/runs/33620580891),
13/13 jobs green. In the `migration-script-lint` job the linting step now
reports:

```
installed Go 1.26.6 in /opt/go/1.26.6
go version go1.26.6 linux/amd64
```

instead of the image-provided 1.20.4.

### Note for reviewers

This PR does not require a new builder image. It works with the current
`mericodev/lake-builder:latest` precisely because `install-go.sh` installs the
toolchain at job runtime — the same mechanism the other three container jobs
already rely on.

### DIFF
--- a/.github/workflows/migration-script-lint.yml
+++ b/.github/workflows/migration-script-lint.yml
@@ -30,6 +30,8 @@ jobs:
     container: mericodev/lake-builder:latest
     steps:
       - uses: actions/checkout@v6
+      - name: Install Go 1.26.6
+        run: backend/scripts/install-go.sh
       - name: migration script linting
         run: |
           go version


klesh (approved):
LGTM
Thanks for your contribution.

PR #9130, opened 2026-09-10
fix(config-ui): group connections by name, not by list position

stigi:
### Summary

The **A-N** and **O-Z** headings on the Connections page put several plugins under
the wrong letter. `Asana`, `Kiro`, `Linear` and `incident.io` all appear under
**O-Z** today.

The grouping cut the list at the first plugin *id* starting with a letter from
`o` to `z`, and treated everything after that point as O-Z:

```ts
const index = filterWebhookPlugins.findIndex((p) => SORT_START_WITH.includes(p[0]));
// A-N = slice(0, index), O-Z = slice(index)
```

That is only equivalent to grouping by letter while each plugin config's `sort`
value happens to run alphabetically. It no longer does: plugins have been
appended in the order they were added, so `opsgenie` (sort 10) sits ahead of
`asana` (12), `kiro` (12), `linear` (13) and `incidentio` (19), and the cut
lands in the middle of the alphabet.

This groups by the first letter of the **displayed name** instead — what a
reader is actually matching the heading against — and extracts the decision
into a pure `splitPluginsByInitial` with unit tests.

No issue filed; the misplaced tiles are visible on the Connections page of any
current build.

### Screenshots

Not included: the change moves four existing tiles between two headings, which
reads more clearly as the plugin ids above than as a screenshot.

### Tests

`config-ui/src/__tests__/connections-grouping.test.ts` — four cases: grouping by
name rather than position, no plugin lost from either group, case-insensitive
comparison (so `incident.io` groups under I), and a fallback to the plugin id
when a config has no name. Verified against the previous implementation: two of
the four fail on it.


klesh (approved):
LGTM.
Thanks for your contribution.

PR #9132, opened 2026-09-11
fix(build): use libgit2's bundled zlib so the server image builds on arm64 hosts

stigi:
### Summary

Building the server image on an arm64 host fails while linking libgit2 for the
amd64 target, before any Go code is compiled:

```
/rootfs-amd64/usr/lib/x86_64-linux-gnu/libz.a(deflate.o): relocation R_X86_64_PC32
  against symbol `z_errmsg' can not be used when making a shared object; recompile with -fPIC
/usr/lib/gcc-cross/x86_64-linux-gnu/12/../../../../x86_64-linux-gnu/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
make[2]: *** [src/libgit2/CMakeFiles/libgit2package.dir/build.make:476: libgit2.so.1.5.0] Error 1
```

`backend/Dockerfile` builds libgit2 with `BUILD_SHARED_LIBS=ON`, and cmake
resolves zlib to the static `libz.a` in the staged cross sysroot. That archive
is not built with `-fPIC`, so it cannot be linked into a shared object. Point
libgit2 at its own bundled zlib for both cross targets instead: it is compiled
as part of the shared library and is therefore position independent.

The failure is host-architecture dependent, which is why CI does not see it —
on an amd64 builder the amd64 target is native and picks up a usable zlib. On
Apple Silicon, which is where a lot of contributors work, `docker build
--platform linux/amd64` cannot produce the image at all.

<details>
<summary>Why not the alternatives</summary>

- Installing `zlib1g-dev` into the staged sysroots would pull the same static
  archive; the shared `libz.so` symlink is what is missing, and staging it
  correctly for both architectures is more moving parts than one cmake flag.
- `CMAKE_POSITION_INDEPENDENT_CODE=ON` does not help: the offending object is
  in a prebuilt Debian archive, not in anything this build compiles.
- Building only the target architecture would also work, but the loop over both
  architectures looks deliberate, so this change leaves it alone.

</details>

No issue filed — the reproduction and the error are above. Happy to open one if
you would prefer the bug tracked separately.

### Screenshots

Not applicable; this is a build change.

### Verification

On an Apple Silicon host (Docker Desktop, `--platform linux/amd64`):

- before: fails at `Dockerfile:64`, the libgit2 stage, with the relocation error above
- after: links `libgit2.so`, the stage completes, and the build proceeds through
  the Go plugin build to a working image

The resulting amd64 image has been running a DevLake instance, collecting from
a live data source, without issue.


klesh (approved):
LGTM.
Thanks for your contribution.

PR #9147, opened 2026-09-17
fix(server): ignore Basic auth header when AUTH_ENABLED=true

### LINKED ISSUE #9146: [Bug][server] Any Basic credentials pass RequireAuth when AUTH_ENABLED=true: OAuth2ProxyAuthentication never verifies the password
### Search before asking

### What happened

With `AUTH_ENABLED=true` (native OIDC, #8854), every non-public API route accepts **any** `Authorization: Basic <user>:<pass>` header as an authenticated user. Nothing checks the password.

`OAuth2ProxyAuthentication` (`backend/server/api/middlewares.go`) predates native auth. It decodes the Basic header and sets `common.USER` to the username, because in the classic layout nginx in config-ui has already verified the credentials against `ADMIN_USER`/`ADMIN_PASS` before proxying to the lake, and the lake only needed the name for audit fields:

```go
func getBasicAuthUserInfo(c *gin.Context, basicRes context.BasicRes) (*common.User, error) {
	...
	userInfo := strings.Split(string(userInfoData), ":")
	if len(userInfo) != 2 {
		return nil, errors.Default.New("invalid user info data")
	}
	return &common.User{Name: userInfo[0]}, nil
}
```

#8854 added `auth.RequireAuth()` right after it (`backend/server/api/api.go`):

```go
router.Use(RestAuthentication(router, basicRes))
router.Use(RequirePushAuthentication(basicRes))
router.Use(auth.OIDCAuthentication())
router.Use(OAuth2ProxyAuthentication(basicRes))
router.Use(auth.RequireAuth())
```

`RequireAuth` only asks whether a user is set, so the unverified Basic username now satisfies the gate. #8880 closed the same hole for `X-Forwarded-User` by requiring `FORWARDED_USER_SECRET`, but left the Basic branch as is.

Any deployment that exposes the lake directly with `AUTH_ENABLED=true` (the documented setup for native OIDC, where config-ui runs without `ADMIN_USER`/`ADMIN_PASS`) is affected: blueprints, connections (with stored tokens), pipelines and API keys are reachable and writable without signing in. `CSRFProtect` does not apply either, since the request carries no session cookie.

With `AUTH_ENABLED=true` the lake is the authenticator. Only a valid session cookie, a valid API key, or forwarded headers with the matching `FORWARDED_USER_SECRET` should produce a user. A Basic header the lake cannot verify must not.

### How to reproduce

Lake with `AUTH_ENABLED=true`, `OIDC_ENABLED=true`, one provider configured, reachable directly (not through config-ui nginx):

```bash
curl -s -o /dev/null -w '%{http_code}\n' "$LAKE/blueprints?page=1&pageSize=1"
# 401
curl -s -o /dev/null -w '%{http_code}\n' -u 'anyone:anything' "$LAKE/blueprints?page=1&pageSize=1"
# 200, full blueprint list
curl -s -o /dev/null -w '%{http_code}\n' -H 'Authorization: Bearer nope' "$LAKE/blueprints?page=1&pageSize=1"
# 401
```

`Authorization: Basic Og==` (empty user) and a non-base64 value give 401, so it is the decoded username alone that authenticates.

### Anything else

Every time, on every non-public route. Verified on `v1.0.3-beta17` (8fe26f4); `middlewares.go` is unchanged on `main`. Present since `v1.0.3-beta13`, the first release with #8854.

Fix I propose: consult the Basic header only when `AUTH_ENABLED` is off, which is the legacy nginx mode. With `AUTH_ENABLED` the supported callers are already what `auth.go` logs: "only API-key/proxy auth will work". Deployments that still front the lake with nginx Basic auth and also set `AUTH_ENABLED=true` can move to `FORWARDED_USER_SECRET`. A 9-line change plus tests; the bypass test fails on current code. Running as a local patch on our instance since 2026-09-17.

### Version

v1.0.3-beta17 (8fe26f46), same code on `main`.

### Code of Conduct

celeronsx:
### ⚠️ Pre Checklist

### Summary

`OAuth2ProxyAuthentication` turns any `Authorization: Basic user:pass` into `common.USER` without checking the password. That was the contract of the nginx era: config-ui verified the credentials against `ADMIN_USER`/`ADMIN_PASS` and the lake kept the username for audit fields. #8854 put `auth.RequireAuth()` behind that middleware, so with `AUTH_ENABLED=true` the unverified username now passes the terminal gate and any `user:pass` reaches every non-public route on a directly exposed lake. #8880 fixed the sibling `X-Forwarded-User` branch with `FORWARDED_USER_SECRET` but left this one.

The change gates the Basic branch on `AUTH_ENABLED` being off (mirroring `oidchelper.LoadConfig`: unset counts as off). With `AUTH_ENABLED=true` the accepted callers are the ones `auth.go` already logs as supported: session cookie, API key, forwarded headers with the shared secret. Legacy deployments without `AUTH_ENABLED` keep the Basic username exactly as before.

Closes #9146

### Other Information

**Tests.** `middlewares_basicauth_test.go`: Basic header is ignored with `AUTH_ENABLED=true`, and is still taken as the user with `AUTH_ENABLED` unset or `false`. `newProxyAuthRouter` in the existing forward-secret test gained a `WithConfig` variant so both files share the router. The new bypass test fails on `main` without the change. `go build`, `go vet` and `go test ./server/api/` are clean after `make mock`. Running as a local patch on our v1.0.3-beta17 instance since 2026-09-17: Basic returns 401, OIDC sign-in and API-key callers unaffected.

**Behaviour change.** A deployment that sets `AUTH_ENABLED=true` and still relies on config-ui nginx Basic auth for the lake will start getting 401 on the API. That combination was never a verified path (the lake trusted the header blindly); `FORWARDED_USER_SECRET` from #8880 is the supported way to pass a proxy-verified identity.

**Labels.** I do not have permission to set them; this is `pr-type/bug-fix`.


klesh (approved):
LGTM
Thanks for your contribution.

PR #9153, opened 2026-09-19
fix(gitlab) - fix added to get pathWithNamespace to render in UI

### LINKED ISSUE #8827: [Bug][GitLab] Same named repositories not well handled
### Search before asking

### What happened

When adding scope to the GitLab connection, if there are multiple repos with the same name, all repos are populated with the same name, and this makes it impossible to tell which exact repo I want to add to a project.

Image

When adding repos to the GitLab connection, if there's an already existing repo with that name, the user should be prompted to take an action. Rename the repo from their source, rename the repo as  devlake, will reference it, or, when a user is choosing the repo to add, it should show the full path of the repo. 

NB: This happens because GitLab has groups and subgroups that might have repos with the same name. 

### How to reproduce

Have to repos with the same name and add them to the GitLab Connection.
Create a project.
Add scopes to the project.
When adding the repo scopes, you'll see both named correctly
It will be difficult to tell which repo you want for this project

### Anything else

_No response_

### Version

v1.0.3-beta10@99376a8

### Code of Conduct

veetmoradiya3628:
### Summary
Added `pathWithNamespace` field reference priority for gitlab connection remote scopes so that it renders full path for options in UI.

Closes #8827 

### Screenshots
image

### Other Information
backend and scopeId is well handled with `gitlabId` so no impact on the existing flow.


klesh (approved):
LGTM

